PR #2, opened 2015-09-14
Add support for AWS STS session tokens when installing convox

jbuck:
Fix https://github.com/convox/cli/issues/86


ddollar:
Looks good but would you please rebase against master. Sorry, things are moving fast today :)

ddollar:
Actually just had to change a Github setting that was preventing me from merging. Thanks!

jbuck:
@ddollar Thanks!

PR #11, opened 2015-09-15
report actual error on login

csquared:


csquared:
note: it would be better to throw a proper error than testing for 401 via `strings.Contains`. that will come

PR #26, opened 2015-09-17
read aws credentials from STDIN

### LINKED ISSUE #19: Use STS tokens for install
As suggested by @slaskis:

`aws sts get-session-token | convox install ...`

csquared:
fixes #19

works like:

`aws sts get-session-token | convox install`


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;wfVmOgsn&#x2F;143-use-sts-tokens-for-install"> Use STS tokens for install</a>

ddollar:
mind adding a few tests around this one? i've added a couple of other examples of cli tests for other commands

csquared:
i thought you'd never ask :)

csquared:
so testing this isn't super straightforward. a call to install makes multiple network requests to four different hosts:

- `convox.s3.amazonaws.com` (via `convox/release/version`
- mixpanel
- aws
- the newly installed kernel

so far i've repurposed some of the controller test code to mock out the interaction w. aws.
right now i just return in install after the call to CreateStack so i could make sure that host was set up correctly.  

stubbing out the new kernel should follow the pattern of some of the other tests.

still not sure what i want to do with S3 and mixpanel. i'm tempted to just let the S3 one hit the network but don't want to do that for mixpanel. thoughts welcome.

ddollar:
I would probably stub out the version-getter and mixpanel. You can create a VersionChecker variable that normally is the object from `convox/release/version` but can be set to a mock object during the tests.

cmd/convox/install_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/convox/rack/api/awsutil"
+	"github.com/convox/rack/test"
+	"github.com/convox/release/version"
+)
+
+func TestConvoxInstallSTDINCredentials(t *testing.T) {
+	stackId := "arn:aws:cloudformation:us-east-1:123456789:stack/MyStack/aaf549a0-a413-11df-adb3-5081b3858e83"
+	cycles := []awsutil.Cycle{
+		awsutil.Cycle{
+			awsutil.Request{"/", "", "/./"},
```
csquared: have to update this to check what we put in

ddollar:
Use pointers to return credentials so you can nil them

cmd/convox/install_test.go
```diff
@@ -0,0 +1,47 @@
...
+		},
+	}
+
+	handler := awsutil.NewHandler(cycles)
+	s := httptest.NewServer(handler)
+	defaults.DefaultConfig.Endpoint = &s.URL
+
+	defer s.Close()
+
+	latest, _ := version.Latest()
```
csquared: just specify an explicit version

csquared:
even if i pass in a version, we make network requests. this is ready to merge

ddollar:
:+1:

PR #27, opened 2015-09-17
Apps scoped to rack

ddollar:
This maps Racks only see apps that they created. Old apps created before this code will be seen by all Racks.


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;clUdQElT&#x2F;144-apps-scoped-to-a-rack"> Apps scoped to a rack</a>

ddollar:
Looking at testing this it seems like `stubAws` and friends are too abstracted. Tests should probably just have XML pairs embedded in the tests.

PR #28, opened 2015-09-17
Fix errors for attached processes

ddollar:


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;8rpCztnz&#x2F;138-respond-with-a-useful-error-if-a-run-is-attempted-against-a-process-that-isn-t-defined"> Respond with a useful error if a run is attempted against a process that isn&#39;t defined</a>

csquared:
This used to trigger a kernel panic, now it doesn't:

```
$ convox run --app compose foo foo
ERROR: no releases for app: compose
```

PR #30, opened 2015-09-17
document building locally

csquared:


ddollar:
Can you put the install --development instructions in that section as well? Can move the whole section above `convox start` and call it **Prerequisites**

csquared:
will do

csquared:
i kinda want to make it possbile to do `convox install —development > .env` or have some other way of getting the ENV data to the `.env` file without cut and paste. closing the terminal before copying has bit me more than once.

some ideas:
- have install print to STDERR by default and write vars to STDOUT
- have install detect stdout is a pipe ?
- take it as a flag `—env-file .env` and still print with other stuff

api/README.md
```diff
@@ -2,12 +2,56 @@
...
+
+Where `<host>` is going to be the hostname or ip address of your local docker install. If you're on linux
+then its simply `localhost`. If you're on OS X, then it is one of:
+
+    - `docker-machine ip <machine>`
+    - `boot2docker ip`
+
+And the `PASSWORD` is in the ENV data in your `.env` file.
+
+### Building locally
```
ddollar: This header is confusing to me. Building a Rack? Maybe just add this to Prereqs.

csquared: I think it might be a little overwhelming to put it in Prequisites. What about changing it to `convox build`?

ddollar:
:+1:

PR #35, opened 2015-09-18
Fix link to cmd/crypt submodule

brandur:
Basic edit to make sure that the link to cmd/crypt is pointing at the right place.


csquared:
Thanks!

PR #36, opened 2015-09-18
Appname regex

mattmanning:
Update appname regex to match the one provided by AWS. Dashes are allowed, not underscores. Also updates the error message to explain in English. The regex is somewhat cryptic, but the rules are simple enough to explain in words.


csquared:
:+1:

PR #38, opened 2015-09-18
[wip] Convox init

csquared:
break `convox init` out of `convox start`

don't merge this before releasing!

let's follow this commit / PR up with better project detection and  a way to scale templates (which do we vendor, which do we pull in?)


csquared:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;G5noQ577&#x2F;154-convox-init"> convox init</a>

PR #40, opened 2015-09-18
Add mysql service template

ddollar:


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;F8hyoEtX&#x2F;160-add-mysql-service-template"> Add MySQL service template</a>

PR #49, opened 2015-09-23
Updated banner message in CLI pointing to right URL, since old one wa…

jjperezaguinaga:
…s broken

Noticed this during a new convox installation, this is related to https://github.com/convox/convox.github.io/issues/46.


ddollar:
Thanks!

PR #54, opened 2015-09-29
Continuous Integration

nzoschke:
Revive CircleCI Continuous Integration that installs convox, builds, deploys and tests and app, and uninstalls convox. Put all this behind a guard, so the ~20 minute test run only happens on demand via a CircleCI Paramaterized Build API call with the RUN_AWS variable.


ddollar:
:+1:

PR #73, opened 2015-10-07
Make CPU share scale with Memory

ddollar:
Should make CPU metric more meaningful

### DIFF
--- a/api/models/fixtures/web_external_internal.json
+++ b/api/models/fixtures/web_external_internal.json
@@ -500,7 +500,7 @@
               "BlankWebService",
               {
                 "CPU": {
-                  "Ref": "Cpu"
+                  "Ref": "WebMemory"
                 },
                 "Command": {
                   "Ref": "WebCommand"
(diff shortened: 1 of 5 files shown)


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;AT01A0No&#x2F;185-cpu-percentage-is-inaccurate"> CPU percentage is inaccurate</a>

ddollar:
This reflects the same behavior AWS uses for Lambda.

nzoschke:
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html

The Task Definition CPU setting is fairly arbitrary on how it affects running containers. It is not required to set at all, and only guides soft guarantees in the kernel and cgroups about how a container runs. What happens in effect is that all running containers compete to use the CPU they need irregardless of what this might set.

So this patch only tweaks how we can account for CPU usage. Before this patch we were using a constant number for all container tasks. Now we set the number proportional to the memory, which is commonly scaled up or down.

mattmanning:
:+1:

Issue #71, opened 2015-10-07
Allow building without cache


ddollar:
Add `--no-cache` options to build and deploy that cause the build to ignore the host's Docker cache.

ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;gDqEG9YJ&#x2F;176-build-without-caching"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Build without caching</a>

PR #74, opened 2015-10-07
Add --no-cache option to build and deploy

ddollar:
Allows use of these commands without the Docker cache

### DIFF
--- a/cmd/convox/start.go
+++ b/cmd/convox/start.go
@@ -110,7 +110,7 @@ func cmdStart(c *cli.Context) {
 		return
 	}
 
-	errors := m.Build(app, dir)
+	errors := m.Build(app, dir, true)
 
 	if len(errors) != 0 {
 		fmt.Printf("errors: %+v\n", errors)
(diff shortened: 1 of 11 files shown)


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;gDqEG9YJ&#x2F;176-allow-building-without-cache"> Allow building without cache</a>

ddollar:
Solves #71

cmd/convox/start.go
```diff
@@ -110,7 +110,7 @@ func cmdStart(c *cli.Context) {
 		return
 	}
 
-	errors := m.Build(app, dir)
+	errors := m.Build(app, dir, true)
```
csquared: could this be a flag to convox start as well ?

mattmanning:
:+1:

cmd/convox/start.go
ddollar: Yep

PR #83, opened 2015-10-09
SSL

nzoschke:
First pass at configuring SSL for an app. Any explicit `balancerPort:containerPort` mapping defined in the docker-compose.yml manifest can be configured to terminate SSL with the given certificate on the balancer, and forward to the container.

Create and deploy an app with 2 port mappings. We will turn 443 into an SSL listener.

```
$ cat docker-compose.yml 
main:
  build: .
  ports:
  - 5000:3000
  - 443:3000

$ convox deploy
...

$ convox apps info
Name       simple-rails
Status     running
Release    RHYSENRLDHK
Processes  main
Hostname   simple-rails-173749661.us-east-1.elb.amazonaws.com
Ports      main:443 main:5000

$ curl -s simple-rails-173749661.us-east-1.elb.amazonaws.com:443 | head -1
<!DOCTYPE html>

$ curl -s simple-rails-173749661.us-east-1.elb.amazonaws.com:5000 | head -1
<!DOCTYPE html>
```

List, create and delete SSL listeners with a port and certificate.

```
$ convox ssl
PORT  EXPIRES  DOMAINS

$ convox ssl create 443 foo.pub foo.key 
Creating SSL listener on port 443... Done.

$ convox ssl
PORT  EXPIRES            DOMAINS     
443   9 months from now  *.foobar.com

$ curl -vk -s https://simple-rails-173749661.us-east-1.elb.amazonaws.com:443/ | head -1
*   Trying 54.165.153.157...
* Connected to simple-rails-173749661.us-east-1.elb.amazonaws.com (54.165.153.157) port 443 (#0)
* TLS 1.2 connection using TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
* Server certificate: *.foobar.com
...
<!DOCTYPE html>

$ convox ssl delete 443
Deleting SSL listener on port 443... Done.

$ curl -vk -s https://simple-rails-173749661.us-east-1.elb.amazonaws.com:443/ | head -1
*   Trying 52.20.107.197...
* Connected to simple-rails-173749661.us-east-1.elb.amazonaws.com (52.20.107.197) port 443 (#0)
* Unknown SSL protocol error in connection to simple-rails-173749661.us-east-1.elb.amazonaws.com:-9847
* Closing connection 0
```

Create and delete validate against the manifest and CF Parameters:

```
$ convox ssl create 4443 foo.pub foo.key 
Creating SSL listener on port 4443... ERROR: Manifest does not specify balancer port 4443

$ convox ssl delete 4443
Deleting SSL listener on port 4443... ERROR: app does not have a certificate on balancer port 4443
```


ddollar:
Can you create an SSL listener on an internal port that is not exposed in the load balancer already?

Something like:

```
ports:
  - 80:3000
  - 4443

$ convox ssl create 4443 foo.pub foo.key
```

nzoschke:
Not with this patch. I know we discussed a few different ways port mappings are defined in docker-compose and how those could be configured for SSL, but I focused on the simplest case to start.

So for now you would have to update the manifest to say 4443:4443, deploy, then you can add the listener.

nzoschke:
Currently no, though that change shouldn't be too difficult to include in this PR.

Matt and I went back and forth about this... Is SSL something you configure one-off on your deployed app, or is it something you more explicitly map out in your docker-compose.yml manifest.

mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;G9kbtY9z&#x2F;107-ssl-certificates"> SSL certificates</a>

ddollar:
If you explicitly declare a `443:NNNN` in your app it means you required to have your app terminate SSL for itself which increases the complexity of using this a great deal. While I agree that it's a good idea I don't think it should be a requirement.

nzoschke:
You don't *have* to have your app terminate SSL... 443 can serve plain text until you correctly add the ssl cert.

```
main:
  build: .
  ports:
  - 5000:3000
  - 443:3000
```

api/controllers/ssl.go
```diff
@@ -0,0 +1,68 @@
...
+func SSLCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
+	a := mux.Vars(r)["app"]
+	port := GetForm(r, "port")
+	body := GetForm(r, "body")
+	key := GetForm(r, "key")
+
+	var validPorts = regexp.MustCompile(`[0-9]+`)
+
+	if !validPorts.MatchString(port) {
+		return httperr.Errorf(403, "balancer port must be in 443 format")
```
ddollar: This error message is a bit confusing. Maybe say "numeric" or something similar?

ddollar:
How do you know whether to configure the load balancer to look for encrypted or unencrypted traffic on the container port?

mattmanning:
We set up an SSL listener associated with the specified external port and a host port (which maps to a container port).

nzoschke:
It keeps the current behavior. Load balancer listerners are unencrypted by default. `convox ssl add` changes that listener to an SSL listener.

api/models/fixtures/web_external_internal.json
```diff
@@ -165,14 +178,32 @@
...
+                "LoadBalancerPort": {
+                  "Ref": "WebPort80Balancer"
+                },
+                "Protocol": "TCP"
+              },
+              {
+                "InstancePort": {
+                  "Ref": "WebPort80Host"
+                },
+                "InstanceProtocol": "TCP",
```
ddollar: Looks like it's being hardcoded to unencrypted container port right now which means rack and grid themselves wouldn't work with this.

ddollar:
If the app declares a 443 outside listener then I think it's fairly reasonable to assume that the listener on port 3000 is encrypted. The current behavior is a bit confusing to me and makes the docker-compose.yml confusing for an outside viewer. Why is this app listening on port 443 with HTTP?

```
ports:
  - 443:3000
```

mattmanning:
A couple of days ago I floated the idea of specifying ssl in `docker-compose.yml`, which I think is very clean and easy to understand, but it breaks compatibility with Compose itself:

```
ports:
  - 443:3000::ssl
```

ddollar:
That breaks compatibility with `docker-compose`. While we may do that at some point we should do it carefully and for a good reason and I don't think we need to here.

nzoschke:
Yeah I understand your confusion. Is anyone other than rack and grid actually doing SSL termination inside the container? 

For grid at least, I feel like we would do that not for a real development use case, but because we didn't have a good way to install a real certificate on the load balancer.

ddollar:
It's more about confusion when looking at a project's `docker-compose.yml`. If you saw that it was exposing 443 you would not expect to find HTTP there.

mattmanning:
Doesn't adding the listener require SSL? Can 443 still serve plain HTTP after it's added?

nzoschke:
At least I could see some indication that the app wants to do something SSL related. If you saw nothing but port 80 you have no information about SSL.

nzoschke:
Note: this design was also the first patch to land because its the easiest to implement.

nzoschke:
Finally, my pipe dream is that `convox start` can see 443:80 and sets up a dev SSL forwarder. A developer shouldn't have to do the HAProxy junk we do in convox/rack.

mattmanning:
FWIW, we've already broken compatibility with `docker-compose` with our environment handling, but I agree we should be careful about breaking it further.

mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;G9kbtY9z&#x2F;107-ssl-certificates"> SSL certificates</a>

api/models/fixtures/web_external_internal.json
nzoschke: Rack is managed by kernel.json.

Grid has a manually configured listener per the current [SSL setup instructions](http://convox.com/docs/custom-certificates/). When this patch lands add one more line to the grid manifest, deploy, and re-upload a cert.

```diff
$ git diff docker-compose.yml
diff --git a/docker-compose.yml b/docker-compose.yml
index 28948e2..f5e20b2 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -14,6 +14,7 @@ web:
     - GRID_HOST
   ports:
     - 80:5000
+    - 443:5000
     - 4443
   command: bin/web
   volumes:
```

nzoschke: And I will do the listener cleanup and migration of Grid when this patch lands.

ddollar: +1

mattmanning:
Documentation PR: https://github.com/convox/convox.github.io/pull/57

mattmanning:
:+1:

PR #96, opened 2015-10-13
add --stats to ps

ddollar:
* process api turns off stats if `?stats=false`
* add `--stats` to `convox ps`


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;B6rgDNyJ&#x2F;193-remove-cpu-memory-stats-from-convox-ps-by-default"> Remove CPU&#x2F;Memory stats from convox ps by default</a>

mattmanning:
:+1:

PR #100, opened 2015-10-14
show url(s) after deploy

csquared:
- refactor and fix build cmd to pipe stdout but not close
- print all open ports and balancer url

### DIFF
--- a/cmd/convox/builds.go
+++ b/cmd/convox/builds.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"io"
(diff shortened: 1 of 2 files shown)


nzoschke:
:+1: 

Using a pipe to indirect around closing is a pattern I've seen a few times now.

csquared:
yeah at first i was going to try changing the method signature to a Writer instead of a WriteCloser, but you need *one* side of the stream to be able to close to signal completion.

`Pipe` and `io.Copy` worked quite well together here

PR #107, opened 2015-10-14
better rolling of asgs

csquared:
simple first pass

- wait for resource signals, 15m timeout
- send signals after `cfn-init` completes

note: does not wait for ecs-agent or convox start to finish.

### DIFF
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -554,7 +554,7 @@
         "AutoScalingRollingUpdate": {
           "MaxBatchSize": 1,
           "MinInstancesInService": 1,
-          "PauseTime" : "PT5M",
+          "PauseTime" : "PT15M",
           "WaitOnResourceSignals": "true"
         }
       }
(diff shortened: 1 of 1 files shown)


csquared:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;JWMk511S&#x2F;188-better-rolling-of-autoscale-groups"> Better rolling of autoscale groups</a>

api/dist/kernel.json
```diff
@@ -551,7 +554,9 @@
       "UpdatePolicy": {
         "AutoScalingRollingUpdate": {
           "MaxBatchSize": 1,
-          "MinInstancesInService": 1
+          "MinInstancesInService": 1,
+          "PauseTime" : "PT15M",
```
ddollar: Does it roll faster than this if it receives the signal?

csquared: Yes that's just a timeout for receiving *all* the signals. We could even bump it to 20 or 30

nzoschke:
:+1:

Issue #65, opened 2015-10-03
Automatic Unhealthy Instance Termination


nzoschke:
A best practice cloud automation strategy is to implement instance level health checks and automatically terminate unhealthy instances.

We should consider implementing this around host level file system health to start.

I observed one production app process started throwing errors around a "no usable temporary directory".

Looking at the host, the container was running on a read-only filesystem

```
$ docker exec -it 5d0b49face0a sh
# touch /tmp/foo
touch: cannot touch '/tmp/foo': Read-only file system

# mount
/dev/mapper/docker-202:1-1048807-5d0b49face0a843673a2063fbbe0db2bdb0a788dc4f5872a06cd4e1c3e9ce54a on / type ext4 (ro,relatime,stripe=16,data=ordered)
```

`dmesg` on both the host and container showed filesystem errors:

```
# dmesg | grep dm-1
[749071.796012] Aborting journal on device dm-1-8.
[749100.844148] EXT4-fs error (device dm-1): ext4_journal_check_start:56: Detected aborted journal
[749100.847153] EXT4-fs (dm-1): Remounting filesystem read-only
[749100.849035] EXT4-fs (dm-1): ext4_writepages: jbd2_start: 1024 pages, ino 524322; err -30
```

I was not able to fully root cause the file system failure nor determine a good way to see the device mapper error or read-only state.

ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;1PS5DmJj&#x2F;183-automatic-unhealthy-instance-termination"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Automatic unhealthy instance termination</a>

gjohnson:
Another one we've hit was Error response from daemon: Cannot start container xxx: ... (cannot allocate memory). Though monitoring + autoscaling may have solved this one, can't remember.

nzoschke:
Some progress towards this goal is in https://github.com/convox/rack/pull/111

However I did not yet pull the trigger on putting the Docker-based checks into the production path, since unforseen problems here could potentially terminate an entire cluster. For example, what happens if Docker Hub is down when trying to run a health check.

So there is more TODO on this issue:

- [ ] Do the Docker checks concurrently
- [ ] Carefully handle errors in the Docker checks, i.e. is the problem Docker Hub or is it a host problem
- [ ] Require a certain number of failed checks before terminating
[...]

PR #111, opened 2015-10-17
Automatic Unhealthy Instance Termination Refactor

nzoschke:
https://github.com/convox/rack/issues/65

This is a nice refactor of the mission critical cluster monitoring worker that should be save to bring into master.

There was also a plan to add additional docker checks to this system, which is laying in wait but not turned on. See the issue for more details.

### DIFF
--- a/api/workers/cluster.go
+++ b/api/workers/cluster.go
@@ -32,41 +32,14 @@ func StartCluster() {
 
 		if err != nil {
 			log.Error(err)
+			continue
 		}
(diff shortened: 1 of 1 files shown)


nzoschke:
Updated the main issue with remaining work here. https://github.com/convox/rack/issues/65

I'd like to commit the safe parts of the refactor now, then regroup.

ddollar:
:+1:

PR #113, opened 2015-10-18
don't fail the system update if we can't determine concurrency

ddollar:


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;eIcVp6jl&#x2F;204-dont-fail-the-system-update-if-we-cant-determine-concurrency"> dont fail the system update if we cant determine concurrency</a>

PR #114, opened 2015-10-18
use segment for rack event tracking

ddollar:


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;CfvPHtgA&#x2F;205-use-segment-for-rack-event-tracking"> use segment for rack event tracking</a>

PR #115, opened 2015-10-19
Specify docker-compose.yml file

mattmanning:
This contains changes to the client and server which allow you to pass a `--file <filename>` option to `convox start`, `convox build`, and `convox deploy` if you want to use a file other than `docker-compose.yml`.  The option is never required and in all cases defaults to `docker-compose.yml`.

Fixes https://github.com/convox/kernel/issues/97

### DIFF
--- a/cmd/convox/start.go
+++ b/cmd/convox/start.go
@@ -26,6 +26,7 @@ func init() {
 		Flags: []cli.Flag{
 			cli.StringFlag{
 				Name:  "file",
+				Value: "docker-compose.yml",
 				Usage: "a file to use in place of docker-compose.yml",
 			},
 		},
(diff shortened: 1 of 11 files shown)


mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;MXDqUXxi&#x2F;169-specify-docker-compose-yml-file-for-start-and-build"> Specify docker-compose.yml file for start and build</a>

cmd/convox/start.go
```diff
@@ -23,6 +23,12 @@ func init() {
 		Description: "start an app for local development",
 		Usage:       "[directory]",
 		Action:      cmdStart,
+		Flags: []cli.Flag{
+			cli.StringFlag{
```
csquared: you can specify a deafult value in this block so you don't need to check it against "" later

mattmanning: Good tip, thank you.

cmd/convox/start.go
```diff
@@ -23,6 +23,13 @@ func init() {
...
+				Name:  "file",
```
csquared: `"file, f"` should take care of the shorthand version `-f` too

Issue #109, opened 2015-10-15
Convox CLI: Wrong URL For AMI User Creation


BenjaminDavison:
"ben1$ convox install --region eu-west-1 --ami ami-b96f59ce --key foo

     ___    ___     ___   __  __    ___   __  _
    / ___\ / __ \ /  _  \/\ \/\ \  / __ \/\ \/ \
   /\ \__//\ \_\ \/\ \/\ \ \ \_/ |/\ \_\ \/>  </
   \ \____\ \____/\ \_\ \_\ \___/ \ \____//\_/\_\
    \/____/\/___/  \/_/\/_/\/__/   \/___/ \//\/_/

This installer needs AWS credentials to install the Convox platform into
your AWS account. These credentials will only be used to communicate
between this installer running on your computer and the AWS API.

We recommend that you create a new set of credentials exclusively for this
install process and then delete them once the installer has completed.

To generate a new set of AWS credentials go to:
https://docs.convox.com/docs/creating-an-iam-user"

The URL leads to a redirect http://convox.com/docs/docs/creating-an-iam-user (note extra docs/) which 404's

Thanks,

Ben

csquared:
this has been fixed. thanks guys!

PR #120, opened 2015-10-20
Documentation fix for #109

joonathan:
Redirection from https://docs.convox.com adds /docs/ to path, so removed the additional docs from the path so the URL wouldn't 404.


ddollar:
Thanks!

PR #122, opened 2015-10-21
show region after install

aaronfeng:
Small change so people aren't confused :)


csquared:
hey @aaronfeng - thanks for the PR but we're gonna try to keep people *away* from the AWS console as much as possible for now.

PR #126, opened 2015-10-22
better deploy messaging

bobzoller:
when `convox deploy` is creating the tarball, it now says "Creating tarball... OK".  then when it uploads the tarball it says "Uploading... OK".

should reduce confusion from folks on slow connections.


ddollar:
Thanks!

PR #130, opened 2015-10-22
Webhooks Service

csquared:
creates a service for http(s) webhooks backed by SNS

`convox services add webhook --url <url>`

creates a handler for Custom::SNSSubscription that allows us to dynamically hook up an SNS subscription

updates kernel.json to have an SNS topic for rack event notifications
creates `Notify`, `NotifySuccess`, `NotifyError` for sending outbound notifications to that SNS topic

webhook service uses the Custom Handler to install *itself* as an https endpoint for SNS. it then forwards the SNS message to the `<url>` configured in the service.  this allows us to control the message payloads.


api/cmd/formation/handler/sns.go
```diff
@@ -0,0 +1,93 @@
...
+		return "failed", nil, err
+	}
+
+	outputs := make(map[string]string)
+	outputs["Endpoint"] = endpoint
+
+	return endpoint, outputs, nil
+}
+
+func SNSSubscriptionUpdate(req Request) (string, map[string]string, error) {
```
ddollar: look into signaling lack of support in update

api/dist/kernel.json
```diff
@@ -68,6 +69,10 @@
     "Kinesis": {
       "Value": { "Ref": "Kinesis" }
     },
+    "NotificationTopic": {
```
ddollar: need notification host here too

PR #142, opened 2015-10-27
allow secure listeners for ssl

ddollar:
Allows the user to specify that the listener behind the SSL is itself TLS-encrypted.

```
$ convox ssl create web:443 --self-signed --secure
Creating SSL listener web:443... Done.

$ convox ssl
TARGET   EXPIRES          SECURE  DOMAIN
web:443  1 year from now  true    sinatra-1477412079.us-east-1.elb.amazonaws.com
```


ddollar:
Test as `20151027150947-secure-ssl`

PR #158, opened 2015-11-10
fix circle

csquared:
- removes "|| true" in places you need to see failures
- uses circle branch whitelisting insteaad of ENV var
  to only run commits to master


ddollar:
:+1:

Issue #55, opened 2015-09-29
Convox logs dies after a while


jordan-thoms:
Whenever I use the `convox logs` command, it exits after some period of time (sometimes a minute or two, sometimes longer). When it exits, it doesn't produce any error messages.

I assume that the websocket is getting disconnected?

ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;kTptzpQw&#x2F;179-convox-logs-disconnection"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> convox logs disconnection</a>

nzoschke:
This was improved in https://github.com/convox/rack/pull/160

PR #160, opened 2015-11-11
fix convox logs

csquared:
- fixes logging disconnect issue #55
- fixes goroutine leak for apps logs
- fixes goroutine leak for service logs

- extends IdleTimeout to 3600 for convox rack ELB
- extends IdleTimeout to 3600 for all convox app ELB(s) (one of which is grid)


csquared:
it doesn't appear to me we're exposing the services logs at the CLI, is that right?

mattmanning:
:+1:

PR #163, opened 2015-11-13
make the list of container ports unique

ddollar:
This prevents the doubling up of ContainerPorts when two external ports map to the same internal port. This prevents the ECS service from being deleted on every deploy because this list of container ports incoming (with dupes) mismatches the existing container ports (uniqued)


csquared:
:+1:

PR #175, opened 2015-11-24
exclude .dockerignore files from the build tarball

### LINKED ISSUE #135: Support excluding certain files from a release (.dockerignore)
I have files that are not required for the build and can grow quite large over time. It really slows down deploys. It would be helpful if there were a way to ignore these files and exclude them from the upload. 

Maybe supporting `.dockerignore` files is the most sensible approach, given that anything in there would be ignored at build time anyway it should be safe not to include them in the release. 

http://docs.docker.com/reference/builder/#dockerignore-file

mattmanning:
If a `.dockerignore` file is present, use it to exclude files from the build tarball.

Fixes #135

### DIFF
--- a/cmd/convox/builds.go
+++ b/cmd/convox/builds.go
@@ -225,7 +225,16 @@ func createTarball(base string) ([]byte, error) {
 		return nil, err
 	}
 
-	cmd := exec.Command("tar", "cz", ".")
+	args := []string{"cz"}
+
+	// If .dockerignore exists, use it to exclude files from the tarball
+	if _, err = os.Stat(".dockerignore"); err == nil {
+		args = append(args, "--exclude-from", ".dockerignore")
+	}
+
+	args = append(args, ".")
+
+	cmd := exec.Command("tar", args...)
 
 	out, err := cmd.StdoutPipe()
 


mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Z0sxoXf6&#x2F;213-dont-upload-ignored-files-during-build"> dont upload ignored files during build</a>

cmd/convox/builds.go
```diff
@@ -225,7 +225,16 @@ func createTarball(base string) ([]byte, error) {
 		return nil, err
 	}
 
-	cmd := exec.Command("tar", "cz", ".")
+	args := []string{"cz"}
+
+	// If .dockerignore exists, use it to exclude files from the tarball
+	if _, err = os.Stat(".dockerignore"); err == nil {
+		args = append(args, "--exclude-from", ".dockerignore")
```
csquared: well dang, that's convenient!

csquared:
+1

cmd/convox/builds.go
shayne: Not sure this supports the following .dockerignore ruleset:
```
*
!build/
!docker-compose.yml
```

PR #186, opened 2015-11-30
Build.IsLocked tests for active docker containers named build-B*. Bui…

nzoschke:
…ldCreate controller test for lock and returns 403

### DIFF
--- a/api/controllers/builds.go
+++ b/api/controllers/builds.go
@@ -66,6 +66,10 @@ func BuildGet(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 func BuildCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	build := models.NewBuild(mux.Vars(r)["app"])
 
+	if build.IsLocked() {
+		return httperr.Errorf(403, "another build is currently running. Please try again later.")
+	}
+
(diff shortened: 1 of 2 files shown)


nzoschke:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;pLX29Kdp&#x2F;234-improve-locking-around-builds-and-agent-image-cleaning"> Improve locking around builds and agent image cleaning</a>

csquared:
LGTM

api/controllers/builds.go
```diff
@@ -66,6 +66,10 @@ func BuildGet(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 func BuildCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	build := models.NewBuild(mux.Vars(r)["app"])
 
+	if build.IsLocked() {
```
csquared: for me, a `lock` would imply actually setting and unsetting a lock. this check seems here to be more around if it is `busy`, `running`, etc. so I think there's a handful of better names than `IsLocked` for this one.  More to consider than a must change.

nzoschke: renamed to IsRunning. The comments were already in that vein

PR #196, opened 2015-12-02
downcase dir name when deriving app name from it

### LINKED ISSUE #105: convox start fails on folders with capitalization
report from joonathan via slack:

> If folder name contains capitalization eg. `Telebot` then docker-compose itself is successfuly able to deal with that, but convox will fail
> convox start would fail, convox apps create would ​_succeed_​ but failure would come when doing convox deploy as image is unable to build due to the name issue

mattmanning:
Fixes #105


mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;cGnDdUxw&#x2F;217-folders-with-captialization"> folders with captialization</a>

PR #197, opened 2015-12-02
GetRackSettings and PutRackSettings to securely get and set sensitive credentials for the rack

nzoschke:
This can be used to store things like private docker registry credentials, and an ssh private key.

/cc @csquared


mattmanning:
:+1:

PR #198, opened 2015-12-03
ssh to instances

csquared:
The purpose of this command is to be able to troubleshoot an instance. You can get in there, run some diagnostics, and root cause your actual problem. Having Rack credentials gives full access to the infrastructure, so this is just a way to not have to download and use SSH keys in addition to the rack parameter.

Requires CloudFormation parameter `Key` to be set and be the name of a valid EC2 KeyPair. This allows amazon to manage ensuring the public key is installed on all instances for ssh.

This PR provides a command to generate an EC2 KeyPair, store it in S3 w. KMS encryption, and update the CF parameter `Key` via:

    $ convox instances keyroll
    Rebooting instances...

After you run this command you will have fresh EC2 KeyPar set on all your instances (which is what requires the roll).  You can run it whether not `Key` is set, and it does not delete the old Key in case you need it for forensics.

To actually ssh to a machine we use the Rack credentials to tunnel the SSH session over a websocket via:

    $ convox instances ssh <id> [command]

For example:

```bash
$ convox instances ssh i-9ca39d22

   __|  __|  __|
   _|  (   \__ \   Amazon ECS-Optimized Amazon Linux AMI 2015.03.e
 ____|\___|____/

For documentation visit, http://aws.amazon.com/documentation/ecs
Amazon Linux version 2015.09 is available.
[ec2-user@ip-10-0-1-172 ~]$
```

There are a few warts around closing STDIN when it is piped in but `convox exec` and `run` suffer from similar issues.  I've been more surprised at how many things work than don't.

Fun commands:

    $ convox instances ssh <id>
    [ec2-user@ip-10-0-1-172 ~]$ vi

    $ convox instances ssh <id> docker run --rm -it networkstatic/star_wars

In the name of immutable architecture, I'd love to see an instance marked as "dirty" once someone has opened a shell with it.
Maybe that marks it as first in line for termination by a Chaos Monkey...

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -243,7 +243,7 @@
 			"Rev": "74f810a0152f4c50a16195f6b9ff44afc35594e8"
 		},
 		{
-			"ImportPath": "golang.org/x/crypto/ssh/terminal",
+			"ImportPath": "golang.org/x/crypto/ssh",
 			"Rev": "74f810a0152f4c50a16195f6b9ff44afc35594e8"
 		},
 		{
--- a/Godeps/_workspace/src/golang.org/x/crypto/LICENSE
+++ b/Godeps/_workspace/src/golang.org/x/crypto/LICENSE
@@ -1,0 +1,27 @@
+Copyright (c) 2009 The Go Authors. All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are
+met:
+
+   * Redistributions of source code must retain the above copyright
+notice, this list of conditions and the following disclaimer.
+   * Redistributions in binary form must reproduce the above
+copyright notice, this list of conditions and the following disclaimer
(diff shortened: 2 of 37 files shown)


csquared:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;GlqMwDZ7&#x2F;243-ssh-to-instances"> SSH to instances</a>

csquared:
note: have to update the swagger.yml

mattmanning:
:+1: Nice work. I agree with your decision to punt on the STDIN issues. This looks like a complete enough solution for debugging.

csquared:
actually none of the other websocket endpoints are in the swagger, but added entries for `POST /instances/keyroll` and `DELETE /instances/{id}`

csquared:
````
$ cx instances ssh i-9ca39d22 exit 10
$ echo $?
10
```

PR #202, opened 2015-12-08
WIP: `tar c . | gzip -f` is more portable?

nzoschke:
Todo:

- [ ] Review error detection and handling on the client `tar cz`
- [ ] Review error detection and handling on the API build endpoint


csquared:
looks good -> perhaps just look for an error on the calls to `Wait()`, return those, and call it a day?

PR #204, opened 2015-12-10
RDS options

mattmanning:
This adds 5 new options to service creation. Currently the new options are only applied in the Postgres template. The options are:

* --size
* --instance-type
* --version
* --multi-az
* --encrypted

### DIFF
--- a/api/controllers/services.go
+++ b/api/controllers/services.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/convox/rack/Godeps/_workspace/src/github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/convox/rack/Godeps/_workspace/src/github.com/gorilla/mux"
 	"github.com/convox/rack/Godeps/_workspace/src/golang.org/x/net/websocket"
 	"github.com/convox/rack/api/httperr"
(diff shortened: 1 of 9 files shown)


mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;a2v0B9DY&#x2F;211-rds-postgres-options"> RDS Postgres Options</a>

csquared:
I'm not a huge fan of how this makes every service (papertrail and redis right now) get these options. Is there a way to do store the service-specific config another way or did that prove really difficult?  

I'm thinking something like a `map[string]string` called config or even doing something with subtypes and embedded types?

ddollar:
I second the map for config map here.

nathany:
Supporting multi-az out of the box will be great. Thanks.

That is one parameter that is a boolean rather than a string. How well will `map[string]string` work for parameters of different types?

csquared:
@nathany the parsing and validation will still need to happen, but can happen on the server side. you could imagine a `Validate()` method that builds up the proper RDS config struct and does the conversions from strings to the proper types, returning an error if any of the conversions fail. But that's really just one way in a sea of infinity so didn't want to get too prescriptive.

(Also, if we have that method, it means you can safely persist the service as strings b.c the conversion still works.  Another different way would be to make a struct for each service so that you don't need to re-convert, but once you've separated the notion of "conversion" and "validation" then what we're really saying is we care more about the validation then exactly how its achieved, and we definitely want to validate data *before* we store it)

nathany:
Thanks for clarifying. In that case, a map is certainly simpler than many little structs.

PR #209, opened 2015-12-11
log into Grid without a valid rack; switch to a valid rack

mattmanning:
* Add `/auth` func to client and call it if proxying to Rack `/apps` fails.
* Change `/switch` to not call `/system` (aka versionCheck). This works around Rack errors so users in a bad state to switch to a known good Rack.

This change enables users who can't log into Grid via the CLI because their current Rack is bad or nonexistent to get back to a good state. Once this is released they can:

1. `convox update`
2. `convox login grid.convox.com`
3. `convox switch [good rack]`

Depends on https://github.com/convox/grid/pull/107

### DIFF
--- a/api/controllers/auth.go
+++ b/api/controllers/auth.go
@@ -1,0 +1,12 @@
+package controllers
+
+import (
+	"net/http"
+
+	"github.com/convox/rack/api/httperr"
+)
+
+func Auth(w http.ResponseWriter, r *http.Request) *httperr.Error {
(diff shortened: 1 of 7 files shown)


mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;kywbtlbi&#x2F;54-can-convox-login-grid-convox-com-without-a-current-rack"> can `convox login grid.convox.com` without a current rack</a>

PR #213, opened 2015-12-14
Load Balance all the ports

csquared:
now every Process with open ports gets one load balancer with listeners
for each port

if all of the ports are of the form `<port>:<port>` the load balancer is
internet-facing, otherwise it is an internal load balancer


csquared:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;hJ0HGNvv&#x2F;244-internal-load-balancers"> internal load balancers</a>

api/models/fixtures/web_external_internal.json
```diff
@@ -493,7 +591,11 @@
       "Type": "AWS::S3::Bucket"
     },
     "WebECSService": {
-      "DependsOn": "CustomTopic",
+      "DependsOn": [
+        "BalancerWebInternal",
+        "BalancerWebInternal",
```
nzoschke: Why is this repeated twice?

api/models/helpers.go
```diff
@@ -373,81 +375,6 @@ func templateHelpers() template.FuncMap {
...
-				},
-				"Links": [ %s ],
-				"Volumes": [ %s ],
-				"Services": [ %s ],
-				"PortMappings": [ %s ]
-			}, { "Ref" : "AWS::NoValue" } ] }`, UpperName(ps), ps, UpperName(ps), UpperName(ps), UpperName(ps), strings.Join(envs, ","), strings.Join(links, ","), strings.Join(volumes, ","), strings.Join(services, ","), strings.Join(mappings, ","))
-
-			return template.HTML(l)
-		},
 		"ingress": func(m Manifest) template.HTML {
```
nzoschke: Can we get rid of other helpers too? A quick check shows that "ingress" is dead code too.

nzoschke:
:+1: LGTM!

The manifest fixture diffs are very enlightening, and give me a lot of confidence in the end results of these changes.

nzoschke:
This didn't end up needing any changes to the lambda handler?

csquared:
negative on the lambda handler - our logic was right. adding the right `DependsOn` for the balancer for each ECS service fixed the issue where the service would throw an error

csquared:
new deploy message:

```term
Pushing tag for rev [d2b6b86d706f] on {https://convox-2033058839.us-east-1.elb.amazonaws.com:5000/v1/repositories/dungeon-mud/tags/BHXGXQCXEKX}
Promoting RRTXPAZMWSD... UPDATING

Use `convox apps info -a dungeon` for available endpoints.
Use `convox ps -a dungeon` for rolling deploy status.
```

ddollar:
Can we just kill the message after "Promoting..."

The endpoints still won't be accurate and we should find a better way to surface deploy status (Noah's work is a good step here)

PR #215, opened 2015-12-15
remove references to "latest" version

csquared:
"latest" was deprecated as a version ->
dont install or reference it from anywhere


nzoschke:
:+1: 

This might cause some surprises on development environment setup, but I added some documentation to help there. A surprise on setup is much better than a surprise later in development.

There will be more follow up work to better support development of kernel.json and contraption changes for folks that can't publish branch releases and set a custom VERSION.

PR #219, opened 2015-12-16
us the ami ids from the ECS docs, not the aws marketplace

csquared:
use:
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_container_instance.html

not:
screen shot 2015-12-15 at 6 37 08 pm


nzoschke:
These match up with:
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_container_instance.html

PR #227, opened 2015-12-18
Service Options (was RDS options)

csquared:
- parses any cli option passed to service create
- passes options through to CF template
- reports errors from CF so we don't need to validate presence, etc in models
- adds status reason to service model
- abstracts service creation to hopefully make definition, addition easier
- changes "URL" return to "exports" key/value pairs for future compatibility

- fixes service tagging as well

### DIFF
--- a/api/models/service.go
+++ b/api/models/service.go
@@ -83,24 +83,15 @@ func (s *Service) Create() error {
 		return err
 	}
 
-	params := make(map[string]string)
-	for key, value := range s.Options {
-		var val string
-		switch value {
-		case "":
-			val = "No"
-		case "true":
-			val = "Yes"
-		default:
-			val = value
-		}
-		params[AwsCamelize(key)] = val
+	// pass through service parameters as Cloudformation Parameters
+	for key, value := range s.Parameters {
+		req.Parameters = append(req.Parameters, &cloudformation.Parameter{
+			ParameterKey:   aws.String(key),
+			ParameterValue: aws.String(value),
+		})
 	}
 
-	for key, value := range params {
-		req.Parameters = append(req.Parameters, &cloudformation.Parameter{ParameterKey: aws.String(key), ParameterValue: aws.String(value)})
-	}
-
+	// tag the service
 	tags := map[string]string{
 		"System":  os.Getenv("RACK"),
 		"Type":    "service",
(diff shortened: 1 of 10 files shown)


csquared:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;LpbC1PQf&#x2F;223-isolate-services-so-they-are-only-visible-to-the-rack-that-created-them"> Isolate services so they are only visible to the Rack that created them</a>

csquared:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;a2v0B9DY&#x2F;211-rds-postgres-options"> RDS Postgres Options</a>

csquared:
```term
$ cx services create webhook
Creating webhook-8368 (webhook)... ERROR: Options: [Url] must have values
$ cx services create webhook --url foo --foo
Creating webhook-3651 (webhook: url="foo" foo="true")... ERROR: Options: [Foo] are not supported by this service
```

csquared:
```term
$ cx services info webhook-5771
Name    webhook-5771
Status  failed
Reason  Failed to create resource. InvalidParameter: Invalid parameter: Endpoint
	status code: 400, request id: 3a069bc0-8a61-5312-8ced-d52dd1eaffd8
```

```term
$ cx services info postgres-2991
Name    postgres-2991
Status  failed
Reason  DB Instance class db.t2.micro does not support encryption at rest
```

api/models/service.go
```diff
@@ -35,7 +27,7 @@ func ListServices() (Services, error) {
 	for _, stack := range res.Stacks {
 		tags := stackTags(stack)
 
-		if tags["System"] == "convox" && tags["Type"] == "service" {
+		if tags["System"] == os.Getenv("RACK") && tags["Type"] == "service" {
```
nzoschke: Nice fix. Though this might cause trouble on some non-standard racks. Like demo...

cmd/convox/stdcli/stdcli_test.go
```diff
@@ -0,0 +1,22 @@
+package stdcli_test
+
+import (
+	"testing"
+
+	"github.com/convox/rack/Godeps/_workspace/src/github.com/stretchr/testify/assert"
+	"github.com/convox/rack/cmd/convox/stdcli"
+)
+
+func TestParseOptions(t *testing.T) {
```
nzoschke: nice

nzoschke:
:+1:

mattmanning:
:+1:

PR #228, opened 2015-12-18
notify on de-converge too

nzoschke:


csquared:
Is this a `Success`? I would assume this would be the error or warning side. Perhaps we need a `NotifyWarning`

csquared:
Fuck it, ship it

csquared:
Its not clear this is an error, so let's stick with calling it success

PR #230, opened 2015-12-18
pass aws message back with some formatting

csquared:
would rather pass back the messages than have a custom one


nzoschke:
:+1:

PR #237, opened 2015-12-23
Fixes #234 -- use crypto/rand in generateId instead of math/rand

### LINKED ISSUE #234: generateId() should use crypto/rand (at least in the cases where it's used to generate a password)
https://github.com/convox/rack/blob/master/api/models/helpers.go#L113-L119

In some cases (for datastores) thsi is used to generate a password. Therefore it should use `crypto/rand`.

alex:
In at least a few callsites generateId is used for generating a password, which really should be done from crypto/rand

### DIFF
--- a/api/models/helpers.go
+++ b/api/models/helpers.go
@@ -8,6 +8,7 @@ import (
 	"html/template"
 	"io"
 	"io/ioutil"
+	"math/big"
 	"os"
 	"regexp"
 	"strings"
@@ -109,7 +110,7 @@ var idAlphabet = []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ")
 func generateId(prefix string, size int) string {
 	b := make([]rune, size)
 	for i := range b {
-		idx, err := rand.Int(rand.Reader, big.NewInt(len(idAlphabet)))
+		idx, err := rand.Int(rand.Reader, big.NewInt(int64(len(idAlphabet))))
 		if err != nil {
 			panic(err)
 		}
--- a/api/models/helpers.go
+++ b/api/models/helpers.go
@@ -2,12 +2,12 @@ package models
 
 import (
 	"bytes"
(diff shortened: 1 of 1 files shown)


api/models/helpers.go
```diff
@@ -113,7 +110,11 @@ var idAlphabet = []rune("ABCDEFGHIJKLMNOPQRSTUVWXYZ")
 func generateId(prefix string, size int) string {
 	b := make([]rune, size)
 	for i := range b {
-		b[i] = idAlphabet[rand.Intn(len(idAlphabet))]
+		idx, err := rand.Int(rand.Reader, big.NewInt(int64(len(idAlphabet))))
+		if err != nil {
+			panic(err)
```
alex: I don't love this, but in practice I think it's vanishingly rare; would you prefer I added the possibility of the error return to the signature and altered the callers to propogate?

Could also introduce a new generateSecurePassword function, but I think that's more trouble than it's worth given that there's no harm in using a CSPRNG everywhere.

csquared:
Thanks for the PR! Found another `generateId` for ya: https://github.com/convox/rack/blob/master/api/cmd/formation/handler/ecs.go#L427-L435

As far as returning the error, my guideline has been to do *something* explicit with it (which the panic counts as), with a preference for bubbling it up. Honestly I see both sides (bubble up vs panic) as pretty equally valid here, so I defer to the person making the PR in those cases.  Just don't assign it to `_` :)

alex:
@csquared That appears to have exactly one caller which isn't sensitive to predictability (it's append to `ServiceName` on an ECS Service). Though it would be nice to just delete that code and have it use the other implementation... that belongs in a separate PR though IMO.

csquared:
@alex Sounds good, going to merge as is and we can address everything else with future PRs. Great catch :+1:

PR #238, opened 2015-12-23
less app level parameters

csquared:
CloudFormation has a hard limit of 60 parameters.
Each app has five params and each load balancer has four. There are also nine app-level params.

That means each app's formation is bound by the inequality:

`5p + 4b + 9 <= 60`

This PR gets us to:

`2p + 4b + 9 <= 60`

more than doubling the max number of processes :)
by hardcoding 2 of the parameter where they are used in the template
and dropping one that isn't used (but we may need later).

Able to drop "Image" and "CommandString" because those are set and not modifiable later (except by hand).

One could construct an argument this is good b.c it would
be pretty non-obvious that these have been changed and lead to surprising behavior.

Able to drop "{{entry.Name}}Service" because I couldn't it find it being set anywhere (could use a
second pair of eyes here) so its always blank.

I left around the condition so all the `Fn::If` clauses that look for it didn't need to be modified.

We need memory and desiredcount for scaling. These also make sense to parameterize.


csquared:
looking for some feedback before updating the test templates

mattmanning:
Thanks for the thorough PR description. This all makes sense to me.

PR #239, opened 2015-12-28
Fetch the convox upstart conf over HTTPS, not HTTP

alex:


nzoschke:
Nice security improvement. Thanks!

PR #242, opened 2016-01-04
Use HTTPS url for accessing kernel.json

alex:
Also factor the URL out into a const.


ddollar:
Merged in #251, thanks!

PR #248, opened 2016-01-12
Include ecr:GetAuthorizationToken permission

xtfxme:
This isn't an integration with ECR, it simply allows ECS to pull images from ECR. This is important for continuous integration of a local convox build chain (we deploy automated convox builds to ECR so CloudFormation is able to boot them).

This change allows ECR to be used with `convox/api` and `convox/registry` images.

### DIFF
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -453,6 +453,10 @@
                   "Action": [
                     "autoscaling:DescribeAutoScalingInstances",
                     "autoscaling:SetInstanceHealth",
+                    "ecr:GetAuthorizationToken",
+                    "ecr:GetDownloadUrlForLayer",
+                    "ecr:BatchGetImage",
+                    "ecr:BatchCheckLayerAvailability",
                     "ec2:DescribeInstances",
                     "ecs:CreateCluster",
                     "ecs:DeregisterContainerInstance",


ddollar:
Do we need the other ecr: permissions as well?

http://docs.aws.amazon.com/AmazonECR/latest/userguide/ECR_on_ECS.html

xtfxme:
We could add the additional permissions here too I suppose. We are adding these permissions via a policy on the repositories:

```
ecr:GetDownloadUrlForLayer
ecr:BatchGetImage
ecr:BatchCheckLayerAvailability
```

I'll try a deploy with those permission and no policy, if it works I'll update the PR. Sound good?

ddollar:
Sounds good :)

xtfxme:
Adding the read permissions explicitly works fine. It also enables repositories without explicit `ALLOW` policies to be used (the default). It should be noted that you still cannot boot or use `FROM` with ECR-based images yet... ECS (and convox in future) uses `ecr:GetAuthorizationToken` to generate docker credentials that last 12 hours. Since these creds are not persistent you can't really add them manually as a private registry.

PR updated to include both auth and read permissions.

ddollar:
Thanks!

PR #256, opened 2016-01-14
SSL chain fix

mattmanning:
Automatic intermediate cert chain generation was causing problems with the `convox ssl create` command for:

* Manually-created self-signed certs
* Self signed-certs created with the `--self-signed` option
* Any other cert with the full chain already bundled

The fixes for these were:

* Detect if the cert is self-signed, and if so, don't try to complete the chain and don't upload a chain to AWS
* For non-self-signed, strip away everything except the end-entity cert and then regenerate the chain

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ ENV GOPATH /go
 ENV PATH $GOPATH/bin:$PATH
 
 RUN apk-install gcc libc-dev libtool libgcc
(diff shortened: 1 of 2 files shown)


mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;2hq9F7z9&#x2F;303-fixes-for-convox-ssl-create"> fixes for `convox ssl create`</a>

ddollar:
:+1:

PR #258, opened 2016-01-14
update to latest ecs amis

ddollar:


api/dist/kernel.json
```diff
@@ -11,11 +11,13 @@
   },
   "Mappings": {
     "RegionConfig": {
-      "us-east-1": { "Ami": "ami-6ff4bd05" },
-      "us-west-1": { "Ami": "ami-46cda526" },
-      "us-west-2": { "Ami": "ami-313d2150" },
-      "eu-west-1": { "Ami": "ami-8073d3f3" },
-      "ap-northeast-1": { "Ami": "ami-6ca38b02" },
+      "us-east-1": { "Ami": "ami-2b3b6041" },
```
alex: Does it make sense to add a `"Comment": "<name of this AMI>"`, so that it's easy to tell if it's up to date from reading this?

PR #263, opened 2016-01-17
ssl update

### LINKED ISSUE #131: Update SSL certificates
Need `convox ssl update`

mattmanning:
This adds a `convox ssl update` command. It uploads the new cert and key, does a zero-downtime hot swap of the cert on the listener, updates the CF cert param, and deletes the old cert.

Fixes #131


mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;r2t7WDzp&#x2F;206-convox-ssl-update"> convox ssl update</a>

api/models/fixtures/multi_balancer.json
```diff
@@ -91,10 +91,28 @@
         "Ref": "WebPort80Balancer"
       }
     },
+    "WebPort80BalancerName": {
+      "Value": {
+        "Ref": "AWS::StackName"
```
nzoschke: could/should these be Fn::GetAtt on the Balancer resources?

api/models/ssl.go
```diff
@@ -273,14 +234,136 @@ func ListSSLs(a string) (SSLs, error) {
...
+		_, err = ELB().SetLoadBalancerListenerSSLCertificate(input)
+
+		if err != nil {
+			return err
+		}
+
+		return nil
+	}
+
+	err = backoff.Retry(operation, backoff.NewExponentialBackOff())
```
nzoschke: this looks neat! does it eventually give up and error?

api/models/ssl.go
```diff
@@ -273,14 +234,136 @@ func ListSSLs(a string) (SSLs, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	ssl := SSL{
+		Port:    port,
+		Process: process,
+	}
+
+	return &ssl, nil
```
nzoschke: how long does this functionl take generally?

api/models/fixtures/multi_balancer.json
nzoschke: Ah I see how it's reusing {{ $balancer.LoadBalancerName }}.

I think it'd be a bit more "correct" to use Fn::GetAtt DNSName or similar to generate an output. Notice how all the other outputs are intrinsic functions (Fn::GetAtt or Ref).

But I also don't see an ELB GetAtt that just gives the name of the load balancer. Can it be inferred from the DNSName?

http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-getatt.html

api/models/ssl.go
mattmanning: In my testing the whole update process took 3 or 4 seconds. I think it could be a little longer on a balancer that was serving heavy traffic and had to flush all the requests associated with the old cert before allowing a delete.

api/models/ssl.go
mattmanning: Yes. If I'm understanding correctly, the default exponential algorithm I've used here has a max time of 15 minutes: https://github.com/cenkalti/backoff/blob/master/exponential.go#L8-L94

It's possible to tweak the settings to achieve different curves and maxes.

api/models/fixtures/multi_balancer.json
mattmanning: I've looked into this and the balancer name is not something that can be fetched directly. http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-elb.html

It looks like extracting it from the DNSName will require regular expressions. You can't just split on hyphens because it could be multiple words or the app name could have a hyphen in it. I'm not sure if it's possible to use a regex in this way, and it's not clear to me from the docs. Do you know?

api/models/ssl.go
mattmanning: I just pushed a commit to use 60s as the max time to retry updating the cert on the listener, so this command should never take longer than 60s to return, even if it fails.

nzoschke: Awesome.

api/models/fixtures/multi_balancer.json
nzoschke: If DNSName to ELB name is not straight foward, then your strategy of copying the LoadBalancerName generator for the output is good. Thanks for looking into it.

nzoschke:
:+1:

api/models/fixtures/multi_balancer.json
mattmanning: :+1:

matthewford:
Could not get this to work, had no app called certs when I tried to specify the app flag. In the end I removed/added SSL manually.

mattmanning:
@matthewford Sorry to hear that. I will contact you in Slack.

PR #268, opened 2016-01-19
convox container linking

csquared:
docker has deprecated its links implementation. this is where we pick up the ball and use the syntax for env-var injection. this PR combines the standard `docker-compose.yml` link syntax with images that set env vars `LINK_USERNAME`, `LINK_SCHEME`, `LINK_PATH`, and `LINK_PASSWORD` . convox picks up those values from the linked image and combines them with the declared ports and known hosts (DOCKER_HOST in local dev and the appropriate load balancer in production) to assemble a link into a URL. when you link to an image named `redis`, you should get the appriopriate `REDIS_URL`.

given this `docker-compose.yml`:

```yml
web:
  build: .
  command: node server.js
  environment:
  ports:
    - 5000:8080
  links:
    - redis

worker:
  command: node worker.js
  environment:
    - REDIS_URL
  build: .
  links:
    - redis

redis:
  image: convox/redis
  ports:
    - 6379
```

you can see the web process doesn't declare its dependency on `REDIS_URL`, but it still gets a value for it. in the worker process i decalre the dependency, which gets picked up and set.

```bash
web    | running: docker run -i --name node-example-web -e REDIS_URL=redis://:password@172.17.0.1:6379/0 -p 5000:8080 node-example/web sh -c node server.js
web    | Runing on http://localhost:8080
worker | running: docker run -i --name node-example-worker -e REDIS_URL=redis://:password@172.17.0.1:6379/0 node-example/worker sh -c node worker.js
```

if i take away the links section in the worker:

```yaml
worker:
  command: node worker.js
  environment:
    - REDIS_URL
  build: .
```

it errors out b.c of the declaration

```bash
$ cx start
ERROR: env expected: REDIS_URL
```

in production, you get the REDIS_URL set to the load balancer of the exposed port:

```bash
$ cx ps
ID            NAME    RELEASE      SIZE  STARTED         COMMAND
a946e6f3bb35  redis   RHIXMHQAXCF  256   14 minutes ago  /docker-entrypoint.sh redis-server /tmp/redis.conf
740808a05c2f  web     RHIXMHQAXCF  256   12 minutes ago  sh -c 'node server.js'
...
~/convox/node-example (master*)$ cx exec 740808a05c2f env
...
REDIS_URL=redis://:password@internal-node-example-redis-internal-851383837.us-east-1.elb.amazonaws.com:6379/0

```


api/manifest/manifest.go
```diff
@@ -130,8 +135,11 @@ func Read(dir, filename string) (*Manifest, error) {
 				}
 			}
 
-			m[name].Volumes[i] = strings.Join(parts, ":")
+			entry.Volumes[i] = strings.Join(parts, ":")
 		}
+
+		entry.Manifest = &m
```
ddollar: The back-reference seems like it could lead to bugs in the future. Is there any way to get this by passing a manifest around where you need it?

ddollar:
Could we add some functionality to the .yml fixtures that map to changes in the CloudFormation .json that we could put under test?

api/manifest/manifest.go
```diff
@@ -278,9 +305,106 @@ func (me *ManifestEntry) EnvironmentArray() []string {
...
+//       assuming local dev is done on DOCKER_HOST
+func (me *ManifestEntry) ResolvedLinkVars() (map[string]string, error) {
+	linkVars := make(map[string]string)
+
+	if me.Manifest == nil {
+		return linkVars, nil
+	}
+
+	for _, link := range me.Links {
+		other := (*me.Manifest)[link]
```
csquared: @ddollar this is what i need that backreference for.  the methods calling ResolvedLinkVars are also defined on `ManifestEntry`, so I can't get it there either. I *could* pass the manifest to `ResolvedEnvironment` in the call to `MissingEnvironment`, but would also have to modify `runAsync` to take the manifest and pass that through... I'll wire this all up in another PR so you can tell me what you think.

csquared: lol was able to remove all the backreferences :dancer:

ddollar: nice :)

ddollar:
:+1:

csquared:
have the CLI *and* json tested, with first tests for the release object

csquared:
added some functionality so this works with images that aren't specially pre-configured with convox env vars: https://github.com/convox-examples/rails-nginx/blob/master/docker-compose.yml

nzoschke:
You can also set the proper LINK_ variables in docker-compose.yml, when you don't want to change the image. So you could use the offical `_/redis` like this:

```
web:
  build: .
  command: node server.js
  environment:
    - REDIS_URL
  ports:
    - 80:8080
  links:
    - redis

worker:
  command: node worker.js
  environment:
    - REDIS_URL
  build: .
  links:
    - redis

redis:
  image: redis
  environment:
    - LINK_USERNAME=
    - LINK_SCHEME=redis
    - LINK_PATH=0
    - LINK_PASSWORD=
  ports:
    - 6379
```

However ideally it's baked into the image, as `convox/redis` does here:

https://github.com/convox/redis/blob/master/Dockerfile#L10-L13

Issue #211, opened 2015-12-14
Amazon EC2 Container Registry (ECR)


nathany:
Are there plans to replace the [custom registry](https://github.com/convox/registry) with [ECR](https://aws.amazon.com/ecr/)? And importantly, how would you transition existing users?

I'm mostly interested in this [to reduce cost](https://github.com/convox/convox.github.io/issues/78) by having more resources available to my applications.

Though it could also reduce some complexity for Convox. Integration vs. invention.

csquared:
Absolutely in our plans and for the reasons mentioned :)

alex:
It's out now https://aws.amazon.com/about-aws/whats-new/2015/12/amazon-ec2-container-registry-now-availlable/ :-)

nathany:
Got a 404 from that link. Try this one: https://aws.amazon.com/blogs/aws/ec2-container-registry-now-generally-available/ linked from http://thenewstack.io/amazons-ec2-container-registry-goes-live/

nzoschke:
This is being worked on. Next steps are rebasing, testing and merging this PR

https://github.com/convox/rack/pull/249

nathany:
Looks like it's merged now.

nzoschke:
It is! `convox rack update` to get the changes.

You need to do 2 deploys to an app to get ECR. The first deploy creates the ECR Repository. The next deploy will start pushing images to that instead of the v1 registry.

ha333823:
Sorry, is this not backward compatible? getting error while creating convo app with latest version but in Oregon region which doesn't support ECR yet.

nzoschke:
Shoot, I only tested in us-east, so thank you for the report. It is intended to be backwards compatible.

I now see this bit of info:

> Effective today, Amazon ECR is available in US East (Northern Virginia) with more regions on the way soon!

For a short term fix you can update your rack back to a version before ECR.

The v1 registry is still running in your rack. I can change CloudFormation to not provision ECR if the region is not us-east-1 until ECR is available in more regions.

nzoschke:
Confirmed app creation is broken in eu-west-1

11:09:39 UTC-0800	CREATE_FAILED	Custom::ECRRepository	RegistryRepository	Invalid PhysicalResourceId
Physical ID:httpd2-RegistryRepository-1AS6DIMGBU8P4

nzoschke:
This was fixed in https://github.com/convox/rack/pull/274

PR #274, opened 2016-01-21
Only provision ECR in us-east-1. Other regions will fall back to the …

nzoschke:
In regions where ECR is not provisioned, it will fall back to the registry running in the cluster.

https://github.com/convox/rack/issues/211


api/models/templates/app.tmpl
```diff
@@ -350,6 +356,7 @@
     "Value": { "Ref": "AWS::AccountId" }
   },
   "RegistryRepository": {
```
csquared: Does `RegistryId` also need the condition `RegionHasRegsitry`?  

It seems to me if `RegsitryId` is set, and it appears to me to reference something that will always be there, then this branch will still (always) execute, which we do not wish to happen:  https://github.com/convox/rack/blob/master/api/models/build.go#L223 .

csquared:
@nzoschke not sure if this change accomplishes the intended purpose, but think just adding the condition should fix that. lmk if my understanding is off for some reason.

api/models/templates/app.tmpl
ddollar: Yep both of them should have the condition

nzoschke: Great catch, fixed.

ddollar:
:+1:

PR #275, opened 2016-01-21
update amis

csquared:
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_container_instance.html


mattmanning:
:+1:

PR #282, opened 2016-01-23
One off processes with ECR

mattmanning:


mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;6XknZrXj&#x2F;312-convox-run-with-ecr"> `convox run` with ECR</a>

PR #295, opened 2016-01-27
add support for docker-compose's `dockerfile` config option

gmelika:
This is to add support for the `dockerfile` config option as described here: https://docs.docker.com/compose/compose-file/#dockerfile

### DIFF
--- a/api/manifest/manifest.go
+++ b/api/manifest/manifest.go
@@ -155,8 +155,7 @@ func buildSync(source, tag string, cache bool, dockerfile string) error {
 	}
 
 	if dockerfile != "" {
-		args = append(args, "-f")
-		args = append(args, dockerfile)
+		args = append(args, "-f", filepath.Join(source, dockerfile))
 	}
 
 	args = append(args, source)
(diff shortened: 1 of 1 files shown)


gmelika:
@nzoschke I've fixed the issue with Dockerfile. please test with this and let me know

api/manifest/manifest.go
```diff
@@ -153,6 +154,14 @@ func buildSync(source, tag string, cache bool) error {
 		args = append(args, "--no-cache")
 	}
 
+	if dockerfile != "" {
+		args = append(args, "-f")
```
ddollar: Could simplify this a bit:
`args = append(args, "-f", "/" + dockerfile))`

gmelika: was going for speed with Buffer.WriteString instead, but I guess that's an overkill during the build step.  I'll change it

gmelika:
good call.  rebased and pushed the change

ddollar:
You can try this with `convox rack update 20160128212241-support-dockerfile`

ddollar:
Thanks!

PR #305, opened 2016-01-28
Set EnableDnsSupport and EnableDnsHostnames for rack VPC

xtfxme:
http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-dns.html

Allows host instances within the rack to have internal hostnames only resolvable
from within the VPC. The host uses an Amazon provided DNS server configured via
DHCP. Containers mount /etc/resolv.conf from the host giving them the same
resolution. This allows unrelated EC2 instances to join the VPC, such as a mail
server, under a common local name (eg. mail.internal) for service discovery or
to simplify some configuration.

### DIFF
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -452,6 +452,8 @@
     "Vpc": {
       "Type": "AWS::EC2::VPC",
       "Properties": {
+        "EnableDnsSupport": "true",
+        "EnableDnsHostnames": "true",
         "CidrBlock": { "Ref": "VPCCIDR" },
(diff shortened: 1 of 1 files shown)


xtfxme:
This is also necessary to use private hosted zones with the VPC via Route53.

xtfxme:
Any concerns here? My understanding is it does not actually affect the normal operations, just allows some DNS trickery.

ddollar:
These are being covered in #318 thanks!

xtfxme:
Perfect thanks!

PR #307, opened 2016-01-28
add support for MultiAZ parameter in the db CF templates

soulware:
not sure what else is required to expose the MultiAZ parameter to `convox service create`
Seems like support for this param is already there in api.go


ddollar:
thanks!

PR #318, opened 2016-02-03
Change docker volume location and instance rolls

ddollar:
* Use `/dev/xvdcz` to mount the Docker directory and use 50GB for the default size
* Don't let the instance count drop below desired count


api/dist/kernel.json
```diff
@@ -623,7 +625,7 @@
       "Properties": {
         "AssociatePublicIpAddress": true,
         "BlockDeviceMappings": [{
-          "DeviceName": "/dev/xvda",
+          "DeviceName": "/dev/xvdcz",
```
nzoschke: Really good catch.

nzoschke:
:+1:

PR #321, opened 2016-02-03
CF throttle

nzoschke:


ddollar:
:+1: Hopefully this should help quite a lot

PR #325, opened 2016-02-07
Use a singleton monitor process

ddollar:


ddollar:
This is rebased on top of `describe-stacks-lock`

mattmanning:
:+1: Looks like a good approach. Splitting monitor out into a separate process and setting DeploymentMaximumPercent to 100 seems like it should prevent more than one monitor from running at a time as I understand things.

nzoschke:
Lots of good CI runs off this branch now: https://circleci.com/gh/convox/rack/448

Issue #200, opened 2015-12-04
App and Stack name collisions on different racks in the same region


gisnalbert:
It is not possible to create a app on Rack1, if an app with the same name already exists on Rack2.
Rack1 and Rack2 deployed on the same AWS region.

I believe this may be related to #159

csquared:
this one is a little tougher because amazon enforces the uniques of stack names in an AWS account.  considering this is under the hood, it might make sense to prefix the Cloudformation Stack name of all apps and services created by a rack with that rack's name. Might be a pain to strip out, but I can't think of a cleaner way.

xtfxme:
I'm going to start looking into this today, so any additional input is appreciated.

Would it be beneficial to use a randomly generated (shared rack and apps) prefix or postfix? This would be easy to strip without knowing the racks name (in the app case). This would however allow racks with the same name, which seems like possibly a problem.

Anything else I should consider?

ddollar:
Prefixing app names with the stack name of the Rack should be ok. App stacks should have an `app` tag that contains the actual app name so anywhere that needs to display it can use that.

gisnalbert:
@xtfxme 
Just have to be careful with prefix/postfix because of maximum length names on some AWS resources (e.g. internal ELBs has a limit of 32 characters on the names, and right now Convox uses the app name on the ELB name).

pyjsorg:
Ah good to know. Since the ELB has its own randomness appended we can probably keep using only the app name there, vs the longer rack + app name.

xtfxme:
Ah good to know. Since the ELB has its own randomness appended we can probably keep using only the app name there, vs the longer rack + app name.

xtfxme:
PR: https://github.com/convox/rack/pull/327

nzoschke:
The PR moved to: https://github.com/convox/rack/pull/341

nzoschke:
This is released. `convox update` to >= 20160223211445 to get it.

This behavior only applies to new apps/services. You will need to delete and re-create old apps/services to stop the old behavior.

Thanks a ton to @xtfxme for contributing to this patch.

PR #326, opened 2016-02-08
Update system.go

xtfxme:
Cache `DescribeStacks` call only if err is nil.

### DIFF
--- a/api/models/system.go
+++ b/api/models/system.go
@@ -48,7 +48,7 @@ func doDescribeStack(input cloudformation.DescribeStacksInput) (*cloudformation.
 
 		res, err := CloudFormation().DescribeStacks(&input)
 
-		if err != nil {
+		if err == nil {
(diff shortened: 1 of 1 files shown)


xtfxme:
The fix for #200 causes several cache misses (every call to `GetApp` checks first for `app` then for `rack-app` and over time the former will always fail) this was caching failed results and returning them as a success: `return s.Output, nil`.

xtfxme:
Wasn't this (`DescribeStacks`) wrapped by a lock at some point? I know I read some code involving a `Mutex` but now I can't find it... otherwise isn't there a race to update the cache? Was this removed?

ddollar:
Yep. #324 also includes this fix so going to close this PR. Thanks for the report :)

xtfxme:
I think we'll need to add negative caching to deal with all the misses generated by checking for `app` instead of `rack-app`

PR #327, opened 2016-02-08
Prefix app and service StackNames with the rack name (fixes #200).

### LINKED ISSUE #200: App and Stack name collisions on different racks in the same region

xtfxme:
Calls to `GetApp` and `GetService` attempt to resolve `name` then `rack-name`.

Patch introduces a new model helper, `shortNameToStackName`, and two new
methods, `App.StackName` and `Service.StackName`.


xtfxme:
This is based on `singleton-notifier` branch because I needed the fixes to `DescribeStack` cache.

This patch causes tons of cache misses in the `DescribeStack` cache... we'll probably need to support negative caching otherwise we'll be undoing the reason for the cache (the constant misses will result in actual API calls).

In general, more stuff should be changed to use `.StackName` methods instead of `DescribeStack(shortNameToStackName(name))`, but maybe that can wait.

nzoschke:
@xtfxme can you provide some sample data about the different app stack name and tag conventions? 

Ideally I'd like to include some tests around this to prevent any regressions.

xtfxme:
Also a couple of open questions:

`LoadBalancerName` in app templates use the StackName, which includes the rack prefix... but this makes the ELB name longer than necessary and likely to blow the 32 char limit. New param? Hard code the short name as necessary? Something/nothing else?

Do certificate names/operations need the full name, or is the short name good enough? eg. `ServerCertificateName: aws.String(certName(a.Name, process, port))`? Is `a.Name` good enough or do we need `a.StackName()`?

xtfxme:
reopening this against master

PR #328, opened 2016-02-08
Prefix app and service StackNames with the rack name.

### LINKED ISSUE #200: App and Stack name collisions on different racks in the same region

xtfxme:
Calls to `GetApp` and `GetService` attempt to resolve `name` then `rack-name`.

Patch introduces a new model helper, `shortNameToStackName`, and two new
methods, `App.StackName` and `Service.StackName`.

### DIFF
--- a/api/controllers/apps.go
+++ b/api/controllers/apps.go
@@ -47,11 +47,15 @@ func AppShow(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 func AppCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	name := r.FormValue("name")
 
-	app := &models.App{
-		Name: name,
+	// Early check for unbound app only.
+	app, err := models.GetAppFast(name, false, true)
+
+	if awsError(err) == "ValidationError" {
+		// If unbound check fails this will result in a bound app.
+		app = &models.App{Name: name}
 	}
 
-	err := app.Create()
+	err = app.Create()
 
 	if awsError(err) == "AlreadyExistsException" {
 		app, err := models.GetApp(name)
--- a/api/controllers/apps.go
+++ b/api/controllers/apps.go
@@ -48,14 +48,13 @@ func AppCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	name := r.FormValue("name")
 
 	// Early check for unbound app only.
-	app, err := models.GetAppFast(name, false, true)
-
-	if awsError(err) == "ValidationError" {
-		// If unbound check fails this will result in a bound app.
-		app = &models.App{Name: name}
+	if app, err := models.GetAppUnbound(name); err == nil {
+		return httperr.Errorf(403, "there is already a legacy app named %s (%s). We recommend you delete this app and create it again.", name, app.Status)
 	}
 
-	err = app.Create()
+	// If unbound check fails this will result in a bound app.
+	app := &models.App{Name: name}
+	err := app.Create()
 
 	if awsError(err) == "AlreadyExistsException" {
 		app, err := models.GetApp(name)
--- a/api/controllers/apps_test.go
+++ b/api/controllers/apps_test.go
@@ -3,6 +3,7 @@ package controllers_test
 import (
 	"encoding/json"
 	"net/url"
+	"os"
 	"testing"
 
 	"github.com/convox/rack/Godeps/_workspace/src/github.com/stretchr/testify/assert"
@@ -12,12 +13,13 @@ import (
 )
 
 func init() {
+	os.Setenv("RACK", "convox-test")
 	models.PauseNotifications = true
 	test.HandlerFunc = controllers.HandlerFunc
 }
 
 func TestAppList(t *testing.T) {
-	aws := test.StubAws(test.DescribeStackCycleWithoutQuery("bar"))
+	aws := test.StubAws(test.DescribeStackCycleWithoutQuery("convox-test-bar"))
 	defer aws.Close()
 
 	body := test.HTTPBody("GET", "http://convox/apps", nil)
@@ -32,7 +34,10 @@ func TestAppList(t *testing.T) {
 }
 
 func TestAppShow(t *testing.T) {
-	aws := test.StubAws(test.DescribeAppStackCycle("bar"))
+	aws := test.StubAws(
+		test.DescribeStackNotFound("convox-test-bar"),
+		test.DescribeAppStackCycle("bar"),
+	)
 	defer aws.Close()
 
 	body := test.HTTPBody("GET", "http://convox/apps/bar", nil)
@@ -47,7 +52,10 @@ func TestAppShow(t *testing.T) {
 }
 
 func TestAppShowWithAppNotFound(t *testing.T) {
-	aws := test.StubAws(test.DescribeStackNotFound("bar"))
+	aws := test.StubAws(
+		test.DescribeStackNotFound("convox-test-bar"),
+		test.DescribeStackNotFound("bar"),
+	)
 	defer aws.Close()
 
 	test.AssertStatus(t, 404, "GET", "http://convox/apps/bar", nil)
@@ -55,8 +63,9 @@ func TestAppShowWithAppNotFound(t *testing.T) {
 
 func TestAppCreate(t *testing.T) {
 	aws := test.StubAws(
-		test.CreateAppStackCycle("application"),
-		test.DescribeAppStackCycle("application"),
+		test.DescribeStackNotFound("application"),
+		test.CreateAppStackCycle("convox-test-application"),
+		test.DescribeAppStackCycle("convox-test-application"),
 	)
 	defer aws.Close()
 
@@ -74,10 +83,24 @@ func TestAppCreate(t *testing.T) {
 	}
 }
 
-func TestAppCreateWithAlreadyExists(t *testing.T) {
+func TestAppCreateWithAlreadyExistsUnbound(t *testing.T) {
 	aws := test.StubAws(
-		test.CreateAppStackExistsCycle("application"),
 		test.DescribeAppStackCycle("application"),
+		test.CreateAppStackExistsCycle("application"),
+		test.DescribeStackNotFound("convox-test-application"),
+		test.DescribeAppStackCycle("application"),
+	)
+	defer aws.Close()
+
+	val := url.Values{"name": []string{"application"}}
+	test.AssertStatus(t, 403, "POST", "http://convox/apps", val)
+}
+
+func TestAppCreateWithAlreadyExistsBound(t *testing.T) {
+	aws := test.StubAws(
+		test.DescribeStackNotFound("application"),
+		test.CreateAppStackExistsCycle("convox-test-application"),
+		test.DescribeAppStackCycle("convox-test-application"),
 	)
 	defer aws.Close()
 
@@ -91,8 +114,8 @@ bucket name to the ephermeral host, so you get `app-XXX.127.0.0.1`
 */
 func TestAppDelete(t *testing.T) {
 	aws := test.StubAws(
-		test.DescribeAppStackCycle("bar"),
-		test.DeleteStackCycle("bar"),
+		test.DescribeAppStackCycle("convox-test-bar"),
+		test.DeleteStackCycle("convox-test-bar"),
 	)
 	defer aws.Close()
 
@@ -107,7 +130,10 @@ func TestAppDelete(t *testing.T) {
 }
 
 func TestAppDeleteWithAppNotFound(t *testing.T) {
-	aws := test.StubAws(test.DescribeStackNotFound("bar"))
+	aws := test.StubAws(
+		test.DescribeStackNotFound("convox-test-bar"),
+		test.DescribeStackNotFound("bar"),
+	)
 	defer aws.Close()
 
 	test.AssertStatus(t, 404, "DELETE", "http://convox/apps/bar", nil)
--- a/api/models/app.go
+++ b/api/models/app.go
@@ -59,7 +59,32 @@ func ListApps() (Apps, error) {
 }
 
 func GetApp(name string) (*App, error) {
-	res, err := DescribeStack(name)
+	return GetAppFast(name, true, true)
+}
+
+func GetAppFast(name string, bound bool, unbound bool) (*App, error) {
+	stackName := shortNameToStackName(name)
+
+	switch {
+	case !bound && !unbound:
+		// User requested no lookups, return empty App.
+		return &App{Name: name}, nil
+	case bound && !unbound:
+		// User requested bound lookup only, reset short name.
+		name = stackName
+	case !bound && unbound:
+		// User requested unbound lookup only, reset long name.
+		stackName = name
+	}
+
+	res, err := DescribeStack(stackName)
+
+	// Setting to the same value results in at most a single lookup.
+	if name != stackName && awsError(err) == "ValidationError" {
+		// Only lookup an unbound app if the name/stackName differ and the
+		// bound lookup fails.
+		res, err = DescribeStack(name)
+	}
 
 	if err != nil {
 		return nil, err
@@ -81,6 +106,19 @@ func GetApp(name string) (*App, error) {
 }
 
 var regexValidAppName = regexp.MustCompile(`\A[a-zA-Z][-a-zA-Z0-9]{3,29}\z`)
+
+func (a *App) StackName() string {
+	if a.Tags == nil {
+		return shortNameToStackName(a.Name)
+	}
+
+	if _, ok := a.Tags["Name"]; ok {
+		// Use the name provided by the user (they *should* match).
+		return shortNameToStackName(a.Name)
+	}
+
+	return a.Name
+}
 
 func (a *App) Create() error {
 	helpers.TrackEvent("kernel-app-create-start", nil)
@@ -111,11 +149,12 @@ func (a *App) Create() error {
 		"Rack":   os.Getenv("RACK"),
 		"System": "convox",
 		"Type":   "app",
+		"Name":	  a.Name,
 	}
 
 	req := &cloudformation.CreateStackInput{
 		Capabilities: []*string{aws.String("CAPABILITY_IAM")},
-		StackName:    aws.String(a.Name),
+		StackName:    aws.String(a.StackName()),
 		TemplateBody: aws.String(formation),
 	}
 
@@ -180,9 +219,7 @@ func (a *App) Cleanup() error {
 func (a *App) Delete() error {
 	helpers.TrackEvent("kernel-app-delete-start", nil)
 
-	name := a.Name
-
-	_, err := CloudFormation().DeleteStack(&cloudformation.DeleteStackInput{StackName: aws.String(name)})
+	_, err := CloudFormation().DeleteStack(&cloudformation.DeleteStackInput{StackName: aws.String(a.StackName())})
 
 	if err != nil {
 		helpers.TrackEvent("kernel-app-delete-error", nil)
@@ -200,7 +237,7 @@ func (a *App) Delete() error {
 
 func (a *App) UpdateParamsAndTemplate(changes map[string]string, template string) error {
 	req := &cloudformation.UpdateStackInput{
-		StackName:    aws.String(a.Name),
+		StackName:    aws.String(a.StackName()),
 		Capabilities: []*string{aws.String("CAPABILITY_IAM")},
 	}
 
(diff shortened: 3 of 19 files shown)


xtfxme:
original: https://github.com/convox/rack/pull/327

xtfxme:
@nzoschke 

I will use the terms "unbound" and "bound" to refer to apps and services created before and after this patch, respectively. An unbound app/service does **not** have a rack prefix.

It's not be possible to create unbound apps or services. All new stacks will be prefixed. However, if an unbound app or service already exists, creation will fail to prevent logical duplicates.

Calls to `GetApp(name)` and `GetService(name)` attempt to locate a stack called `name` before trying `rack-name`. That is, older, unbound names are preferred to preserve backwards compatibility. If `name` matches the current rack name, only `name` will be attempted.

Bound apps and services have an additional `Name` tag used by the `App.StackName()` and `Service.StackName()` methods to determine the fully qualified name. **This is the only detectable difference between bound and unbound stacks.** As such, using one of these methods is the preferred and safest way to derive the correct stack name.

Per the last section, `app = &models.App{Name: name}` is no longer a safe way to interact with `App` or `Service` models, you should instead use `GetApp` or `GetService`. Tag information is necessary to determine the real `.StackName()`.

Docker repositories use `.StackName()` to avoid clashing with similar apps on a different rack.

Everything needing account-wide uniqueness should also use `.StackName()`.

xtfxme:
Updated to include prefixes on Certificate names in IAM.

nzoschke:
Some pointers for the Travis test failures...

controllers/apps_test.go uses a testing strategy reminiscent of the [ruby VCR gem](https://github.com/vcr/vcr). The client talks to a mock HTTP server that has expectations around a series of requests coming in and responses that they return.

Your patch has changed the expected request/response cycles in the `controllers/apps_test.go` TestAppCreate and TestAppShowWithAppNotFound tests.

The TestAppShowWithAppNotFound "No cycles remaining to replay." error means that `GET http://convox/apps/bar` call now makes one or more additional AWS HTTP calls that the test harness is not expecting.

Look at the `DescribeStackNotFound` function in `test/aws_cycles.go`. That has a single mock CloudFormation call that returns a 'does not exist' 400 for the stack.

With your changes, it looks like we now expect *2* CloudFormation calls for the two different stack name format lookups.

The TestAppCreate error is similar:

```
Request does not match next cycle.
CYCLE REQUEST:
RequestURI: /
Operation: 
Body: /^Action=CreateStack/
ACTUAL REQUEST:
RequestURI: /
Operation: 
Body: Action=DescribeStacks&StackName=application&Version=2010-05-15
```

Your change now performs a CF DescribeStacks call first, where before it was expecting a CF CreateStack first.

nzoschke:
So to fix these tests, you will end up adding at least one more cycle to the `aws := test.StubAws()` setup in the TestAppShowWithAppNotFound and TestAppCreate test methods.

xtfxme:
@nzoschke YAY I think we're good now!

A few tests could probably have bound/unbound variants, but that seemed like overkill for first pass and I'm not sure how much real value it would add.

I added `Tags["Rack"] == "convox-test"` to the appxml to deal with the fact I added the `RACK` environment variable.

We might want to test `Tag["Name"] == "shortname"` at some point, but the various functions around `DescribeStacks` do not have a concept of short vs. long names, so I left that out.

Let me know if you need anything else!

xtfxme:
As is, this patch does **not** address the concern raised here: https://github.com/convox/rack/issues/200#issuecomment-178661675

I wasn't sure the best way to handle this (hard code in template, new CF param, something else?). The CF template currently references `StackName` directly to generate an ELB name, as it did before.

api/models/app.go
```diff
@@ -82,6 +86,19 @@ func GetApp(name string) (*App, error) {
 
 var regexValidAppName = regexp.MustCompile(`\A[a-zA-Z][-a-zA-Z0-9]{3,29}\z`)
 
+func (a *App) StackName() string {
+	if a.Tags == nil {
+		return shortNameToStackName(a.Name)
+	}
+
+	nameTag, ok := a.Tags["Name"]
```
nzoschke: More idiomatic Go is to put the if and assignment on the same line, and to do the positive condition first to limit the scope of the name variable.

```go
if name, ok := a.Tags["Name"]; ok {
  return shortNameToStackName(name) 
}

return a.Name
```

api/controllers/apps.go
```diff
@@ -47,11 +47,13 @@ func AppShow(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 func AppCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	name := r.FormValue("name")
 
-	app := &models.App{
-		Name: name,
+	app, err := models.GetApp(name)
```
nzoschke: why do you look up the app here?

api/controllers/apps_test.go
```diff
@@ -68,14 +76,15 @@ func TestAppCreate(t *testing.T) {
 		err := json.Unmarshal([]byte(body), &resp)
 
 		if assert.Nil(t, err) {
-			assert.Equal(t, "application", resp["name"])
+			assert.Equal(t, "convox-test-application", resp["name"])
```
nzoschke: the /apps API should still return the logical app name without the stack prefix.

api/controllers/apps_test.go
```diff
@@ -47,7 +49,10 @@ func TestAppShow(t *testing.T) {
 }
 
 func TestAppShowWithAppNotFound(t *testing.T) {
-	aws := test.StubAws(test.DescribeStackNotFound("bar"))
+	aws := test.StubAws(
+		test.DescribeStackNotFound("bar"),
```
nzoschke: I wonder if this is the right order... After this lands, the majority of apps will have the name prefix, so perhaps its more correct to look the `convox-test-bar` stack up first, and do the second lookup as a fallback for legacy apps.

api/controllers/apps_test.go
```diff
@@ -55,8 +60,11 @@ func TestAppShowWithAppNotFound(t *testing.T) {
 
 func TestAppCreate(t *testing.T) {
 	aws := test.StubAws(
-		test.CreateAppStackCycle("application"),
-		test.DescribeAppStackCycle("application"),
+		test.DescribeStackNotFound("application"),
+		test.DescribeStackNotFound("convox-test-application"),
+		test.CreateAppStackCycle("convox-test-application"),
+		test.DescribeStackNotFound("application"),
+		test.DescribeAppStackCycle("convox-test-application"),
```
nzoschke: Related to the comment in AppCreate, I don't see why POST /apps needs to make more CF calls. It should be able to stick with 2 API calls: CreateAppStack for convox-test-application, and DescribeAppStack for convox-test-application.

nzoschke: Ok, I'm starting to see it, based on the TestAppCreateWithAlreadyExistsBound test...

nzoschke:
I made some unit tests to help me understand the behavior:

```diff
diff --git a/api/models/app_test.go b/api/models/app_test.go
new file mode 100644
index 0000000..c802922
--- /dev/null
+++ b/api/models/app_test.go
@@ -0,0 +1,48 @@
+package models_test
+
+import (
+       "os"
+       "testing"
+
+       "github.com/convox/rack/api/models"
+       "github.com/stretchr/testify/assert"
+)
+
+func init() {
+       os.Setenv("RACK", "convox-test")
+}
+
+func TestRackStackName(t *testing.T) {
+       r := models.App{
+               Name: "convox-test",
+       }
+
+       assert.Equal(t, "convox-test", r.StackName())
+}
+
+func TestAppStackName(t *testing.T) {
+       // unbound app (no rack prefix, no Name tag)
+       a := models.App{
+               Name: "httpd",
+               Tags: map[string]string{
+                       "Type":   "app",
+                       "System": "convox",
+                       "Rack":   "convox-test",
+               },
+       }
+
+       assert.Equal(t, "httpd", a.StackName())
+
+       // bound app (rack prefix, and Name tag)
+       a = models.App{
+               Name: "httpd",
+               Tags: map[string]string{
+                       "Name":   "httpd",
+                       "Type":   "app",
+                       "System": "convox",
+                       "Rack":   "convox-test",
+               },
+       }
+
+       assert.Equal(t, "convox-test-httpd", a.StackName())
+}
```

nzoschke:
I have one big stylistic question... What if the prefix is only added for racks where `ENV["RACK"] != "convox"` ?

The vast majority of installs will be one "convox" stack per region and account, and I've grown quite fond of seeing literal app stack names.

api/controllers/apps.go
nzoschke: Ok I reviewed everything and I think I get it. When a user says create app "foo", you want to see if there's an unbound app "foo" before creating "convox-foo". Can you add a comment or two to this effect?

api/controllers/apps_test.go
nzoschke: Still could the logic be changed to tighten this up to something like:

```
test.DescribeStackNotFound("application"),                 // legacy unbound app check
test.CreateAppStackCycle("convox-test-application"), // expected create/describe cycles
test.DescribeAppStackCycle("convox-test-application"),
```

api/controllers/apps_test.go
nzoschke: I think changing the order would also address up a few of the test cycle questions I have.

api/controllers/apps_test.go
```diff
@@ -85,6 +94,20 @@ func TestAppCreateWithAlreadyExists(t *testing.T) {
 	test.AssertStatus(t, 403, "POST", "http://convox/apps", val)
 }
 
+func TestAppCreateWithAlreadyExistsBound(t *testing.T) {
+	aws := test.StubAws(
+		test.DescribeStackNotFound("application"),
+		test.DescribeAppStackCycle("convox-test-application"),
+		test.CreateAppStackExistsCycle("convox-test-application"),
+		test.DescribeStackNotFound("application"),
+		test.DescribeAppStackCycle("convox-test-application"),
```
nzoschke: Could the logic be changed to tighten this up to something like:

```
test.DescribeStackNotFound("application"),
test.DescribeAppStackCycle("convox-test-application"),
test.CreateAppStackExistsCycle("convox-test-application"),
test.DescribeAppStackCycle("convox-test-application"),
```

nzoschke:
Ok, this is pretty close. The issues I see are:

- [ ] The TestAppCreate response not returning the logical app name
- [ ] Do other APIs have the same problem?
- [ ] Changing logic to tighten up the request/response cycles if appropriate
- [ ] Failing CI

nzoschke:
Unfortunately this failed CI: https://circleci.com/gh/convox/rack/457

It looks like `convox ps` isn't doing the right thing:

```
+ convox ps --app httpd-457
+ grep -i PENDING
ERROR: ValidationError: Stack with id httpd-457 does not exist
	status code: 400, request id: 2e7de641-cf89-11e5-8b6b-99bd8d2754b6
```

It also looks like the CI script is now failing to extract the right URL to test:

```
+ convox apps info --app httpd-457
Name       httpd-457
Status     running
Release    RVFFDCMIDER
Processes  web
Endpoints  convox-457-httpd-457-1246181793.us-east-1.elb.amazonaws.com:80 (web)
++ convox apps info --app httpd-457
++ egrep -o 'httpd.*.amazonaws.com:80'
+ url=http://httpd-457-1246181793.us-east-1.elb.amazonaws.com:80
```

nzoschke:
I pushed a couple commits to https://github.com/convox/rack/tree/stack-prefixes

api/controllers/apps.go
xtfxme: Yeah that is the reason, I can definitely add more comments.

api/controllers/apps_test.go
xtfxme: I think you are right. I actually had it this way to begin with, and changed it for some raisin... I think it had something to do with preventing duplicates, but whatever the reason it's a non-issue now. I'll try reversing the order.

api/controllers/apps_test.go
xtfxme: I'll play around here after doing the reversal. There is still going to be a `DescribeStackNotFound` if `AppCreate` continues using `GetApp`, since that wants to check for both styles... buuut since `AppCreate` is special, maybe it should simply call `DescribeStacks("application")` directly and avoid the just as expected cache miss on `convox-test-application` (you were asked to create it after all!). I'll fix up.

api/controllers/apps_test.go
xtfxme: Oh good catch, I originally overlooked that as the result of a `DescribeStacks` call. I'll fix up.

api/controllers/apps_test.go
xtfxme: Once `AppCreate` uses `DescribeStacks` directly, and the un/bound order is swapped, this *should* fall out.

api/models/app.go
xtfxme: Oh sure. Heh I only changed it so I could use `nameTag` later... but... `a.Name`... (doh)! I'll fix up.

xtfxme:
Awesome thanks for all the comments, I'll roll in your stuff and make the additional changes.

nzoschke:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;7WyJznqw&#x2F;329-rack-name-app-prefix"> rack name app prefix</a>

nzoschke:
I made a decision, having the prefix on every app name is best for consistency.

xtfxme:
@nzoschke I included your additions, made the updates, then squashed and rebased to master.

I didn't see your update regarding the prefix, so this iteration special cases the `convox` rack, the upside to this is (current) default installs will not generate `DescribeStack` cache misses, ever. I can change it if you decide.

xtfxme:
Beh I guess I broke stuff, looking

nzoschke:
Yeah if you can change it to add the `convox-` prefix, that would be appreciated.

nzoschke:
I still need to review this closely, but I got the tests passing here:

https://github.com/convox/rack/tree/stack-prefixes
https://github.com/convox/rack/commit/13ef519f68a43332ba034ffa46748291ed94a160

api/models/app.go
```diff
@@ -59,7 +59,32 @@ func ListApps() (Apps, error) {
 }
 
 func GetApp(name string) (*App, error) {
-	res, err := DescribeStack(name)
+	return GetAppFast(name, true, true)
+}
+
+func GetAppFast(name string, bound bool, unbound bool) (*App, error) {
```
nzoschke: I'm having a tough time with this function definition. My first impression is that it's hard to know how to use it and what to expect.

xtfxme: My first draft had some 2 consts in the file for use as flags, but that looked like overkill to me, so I removed them. The only real motivation for the 2 separate definitions is so most code can continue using `GetApp` but `(App|Service)Create` -- specifically -- can use `GetAppFast` to avoid cache misses and force query the unbound app first (since bound is done first now). LMK if changes are needed... even to terminology (I'm not sure I care for the un/bound terms, but it works well enough).

api/controllers/apps.go
```diff
@@ -47,11 +47,15 @@ func AppShow(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 func AppCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	name := r.FormValue("name")
 
-	app := &models.App{
-		Name: name,
+	// Early check for unbound app only.
+	app, err := models.GetAppFast(name, false, true)
```
nzoschke: How about breaking up GetAppFast into smaller and more descriptive functions? Since this lookup is for backwards compatibility only, the call logic could make it more obvious:

```
if app, err := models.GetAppUnbound(); err != nil {
  return httperr.Errorf(403, "there is already a legacy app named %s (%s). We recommend you delete this app and create it again.", name, app.Status)
}

err := app.Create()
...
```

api/models/app.go
```diff
@@ -59,7 +59,32 @@ func ListApps() (Apps, error) {
 }
 
 func GetApp(name string) (*App, error) {
-	res, err := DescribeStack(name)
+	return GetAppFast(name, true, true)
```
nzoschke: GetApp() is a really crucial function, so I'd prefer that it still inlines the logic to prepend the rack name, describe the stack, etc.

GetAppUnbound() or GetAppLegacy() can duplicate code too.

After all, the long term goal is to return to a single path, which can be as simple as removing the GetAppUnbound() function.

api/controllers/apps_test.go
```diff
@@ -74,9 +83,11 @@ func TestAppCreate(t *testing.T) {
 	}
 }
 
-func TestAppCreateWithAlreadyExists(t *testing.T) {
+func TestAppCreateWithAlreadyExistsUnbound(t *testing.T) {
 	aws := test.StubAws(
+		test.DescribeAppStackCycle("application"),
 		test.CreateAppStackExistsCycle("application"),
```
nzoschke: Ideally this test case can return a 403 based on a single CF call:

```
test.DescribeAppStackCycle("application")
```

It would also make the test clearer to demonstrate that the error string says something about a legacy app.

api/controllers/apps_test.go
```diff
@@ -85,14 +96,26 @@ func TestAppCreateWithAlreadyExists(t *testing.T) {
...
+	aws := test.StubAws(
+		test.DescribeStackNotFound("application"),
+		test.CreateAppStackExistsCycle("convox-test-application"),
+		test.DescribeAppStackCycle("convox-test-application"),
+	)
+	defer aws.Close()
+
+	val := url.Values{"name": []string{"application"}}
+	test.AssertStatus(t, 403, "POST", "http://convox/apps", val)
+}
```
nzoschke: The request cycles look great here.

It would also make the test clearer to demonstrate the error string doesn't say anything about a legacy app.

api/controllers/apps_test.go
```diff
@@ -12,12 +13,13 @@ import (
 )
 
 func init() {
+	os.Setenv("RACK", "convox-test")
 	models.PauseNotifications = true
 	test.HandlerFunc = controllers.HandlerFunc
 }
 
 func TestAppList(t *testing.T) {
```
nzoschke: This test looks perfect.

api/controllers/apps_test.go
```diff
@@ -32,7 +34,10 @@ func TestAppList(t *testing.T) {
 }
 
 func TestAppShow(t *testing.T) {
```
nzoschke: This should test the happy path, so do a single DescribeAppStackCycle("convox-test-bar").

If you want to test the unbound app logic, make another test.

api/controllers/apps_test.go
```diff
@@ -47,7 +52,10 @@ func TestAppShow(t *testing.T) {
 }
 
 func TestAppShowWithAppNotFound(t *testing.T) {
```
nzoschke: this looks perfect

api/controllers/apps_test.go
```diff
@@ -55,8 +63,9 @@ func TestAppShowWithAppNotFound(t *testing.T) {
 
 func TestAppCreate(t *testing.T) {
```
nzoschke: this looks perfect

api/controllers/apps_test.go
```diff
@@ -85,14 +96,26 @@ func TestAppCreateWithAlreadyExists(t *testing.T) {
...
+
+	val := url.Values{"name": []string{"application"}}
+	test.AssertStatus(t, 403, "POST", "http://convox/apps", val)
+}
+
 // TODO: test bucket cleanup. this is handled via goroutines.
 /* NOTE: the S3 stuff fucks up b.c the client ries to prepend the
 bucket name to the ephermeral host, so you get `app-XXX.127.0.0.1`
 */
 func TestAppDelete(t *testing.T) {
```
nzoschke: perfect

api/controllers/apps_test.go
```diff
@@ -107,7 +130,10 @@ func TestAppDelete(t *testing.T) {
 }
 
 func TestAppDeleteWithAppNotFound(t *testing.T) {
```
nzoschke: Perfect

api/models/app.go
nzoschke: Same with GetService()

nzoschke:
This is really close! The logic looks really good so final changes are somewhat stylistic. I think it'd be easier to maintain with one more refactor, where all the legacy stuff is encapsulated in methods / tests that we can simply delete at some point in the future.

So GetApp() and GetService() inline the happy path, and GetAppUnbound() and GetServiceUnbound() do the special case testing.

Similarly TestAppShow() asserts the happy path, and tests like TestAppShowUnbound() assert the special case.

nzoschke:
Closing in favor of https://github.com/convox/rack/pull/341

Issue #155, opened 2015-11-07
Update convox to pass privilege mode settings as part of docker-compose.yml through to ECS 


johnboy14:
I need to run a container in privileged mode but when I pass the privileged setting as part of my docker-compose.yml file, convox seems to ignore it.  I discussed this issue with Noah and he recommended creating an issue for this.

csquared:
Good call -> just created a card out of this.

nzoschke:
Thanks Chris. We talked through this a little more. Recently ECS added support in the Task Definition for a lot more Docker options available in `docker run`. Including privileged mode. It should be very useful to wire up more docker-compose.yml to Task Definition options for convox apps.

http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html

nzoschke:
We should also double check the ECS AMI and resulting docker version when adding support for more docker config.

johnboy14:
Is it possible to update the ECS AMI or the version of docker running on them?.

nzoschke:
Right now it is possible only by specifying a different Ami for the convox stack.

https://github.com/convox/rack/blob/master/api/dist/kernel.json#L104

nzoschke:
http://docs.aws.amazon.com/AmazonECS/latest/developerguide/container_agent_versions.html is the list of ECS AMI Docker versions

johnboy14:
Are you guys going to make the privileged mode change or wait for AWS to update the AMI?.

johnboy14:
Is it possible to determine when this issue might get fixed?.

nzoschke:
Thank you for your patience. @mattmanning just submitted privileged for `convox start` here:

https://github.com/convox/rack/pull/332/files

nzoschke:
We've also upgraded the AMIs and now have Docker 1.9 available everywhere. So one thing remains:

- [ ] Pass privileged true/false to the CloudFormation Custom Resource Lambda function, and ECS task definition

johnboy14:
Thanks for this guys.  I assume it can be read from the manifest also and passed to ECS?.

nzoschke:
@johnboy14 yes, this now gets passed through to ECS too. 

`convox rack update` and you should get version 20160210165552 or later which handles privileged: true in the docker-compose.yml. Example:

https://github.com/convox/rack/blob/fe229e20bd9543c76cf31ba1e8314df1ee06c471/api/models/fixtures/worker.yml

Thanks again for the feature request.

PR #332, opened 2016-02-09
Apply privileged from manifest

mattmanning:
This enhances `convox start`.

I want to use s3fs. In order to access `/dev/fuse` I need to pass the `--privileged` flag to `docker run`. This reads it from `docker-compose.yml` into the Manifest struct and then sets it on the command if true.

https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities


nzoschke:
This looks good for `convox start`. We also want to set this flag in the ECS task definition. See this issue for more info:

https://github.com/convox/rack/issues/155

nzoschke:
Are we doing the right thing in CF template and handler with the boolean? Should it be "true" and "false" or true and false?

mattmanning:
This works great for me, but now I get a message like:

    ERROR: read tcp 172.17.0.5:57035->10.0.2.25:2376: io: read/write on closed pipe

Every time I run a `convox exec` command. So far I haven't been able to figure out exactly where that's coming from.

mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;gVfqSVAt&#x2F;328-apply-privileged-from-manifest"> Apply privileged from manifest</a>

nzoschke:
I just verified that this error is not unique to this branch. Merging and will track down the closed pipe error later.

PR #334, opened 2016-02-10
Support pulling images from other ECR repositories

ddollar:


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;hNr33PWe&#x2F;319-support-ecr-based-private-registries"> support ecr-based private registries</a>

nzoschke:
Is this so you can build an app with a Dockerfile that has `FROM an-image-in-ecr`?

ddollar:
Yes, so you can build from images in your own ECR repositories

nzoschke:
Recently Registry login logic was extracted into a `models/docker.go` AppDockerLogin function, and the responsibility for logging into registries was removed entirely from the cmd/build shim.

Can that strategy still work for this use case? I think so.

And I think that strategy is correct. The API and Build Service is the best layer to handle secrets and authentication, the cmd/build container can have no secrets and be a dumb docker push/pull pipeline.

nzoschke:
Does building from images in ECR happen in development too? How are people managing the expiring token in this case?

ddollar:
Won't work in this case. We don't know which registries to log in to except inside the build shim. The build shim returns the manifest to rack but it's too late by then (already needed to build)

ddollar:
For local development they will need to manage ecr login themselves, same with docker hub

nzoschke:
Ok I see. I have one more question with this strategy. Does passing AWS_ACCESS and AWS_SECRET like this pose any security concerns? My general concerns are that this might expose the secrets in the log, process table, and to untrusted things that can execute during the build.

I do see another strategy where we put this into the Private Registries stuff: https://convox.com/docs/private-registries/

We could extend to support ECR, and prompt for IAM creds instead of a username/password

ddollar:
Yep that would be pretty great

nzoschke:
Do you want to get this merged for testing as a proof of concept? Or hold back for the `convox registries` approach?

nzoschke:
Closing in favor of https://github.com/convox/rack/pull/357

PR #339, opened 2016-02-11
RunAttached uses environment from ECS Task Definition

### LINKED ISSUE #336: one-off processes should inherit the full environment
given an app with a docker-compose.yml:
```
foo:
  image: alpine
  command: watch env
  environment:
    - SECRET=bar
```

do this:
```
convox deploy
[... snip ...]
convox logs
[ expected: includes SECRET ]
convox run foo env
[ unexpected: does not include SECRET ]
```

the one-off environment also does not include (useful) platform managed variables like `RELEASE` or `PROCESS` either.

bobzoller:
fixes #336 

@ddollar thanks for your pointers today.  Originally I merged in the environment from the manifest, but it was missing `PROCESS`, `RELEASE`, etc.  Instead, I pulled in the environment from the ECS Task Definition.  Now attached runs match detached runs match the running task itself.

Tested locally as we discussed.

I'm still new to Go -- if there are more idiomatic ways to code this I'm happy to fix it!

### DIFF
--- a/api/models/app.go
+++ b/api/models/app.go
@@ -357,16 +357,32 @@ func (a *App) ExecAttached(pid, command string, rw io.ReadWriter) error {
 }
 
 func (a *App) RunAttached(process, command string, rw io.ReadWriter) error {
-	env, err := GetEnvironment(a.Name)
+	resources, err := a.Resources()
 
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


nzoschke:
Thanks for the PR. I'll start testing this now

nzoschke:
Thanks for the patch!

PR #341, opened 2016-02-11
Stack Prefixes

nzoschke:
Taking over from @xtfxme's PR: https://github.com/convox/rack/pull/328 

> Calls to GetApp and GetService attempt to resolve name then rack-name.
> 
> Patch introduces a new model helper, shortNameToStackName, and two new
methods, App.StackName and Service.StackName.

## Release Playbook
- [x] Rebase against master
- [x] Pass checks
- [ ] Code review
- [x] Release branch (20160223202128-stack-prefixes)
- [x] Pass CI (https://circleci.com/gh/convox/rack/570)
- [ ] Merge into master
- [ ] Release master
- [ ] Pass CI
- [ ] Publish release
- [ ] Release CLI


nzoschke:
@ddollar can I get a code review from you on this?

nzoschke:
This failed CI on the app convox-463-node-workers-463

```
11:00:46 UTC-0800	UPDATE_ROLLBACK_IN_PROGRESS	AWS::CloudFormation::Stack	convox-463-node-workers-463	The following resource(s) failed to create: [Balancer, BalancerRedisInternal].
11:00:44 UTC-0800	CREATE_FAILED	AWS::ElasticLoadBalancing::LoadBalancer	Balancer	Resource creation cancelled
11:00:43 UTC-0800	CREATE_FAILED	AWS::ElasticLoadBalancing::LoadBalancer	BalancerRedisInternal	LoadBalancer name cannot be longer than 32 characters
```

xtfxme:
I forgot to update the regex in `models/app.go`:

```go
var regexValidAppName = regexp.MustCompile(`\A[a-zA-Z][-a-zA-Z0-9]{3,29}\z`)
```

This should be shortened to only allow 24 chars instead of 29, this gives 7 chars for a rack name + "-" + 24 app name = 32 ELB name... should we also enforce short rack names?

xtfxme:
Could also get fancy and generate the limit, eg. `limit = 31 - len(os.Getenv("RACK"))`

nzoschke:
For starters, I'm changing the CI test harness... convox-463-node-workers-463 ->  c463-node-workers-463.

In reviewing the problem, the ELB name also takes stuff from the manifest, so even with enforcing the rack name and app name you could still run into problems.

Yes, it makes sense to change the app name validation to account for the `convox-` prefix we expect in most cases.

Validating the rack name length probably makes sense too. If someone calls their rack 'convox-production-microservices' they are going to have a bad time.

nzoschke:
That said, the CF limit for stack names is 128 characters. And it's annoying to worry about rack and app names because of the load balancer limitation.

xtfxme:
Yeah I read that too... 32 limit seems insanely small to me. Originally I was hoping to keep the ELB name short but it needed the extra info to make it unique within a region. Maybe ELB names should just be completely or partially random instead of using the StackName and whatever else?

xtfxme:
Also worth noting that a future Route53 integration would be able to mask random ELB names behind pretty CNAMEs.

nzoschke:
Yeah really seeing that the ELB name format should change...

nzoschke:
But I'm going to wait and see that this PR passes CI before addressing any more ELB stuff: https://circleci.com/gh/convox/rack/464

nzoschke:
CI is failing on `convox run` calls. Investigating... https://circleci.com/gh/convox/rack/464

xtfxme:
Hrm, not much info as to why. I tried something similar on my dev rack, eg:

```shell
# convox ps --app barebones
ID            NAME  RELEASE      SIZE  STARTED     COMMAND                                     
48dbca0a7985  main  RZIRJCCRKWC  256   2 days ago  /bin/sh -c 'ncat -l 2000 -k -e /usr/bin/env'

# convox run --app barebones main echo hello world
hello world
```

Which seems to be working. My initial guess is it's messing up on there being 3 `web` processes? I'm currently only testing on a rack with 1 host instance (therefore only 1 app instance) for speed.

nzoschke:
I haven't found a smoking gun yet, but I am seeing 100% failures / inconsistency on this PR where that hasn't been happening on other PRs / releases. I'll have to keep digging in as part of the release process. 

https://circleci.com/gh/convox/rack/tree/stack-prefixes

nzoschke:
@ddollar we could use some design input on the ELB name thing. Is it design goal to preserve the current naming scheme? If not do you have a design proposal for a more robust scheme?

If we can change it, this seems like the perfect opportunity to fix this, since this PR introduces a new set of AWS logic for new apps.

xtfxme:
My suggestion re:ELB names is `process-random`, eg. `api-abcd1234`. The randomness could be a new param to the app stack. Future code needing app-specific random bits can use it directly or truncate to an appropriate size. We should use something like `sha256(StackName)` that can be safely truncated (vs. sha1, which cannot).

Adding the process name to the beginning makes it easier to identify in your browser and elsewhere. The pseudo-algorithm would look something like `elbName = sprintf("%s-%s", trunc(processName, 23), trunc(hashName, 7))`.

xtfxme:
I said random then started talking about hashes... given the general design of what I said above, is it necessary for the random bits to be derived from the app in some way (eg. a hash) or is a real random number fine? Random number as a new param seems the best to me, but maybe there is some benefit from the predictability of `hash(StackName)`.

ddollar:
I like the idea of the repeatable hash. It is critically important, however, that existing ELBs do not get renamed.

xtfxme:
Oh true. Unbound apps will need their CF templates to continue referencing the ELB name in the old way, for all future update operations. Is this correct? A conditional change in how the formation update is created?

nzoschke:
I fixed up CI some more, then rebased this and ran it to get a clearer picture. This is having consistent problems in us-west but not us-east...

https://circleci.com/gh/convox/rack/495
https://circleci.com/gh/convox/rack/493

```
+ convox run --app httpd-495 web echo 'hello world'

+ grep 'hello world' ci/track.sh ci/tests/example-app httpd returned exit code 1
```
```
+ convox run --app node-workers-495 web echo 'hello world'

+ grep 'hello world' ci/track.sh ci/tests/example-app node-workers returned exit code 1
```

One big difference between the two regions is that us-east uses ECR, where us-west uses the self-hosted registry.

nzoschke:
I'm running out of time for today. A next step is to install or update a rack in us-west to `20160213192554-stack-prefixes` and see what's up.

xtfxme:
OK, I must have missed some old registry stuff, I only tested on `us-east-1` with ECR.

Only thing in the codebase to reference `REGISTRY` instead of `REGISTRY_HOST`? Also uses `b.App` instead of `app.StackName()`:
https://github.com/convox/rack/blob/stack-prefixes/api/models/build.go#L373

Uses `r.App` instead of `app.StackName()`
https://github.com/convox/rack/blob/stack-prefixes/api/models/release.go#L300

nzoschke:
Good catch. I did suspect there would be subtle breakage now that stack name != app name. Testing it now with formation referencing StackName

nzoschke:
Still have a problem with build/tag/push/pull on the old registry path:

https://circleci.com/gh/convox/rack/504

RUNNING: docker tag -f httpd httpd-504/web
RUNNING: docker tag -f httpd-504/web c504-951219826.us-west-2.elb.amazonaws.com:5000/httpd-504-web:BCWVWPFPUSE
RUNNING: docker push c504-951219826.us-west-2.elb.amazonaws.com:5000/httpd-504-web:BCWVWPFPUSE
...
Pushing tag for rev [e2f96ec84f28] on {https://c504-951219826.us-west-2.elb.amazonaws.com:5000/v1/repositories/httpd-504-web/tags/BCWVWPFPUSE}

Promoting RJSQBERIAGP... ERROR: could not pull "c504-951219826.us-west-2.elb.amazonaws.com:5000/c504-httpd-504-web:BCWVWPFPUSE" ci/track.sh ci/tests/example-app httpd returned exit code 1

xtfxme:
See/pull: https://github.com/xtfxme/rack/commit/1c80668fc484887d3b3bd8be267c657b824f069b

I think we've been chasing a non-issue. In the original patch, `StackName()` was added to repo names to ensure uniqueness within the rack... but they were already unique within the rack! non-ECR racks have a dedicated, rack-local registry so there is no need to prefix the names. ECR changes this namespace to global, but since the app CF template provisions new repos by referencing the `StackName` anyway, everything Just Works. We can remove any notion of prefixing from registry operations since both ECR and non-ECR paths are not subject to namespace collisions.

nzoschke:
CI passed, so we have confidence that the image name revert is working both in us-east-1 and us-west-2.

nzoschke:
While I'm aware of scope creep, the primary process ELB logic has a pretty bad side effect: https://github.com/convox/rack/issues/354

We can use this as an opportunity to unwind that too. Apps under this new scheme do not need any special primary process naming logic.

xtfxme:
@nzoschke @ddollar please review-and-pull (`stack-prefixes` branch) to PR: https://github.com/xtfxme/rack/commit/0c947f87a5c689dea5b7ea9d7bbd35901d610d46

Adds `App.IsBound` and preserves current ELB name indefinitely for unbound apps.

Bound app ELB name changes from something like `my-what-a-handsome-little-rack-notmain` to `notmain-S73J3MQ`. This suffix is built from `base32(sha256(my-what-a-handsome-little-rack))[:7]` with a possible `-i`. The app name can consume the remaining chars, with a max of 32 total.

Let me know if any of this should change!

The diff makes it look like I butchered `Release.Formation` but really I just moved up the call to `GetApp` and indented the old-ELB "primary" related code :-)

xtfxme:
Truncating the process name can cause a double dash in the ELB name. I don't care, does anyone else?

Old-style ELB name generation is used even if there are no existing ELB names to actually preserve (eg. app is created but not deployed). Undeployed apps created before a rack upgrade will still use the old naming rules. It's not clear to me if this poses a problem, and changing the behavior means `Manifest` now has to also care if it's ever been deployed. Seems not worth it to me unless there is some need.

xtfxme:
Small update: https://github.com/xtfxme/rack/commit/c7497d5c6f0c0df8e4d2064225ca2a9242fc5973

Keeps the suffix stable (3QVBPWQ):

```
Endpoints  main-3QVBPWQ-0000000000.us-west-2.elb.amazonaws.com:2000 (main)
           notmain-3QVBPWQ-0000000000.us-west-2.elb.amazonaws.com:2000 (notmain)
```

Instead of (S73J3MQ and FAAOZR7):

```
Endpoints  main-S73J3MQ-0000000000.us-west-2.elb.amazonaws.com:2000 (main)
           notmain-FAAOZR7-0000000000.us-west-2.elb.amazonaws.com:2000 (notmain)
```

nzoschke:
I rebased and pulled in your changes. Unless I missed something, tests need to be updated again to fix:

```
# github.com/convox/rack/api/models
api/models/manifest_test.go:54: not enough arguments in call to LoadManifest
api/models/manifest_test.go:83: not enough arguments in call to LoadManifest
api/models/manifest_test.go:101: not enough arguments in call to LoadManifest
FAIL	github.com/convox/rack/api/models [build failed]
error: exit status 2
```

I would like to update the manifest tests to mostly test the new bound path, but also write at least one for the unbound path to catch regressions.

xtfxme:
Beh I forgot to update the `*_test.go` files... I suppose there is no way for me to claim I ran the tests then.

I can update those tomorrow if not already done.

xtfxme:
@nzoschke please see: https://github.com/xtfxme/rack/commit/4a3ee2688f55a4715e2a24bbd7f04beec478e20e

This only tests the bound variant.

ddollar:
let's ship as, we'll refactor later

xtfxme:
@nzoschke https://github.com/xtfxme/rack/commit/dffac3fe0e3d5adc7382d647b97ce91866ad1ca1

As discussed on slack but you might want to tweak some. Allows `ManifestEntry` to carry an `App` for later use. If the app is nil, then `LoadManifest` and `ManifestEntry` methods operate as if the app is unbound (like it used to). If app is not nil, then it's used to determine bound or unbound.

cmd/app/main.go
```diff
@@ -23,7 +23,7 @@ func main() {
 			die(err)
 		}
 
-		manifest, err = models.LoadManifest(string(data))
+		manifest, err = models.LoadManifest(string(data), nil)
```
nzoschke: @ddollar I'm really confused about where/how/why this is used. Can you shed some light? I'm also not sure if we can get away with nil here...

nzoschke:
Getting very close... My remaining questions:

- [ ] is cmd/app/main.go ok?
- [ ] can we finalize on `httpd-web-JBVEG3F`, where the last part is a short hash of the rack name?
- [ ] why don't the CF outputs reference the balancer resources?

cmd/app/main.go
ddollar: I'm pretty sure cmd/app can be removed

xtfxme:
I don't like the hash being only the rack name because then its not good enough to "ensure" no conflict. How about `APP-PROCESS-(hashbase32slice(app.StackName()))`?

This makes the hash part unlikely to collide with any other app, ever, and also keeps the suffix stable across processes. We can then truncate `APP` and `PROCESS` however we like without worry.

ddollar:
:+1: on hashing the whole stack name

nzoschke:
I see, a long app name could be troublesome. Makes sense and hopefully the final detail here!





On Mon, Feb 22, 2016 at 8:15 PM -0800, "David Dollar" <notifications@github.com> wrote:












 on hashing the whole stack name

xtfxme:
Eh, the truncate rules would have to favor the process name otherwise app names over 22 chars will cause conflict.

The alternative is to go back to an unstable suffix that uses all sources (process, app, rack). Even if the process name is favored, long process names will cause ELB name conflicts.

The hash input could be an impossible stack name too, like `RACK:APP:PROCESS` to prevent collisions from naming variations that match. I don't care too much if the suffix is unique to the process because it will be masked.

xtfxme:
Also it makes the most sense to me for the suffix to be unique to the process because ELB is inherently a process level thing.

It does look pretty when they all match but, shrug, they already have some random number from AWS.

nzoschke:
Ok thanks again for your help navigating this subtle decision. I settled on:

- hash(rack-app-process), guaranteed unique for every process type
- trunc(app-process), pretty and informative for the majority of apps/manifests

This scheme should avoid name collision side effects for all manifest inputs.

If we ever want to change it again, we can do so by adding another tag to app stacks to branch off of, e.g. `ResourceScheme=2`

xtfxme:
Can we make the delimiters something not possible to be in a stack name?

These rack-app-procs would hash the same (because part of the app name in one is now part of the proc name in the second, but they are the same hash input) and truncate too early to fix the hash collision:

```
# rack, aaaaaaaaaaaaaaapp, hai-process
rack-aaaaaaaaaaaaaapp-B32HASH(rack-aaaaaaaaaaaaaaapp-hai-process)
# rack, aaaaaaaaaaaaaapp-hai, process
rack-aaaaaaaaaaaaaapp-B32HASH(rack-aaaaaaaaaaaaaaapp-hai-process)
```

Using a delimiter that cannot be in a stack name results in this:


```
# rack, aaaaaaaaaaaaaaapp, hai-process
rack-aaaaaaaaaaaaaapp-B32HASH(rack:aaaaaaaaaaaaaaapp:hai-process)
# rack, aaaaaaaaaaaaaapp-hai, process
rack-aaaaaaaaaaaaaapp-B32HASH(rack:aaaaaaaaaaaaaaapp-hai:process)
```

I know it's a little contrived, but it's a possible collision and easy to fix right now.

xtfxme:
Awesome, once this latest update passes tests/CI I'm pretty sure we are golden! All potential problems I'm aware of have been addressed.

xtfxme:
:+1: Another job well done.

Issue #203, opened 2015-12-09
Document how to build and run a development API branch


nzoschke:
Related: Respect DockerImageApi parameter in ECS definitions https://github.com/convox/rack/pull/352

nzoschke:
Landing here https://github.com/convox/rack/pull/478

PR #352, opened 2016-02-16
Respect DockerImageApi parameter in ECS definitions

xtfxme:
Maybe there is a reason for this I'm missing, but as-is, I can't use the `DockerImageApi` param to test new convox api builds (by pointing it at my own ECR repos). This makes it hard to quickly switch a rack between builds for testing.

Side-note, why does this param have a `Development` condition? How does that come into play exactly?

### DIFF
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -1041,7 +1041,10 @@
               "SUBNETS": { "Fn::Join": [ ",", [ { "Ref": "Subnet0" }, { "Ref": "Subnet1" }, { "Ref": "Subnet2" } ] ] },
               "VPC": { "Ref": "Vpc" }
             },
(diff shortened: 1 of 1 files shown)


nzoschke:
The Development condition is a small utility for those of us doing lots of rack development. If you set that to "Yes" many additional Outputs are set, and these values can be copied into `convox/rack/.env` as environment variables. Then `convox start` will start a local rack similar to how it runs in AWS. But I don't think it's important for this parameter.

This looks like a nice fix for an oversight with DockerImageApi. I'll review this more closely shortly.

nzoschke:
Thanks! And thanks for figuring this out while docs are still lagging.

https://github.com/convox/rack/issues/203

PR #365, opened 2016-02-19
Fixes for problems updating to 20160218212805 and 20160218230849

nzoschke:
- [x] restore Encryption param until removing params is handled better
- [x] root cause kernel ELB replacement error
- [x] review app ELB replacement 
- [x] test update from 20160213224638
- [x] test existing app deploys
- [x] stop auto-publishing release for the time being
- [x] update release playbook to include testing upgrades
- [x] strip out removed parameters for rack update

## Later
- [ ] automated test for upgrade path
- [ ] notification on rack update rollback
- [ ] remove Encryption param after this release has trickled out enough to safely remove rack params. Never?

## Release Playbook
- [x] Rebase against master
- [x] Pass checks
- [x] Code review
- [x] Release branch (20160219233644-update-fix)
- [x] Pass CI (https://circleci.com/gh/convox/rack/531)
- [ ] Merge into master
- [ ] Release master
- [ ] Pass CI
- [ ] Publish release
- [ ] Release CLI


nzoschke:
Master is in a bad state due to removing the Encryption param. We can't remove params without a required release that handles it better.

ddollar:
Let's use the `formationParameters()` helper to strip out bad parameters rather than the manual blacklist and required release.

nzoschke:
The Balancer error comes from going from no Scheme (implicit internet-facing) to an explicit Scheme in kernel.json.

```
"Scheme": { "Fn::If": [ "PrivateApi", "internal", "internet-facing" ] },
```

A condition on the resource property should fix this...

nzoschke:
thanks for the tip about formationParameters, I was headed down the wrong path with the blacklist/whitelist.

ddollar:
If the rack params stuff is now solid backwards/forwards probably worth making this a required release.

mattmanning:
:+1:  LGTM
+1 for required release

nzoschke:
Yes this is exactly what required release is for. We need to get this out everywhere to best handle rack upgrades / rollbacks going forward

PR #371, opened 2016-02-21
wip: provider interface

ddollar:


nzoschke:
I think it would be best to carve off some serious sync time to get something like this through, vs doing it peace meal.

nzoschke:
I'm going to close this now that a big chunk of additional provider interface is landing on https://github.com/convox/rack/pull/530. 

Thanks for the initial work here, it's feeling like a massive improvement to the code base.

PR #380, opened 2016-02-24
Introduce Service.IsBound

xtfxme:
As discussed, this is simply a followup to #341 that adds `Service.IsBound` similar to `App.IsBound`.


nzoschke:
:+1: Thanks.

Issue #355, opened 2016-02-17
/apps/<appname>/processes API call hangs


mattmanning:
```
$ curl -v "convox:<redacted>@convox-dev-128493365.us-east-1.elb.amazonaws.com/apps/feedchirp/processes"
*   Trying 54.175.204.251...
* Connected to convox-dev-128493365.us-east-1.elb.amazonaws.com (54.175.204.251) port 80 (#0)
* Server auth using Basic with user 'convox'
> GET /apps/feedchirp/processes HTTP/1.1
> Host: convox-dev-128493365.us-east-1.elb.amazonaws.com
> Authorization: Basic <redacted>
> User-Agent: curl/7.43.0
> Accept: */*
>
```

This never returns ^

mattmanning:
More info: This returns if I add the query string `?stats=false` to the URL. Looks like stats=true is the default and is broken. `convox ps --stats` from the CLI is broken too.

nzoschke:
I can't yet reproduce the hanging on testing racks, even with a ton of processes.

bitwalker:
@nzoschke @xtfxme and I are seeing this on our racks, though I have yet to pin it down to a specific sequence of events - just wanted to at least let you know that we're seeing this as well. We don't have the latest from master on our racks yet (I'd have to double check with @xtfxme to know for sure), so it's possible this has already been addressed.

nzoschke:
@awsmsrc and I are working on a fix here:

https://github.com/convox/rack/pull/394

The root cause appears to be the done channel hanging sporadically on the Docker client Stats.

I'll update close this ticket when the patch lands.

nzoschke:
This is fixed and released in 20160302193502

PR #384, opened 2016-02-25
PS Throttle

nzoschke:
Addresses this error which happens frequently on bigger apps:

```
$ convox ps
ERROR: ThrottlingException: Rate exceeded
    status code: 400, request id: 155f75b2-dac8-11e5-8038-cb0464aa31e7
```

* Increase ECS client retries to better deal with throttling
* Describe ECS instances only once
* Describe EC2 instances only once

This results in less requests total, and more time to backoff/retry if throttled.


nzoschke:
I spent a short timebox to see if I could also fix the `convox ps --stats` hanging bug (https://github.com/convox/rack/issues/355) with no success.

But I think the same strategy could help, fetching docker container info up front instead of during the fanout.

awsmsrc:
The go code looks :+1:

PR #385, opened 2016-02-25
Incremental builds

ddollar:
* scan local directory and take sha256 of every file
* ask server which hashes it doesn't know about
* upload the file content for the missing hashes
* capture current directory as a manifest of hashes to (filename, mode, mtime)
* upload manifest to server
* recreate source directory on the server side using manifest and hashed files
* build using that

## Release Playbook
- [x] Rebase against master
- [x] Pass checks
- [x] Release branch (20160226025613-incremental-builds)
- [x] Pass CI (https://circleci.com/gh/convox/rack/589)
- [x] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Publish release
- [ ] Release CLI


ddollar:
`20160225030413-incremental-builds`

api/controllers/builds.go
```diff
@@ -134,7 +135,30 @@ func BuildCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
...
+		fmt.Printf("build %+v\n", build)
+
+		if err != nil {
+			return httperr.Server(err)
+		} else {
+			return RenderJson(rw, build)
+		}
+	}
+
+	return httperr.Errorf(403, "no source or repo")
```
nzoschke: Did you intend to remove the tracking here?

ddollar: No. This is all new code so it just never had it.

api/models/build.go
```diff
@@ -284,6 +285,58 @@ func (b *Build) ExecuteRemote(repo string, cache bool, config string, ch chan er
...
+	args = append(args, b.App, "-")
+
+	err = b.execute(args, bytes.NewReader(tgz), ch)
+
+	if err != nil {
+		b.buildError(err, ch)
+		return
+	}
+
+	NotifySuccess("build:create", map[string]string{"id": b.Id, "app": b.App})
```
nzoschke: I'd like to TrackSuccess here too to help measure QoS for this path

ddollar: Added

api/controllers/builds.go
ddollar: Added

nzoschke:
The outline sounds solid. This is a lot like [Anvil](https://github.com/ddollar/anvil), right? That worked very well.  

I'm having a hard time grokking all the code on my first pass. Seems like getting the details perfect is really important.

Can you add some tests that demonstrate and validate how the protocol and API works?

It doesn't block this strategy, but my idea for efficient data transfer is to use the standard git protocol. We could provision a CodeCommit repo for every app, and use an single `httputil.NewSingleHostReverseProxy` handler to proxy with the right auth.

* Extremely efficient transport via the Smart HTTP Git protocol
* Application data encrypted at rest
* Standard .gitignore

xtfxme:
Is there some kind of sync library in Go? eg. an `rsync`-like thing (or git as previously noted)? It seems unfortunate to not lean on some existing, well-established file-transfer proto.

nzoschke:
This is always my goal too:

> lean on some existing, well-established $X

This protocol and implementation is simple and does have prior art in Anvil so theres no question it works.

But if there is a well-established client and server we could use I'll always push for that over writing a custom client and API.

ddollar:
@nzoschke just for clarity are you asking to block the PR on those tests?

nzoschke:
If this is an experiment it's ok without tests

nzoschke:
is there a way to fall back to the tarball path?

ddollar:
the api path still exists so old clients will continue to use it

there's no cli flag or anything to toggle

nzoschke:
I think that deserves more consideration... 

There's an API compatibility issue there (and the version check stuff is removed completely now?) And if a bug shows up in this path, it'd be nice to have the option of going back to the old path

ddollar:
Yep. All of the server side bits are still there. You could grab an older CLI release or we could quick-release a reversion to the CLI if that became necessary.

The version check stuff deserves more thought. We need something much lighter-weight.

ddollar:
Actually once this has a release number I can just make it use the old client code path automatically if the rack version is too old.

ddollar:
Added a `--classic` option for old-style builds. Will remove in a few weeks.

PR #391, opened 2016-02-26
add a helpful error message when trying to set vars with `convox env`

ddollar:
As noted by @ukd1


mattmanning:
:+1: Much nicer error.

Issue #356, opened 2016-02-17
Enable customization of Convox Rack instances


mwarkentin:
For various reasons, someone running a Convox Rack might want to be able to update the EC2 instances on the rack with their own customizations.

### Use cases

* Running an agent to help with monitoring, logging, etc.
  * New Relic Server Monitoring for 
  * SumoLogic for logs
  * etc.
* ?

#### Customized AMIs

Convox supports an AMI parameter which would allow for provisioning a new AMI on top of the base Amazon ECS AMIs, and then passing that in: `convox install --region chosen-region --ami chosen-ami-id --key chosen-ec2-key-name` (from [Using Weave and Scope with Convox](https://www.weave.works/using-weave-and-scope-with-convox/))

These AMIs could be generated any way the end user would want (Packer, manual updates and snapshot)

Pros:

* Already supported via CLI (might be nice to support in Grid - maybe only for paid accounts?)
* No provisioning required on instance launch
* More complex changes would probably be more maintainable

Cons:

* Heavyweight, needs a separate build pipeline for AMIs which is quite a bit more complex
* Need to rebuild when a new Convox / AWS ECS AMI is released

#### UserData customization

Convox already uses `UserData` (reference / details here) to install the Convox agent (anything else?). It could be possible to expose further `UserData` customizations to the end user, allowing them to provision extra customizations on instance launch.

Pros:

* Not supported right now
* Simpler for small changes (installing a single agent, etc.)

Cons:

* Would require testing on a rack after an upgrade to verify that your customizations are still compatible (might have the same issue w/ custom AMIs, but you might be able to catch at build time in your pipeline)
* Is there a practical limit to how many changes you can do via `UserData`?

mwarkentin:
Slack logs from initial discussion:

```
mwarkentin [11:25 AM] 
Any suggestions for customizations to the base image? Say we wanted to run the New Relic agent on each of our ECS instances for monitoring. One option that pops to mind would just be setting up a Packer pipeline that takes the latest official convox AMI and adds the agent - then I guess the question would be how to provide the new AMI to the convox Cloudformation templates?

mattmanning [11:27 AM] 
It would probably be a little more straightforward to use your Dockerfile(s) to install the agent in your containers rather than on the host instance. That wouldn’t require any AWS-level customization.

simon [11:34 AM] 
I think NewRelic recommends against installing their agent as a docker container - and to install it directly on the AWS instance itself

[11:35] 
[...]

PR #414, opened 2016-03-03
Agent Flags via InstanceBootCommand parameter

nzoschke:
* Use agent 0.66 (PR: https://github.com/convox/agent/pull/18)
* Add InstanceBootCommand Parameter
* Add set -e to UserData to guard against failures in InstanceBootCommand or anything else really

The new parameter opens up a new ability to configure individual racks differently. I want this to do agent experiments on a specific rack that is having problems at production scale. 

I.e. `InstanceBootCommand='curl -s https://convox.s3.amazonaws.com/agent/0.66-experimental/convox.conf > /etc/init/convox.conf ; restart convox'`

This also opens up lots of possibilities to configure a logging container on every instance, install extra packages, etc. See: https://github.com/convox/rack/issues/356

## Release Playbook
- [x] Rebase against master
- [x] Pass checks
- [x] Code review
- [ ] Merge into master
- [ ] Follow Multi-PR Release Playbook

### DIFF
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -853,6 +853,7 @@
         "UserData": { "Fn::Base64":
           { "Fn::Join": [ "\n", [
             "#!/bin/bash",
+            "set -e",
             "fallocate -l 5G /swapfile && chmod 0600 /swapfile && mkswap /swapfile && swapon /swapfile",
             "yum -y update",
             "yum -y install aws-cfn-bootstrap",
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -153,6 +153,11 @@
       "Description": "SSL certificate",
       "Default": ""
     },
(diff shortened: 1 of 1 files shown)


ddollar:
Perhaps a different parameter name that suggests it's just an addition? Completely replacing the Userdata should probably be an option at some point for people using custom AMIs.

nzoschke:
Good feedback. The AWS::CloudFormation::Init subsystem lets you set "Commands"

So something like `AdditionalBootCommand` or `ExtraInitCommand` could be more clear.

ddollar:
How about `InstanceBootCommand`? I can tell what that does without reading the description.

nzoschke:
Sounds good.

nzoschke:
set -e was blocking upgrades. I reverted that change.

PR #428, opened 2016-03-05
Datastore Service Options

nzoschke:
Fix up / augment options for datastore services (mysql, postgres, redis). 

Includes a patch from https://github.com/convox/rack/pull/451.

This introduces some template changes as well as a `convox services update` command. Users with data availability / integrity should be very careful with updates. Existing services are not modified under any circumstance other than `update`.

- [x] Add redis replication and multi-az options
- [x] Add postgres max connections option
- [x] Document all the supported options
- [x] Make engine versions explicit
- [x] Add --private to run in private subnets
- [x] Add an experimental `convox services update`
- [x] stop cargo culting CAPABILITY_IAM

## Demo

```
$ convox services create --help
convox services create: create a new service

Usage:
  convox services create <type> [--name=value] [--option-name=value]

Supported types / options:
  mysql       [--allocated-storage=10] [--instance-type=db.t2.micro] [--multi-az]
  papertrail  --url=logs1.papertrailapp.com:11235
  postgres    [--allocated-storage=10] [--instance-type=db.t2.micro] [--max-connections={DBInstanceClassMemory/15000000}] [--multi-az]
  redis       [--automatic-failover-enabled] [--instance-type=cache.t2.micro] [--num-cache-clusters=1]
  webhook     --url=https://grid.convox.com/rack-hook/1234

$ convox services create redis
Creating redis-5217 (redis)... CREATING

$ convox services update --help
convox services update: update a service.

WARNING: updates may cause service downtime.
...

$ convox services update redis-5217 --automatic-failover-enabled --instance-type=cache.m3.medium --num-cache-clusters=2 
Updating redis-5217 (automatic-failover-enabled="true" instance-type="cache.m3.medium" num-cache-clusters="2")... UPDATING

$ convox services update redis-5217 --foo=bar
Updating redis-5217 (foo="bar")...ERROR: No updates are to be performed.
```

## Release Playbook
- [x] Rebase against master
- [x] Pass checks
- [x] Release branch (20160317013909-services)
- [x] Pass CI
- [x] Code review
- [x] Merge into master
- [ ] Release master
- [ ] Pass CI
- [ ] Publish release
- [ ] Release CLI


nzoschke:
Another question... Certain sets of options parse just fine but eventually end up in a rolled back stack. For example with redis:

`$ convox services create redis --automatic-failover-enabled `

> Automatic failover is not supported for T1 and T2 cache node types.

Should the CLI validate these inputs? Or should the template have conditional logic around parameters like AutomaticFailoverEnabled?

mwarkentin:
@nzoschke I chatted about this quickly w/ @ddollar when I opened the initial ticket:

> 
mwarkentin [11:49 AM] 
Ok, created a ticket for it: https://github.com/convox/rack/issues/448 - might take a crack at the PR at some point as well

> [11:52] 
@ddollar: Had one question in there around parameter validation - is that something that you see happening in convox itself, or just rely on CloudFormation validation (I assume create/update stack will blow up if you set an invalid combination of params - eg. AutomaticFailoverEnabled = `True` && NumCacheClusters = `1`)

> ddollar [11:56 AM] 
yep, just lean on cloudformation

Seems reasonable to at least start with relying fully on CloudFormation for validation (at least to start).

nzoschke:
Yes, we're leaning on CF for everything. It does the initial parameter validation very well. The new errors I saw are more subtle in that the parameters validate, and the stack creation starts, but it fails after a minute into a rollback state.

Still since the service ends up in a 'failed' state I consider this acceptable for now too.

```
$ convox services create redis --automatic-failover-enabled
Creating redis-4799 (redis: automatic-failover-enabled="true")... CREATING

$ convox services
NAME                    TYPE      STATUS  
redis-4799              redis     failed  
```

nzoschke:
ElastiCache has the same concept of "MultiAZ" even though it's not a literal CF parameter.

screen shot 2016-03-16 at 7 54 39 am

It's tempting to have a `--multi-az` option that properly configures the CF stack with AutomaticFailoverEnabled and a minimum node type of cache.m3.medium.

nzoschke:
For redis i'm leaning very strongly towards the single `--multi-az` since its a first-class citizen across all datastores.

If `MultiAZ=false`, then the default `InstanceType=cache.t2.micro` and `NumCacheClusters=1`

If `MultiAZ=true`, then the default `InstanceType=cache.m3.medium` and `NumCacheClusters=2`.

I'm a bit split on the best way to implement this. 

The CLI or Controller could directly validate or select defaults for the set of MultiAZ, InstanceType and NumCacheClusters but then we move away from letting CloudFormation do everything.

The CF template could have some relatively straightforward Conditionals around the MultiAZ parameter. My main question here is how to handle defaults for NumCacheClusters and InstanceType. 

```
      "NumCacheClusters": {
        "Type": "Number",
        "Default": "2",
        "Description": "If MultiAZ, the number of cache clusters for this replication group. Minimum 2, maximum 5."
      },
```

```
      "InstanceType": {
        "Type": "String",
        "Default": "cache.t2.micro",
        "Description": "The type of instance to use."
      },
```

makes sense as the main default, but should the stack auto-magically pick `cache.m3.medium` if `MultiAZ=true`.

mwarkentin:
Definitely seems like a much more elegant way of enabling multi-az - seems reasonable to pick valid defaults if we want to keep it to a single `--multi-az` option. Might be worth adding a note in the help text if possible.

nzoschke:
Thanks for your input. multi-az redis is awesome and all motivated by your initial PR.

nzoschke:
Never mind, I learned in testing that NumCacheClusters can be used without AutomaticFailoverEnabled, so it makes sense to keep them all separate and explicit for now.

nzoschke:
This is looking good to me. I'm seeing a consistent problem on redis when switching between private not.

> 17:52:52 UTC-0700	UPDATE_FAILED	AWS::ElastiCache::SubnetGroup	CacheSubnetGroup	The subnet ID subnet-b5578fc3 is in use.

It rolls back just fine.

It works fine on postgres/mysql, albeit with an instance replacement.

Unless an obvious fix appears, I think the redis problem is acceptable for now.

nzoschke:
For the record, here's how to reproduce the redis rollback:

```
$ convox install
$ convox services create redis --name redis-foo
$ convox rack params set Private=true
$ convox services update redis-foo --private
```

The update will roll back because of the subnet is in use error.

ddollar:
:+1:

PR #434, opened 2016-03-09
Use file synchronization in place of volume mounts

ddollar:
This pull request changes our suggested mechanism for dealing with local file synchronization in development. We have noticed huge performance penalties for using VMWare or Virtualbox mounted volumes for file synchronization from the local machine and this aims to provide an alternative.

With these changes `convox start` will read your Dockerfile and automatically set up synchronization for any files referenced by `ADD` or `COPY` statements, eg: `COPY . /app` would set up synchronization from your local `.` directory to `/app` inside the container.

This synchronization works by watching for changes and then batching them up every 1s as `docker cp` tarballs.

This behavior is activated by including the `--sync` option to `convox start` or by creating a `.convox/sync` file with the contents `true`


api/manifest/manifest.go
```diff
@@ -17,10 +18,13 @@ import (
 	"regexp"
 	"sort"
 	"strings"
+	"sync"
+	"syscall"
 	"time"
 
 	"github.com/convox/rack/Godeps/_workspace/src/github.com/fatih/color"
 	yaml "github.com/convox/rack/Godeps/_workspace/src/gopkg.in/yaml.v2"
+	"github.com/fsouza/go-dockerclient"
```
nzoschke: godeps?

nzoschke:
I tried this and it isn't working for me. Seems like m.Run() doesn't return so the sync doesn't start.

ddollar:
Should be fixed. Also can now set a `.convox/sync` file to `true` to have it on by default.

nzoschke:
:+1:

PR #446, opened 2016-03-11
Try to cleanup S3 buckets on app deletion

anthonyrisinger:
## Release Playbook
- [x] Rebase against master
- [x] Pass checks
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Publish release
- [ ] Release CLI

### DIFF
--- a/api/models/app.go
+++ b/api/models/app.go
@@ -238,6 +238,11 @@ func (a *App) Cleanup() error {
 		if ae, ok := err.(awserr.Error); ok {
 			if ae.Code() == "ValidationError" {
 				helpers.TrackEvent("kernel-app-delete-success", nil)
+				// Last ditch effort to remove the empty bucket CF leaves behind.
+				_, err := S3().DeleteBucket(&s3.DeleteBucketInput{Bucket: aws.String(a.Outputs["Settings"])})
+				if err != nil {
+					fmt.Printf("error: %s\n", err)
+				}
 				return nil
 			}
 		}


anthonyrisinger:
There is a small race here on stack delete and emptying the bucket... if it takes a brillion years for `cleanupBucket` to empty the target bucket, and the stack is dropped **really** fast, there is a chance the bucket delete will fail with some nonempty error... but that just brings us back to where we are now :-)

anthonyrisinger:
Is there a reason apps spawn their own S3 buckets instead of using a prefix within a shared rack bucket? The default bucket limit is only 100 and before this patch we bumped that pretty quick just from development churn (add/deleting apps).

nzoschke:
If the bucket is not empty, the delete will always fail...

anthonyrisinger:
How long does it take for the other goroutines to clean the bucket? In practice so far it's been trying to drop the bucket long after emptying it, succeeding... I think. I only tried with apps deployed once. I'll look closer tomorrow.

Maybe apps with many more objects/logs would take too long?

What is the right thing?

nzoschke:
Is there another part that empties the bucket contents? I'm not seeing that so I was guessing this delete call would fail.

Good question about sharing a bucket and using prefixes. 100 isn't very many and seems like a pretty tough limit to get bumped up on AWS. I think we went for it because it was easy and gives lots of isolation between apps.

anthonyrisinger:
`App.Delete` calls `cleanupBucket` via `go a.Cleanup()` after a successful `DeleteStack` call. Since the stack takes time to delete, the goroutines dropping S3 objects seem to have enough time to drain everything before all the other resources in the stack are cleaned up by CF. `Cleanup` waits for the stack drop (up to 10 minutes) and if it happens, the bucket is dropped then too.

nzoschke:
Got it. Closing this and opening a new PR for this and a few other small features to release

nzoschke:
Closed on https://github.com/convox/rack/pull/488

PR #447, opened 2016-03-14
implement a racks command for listing racks

awsmsrc:
## Release Playbook
- [x] Rebase against master
- [x] Pass checks
- [x] Release branch ()
- [ ] Pass CI ()
- [x] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Publish release
- [ ] Release CLI

### DIFF
--- a/api/controllers/routes.go
+++ b/api/controllers/routes.go
@@ -46,6 +46,7 @@ func NewRouter() (router *mux.Router) {
 	router.HandleFunc("/instances", api("instances.get", InstancesList)).Methods("GET")
 	router.HandleFunc("/instances/{id}", api("instance.delete", InstanceTerminate)).Methods("DELETE")
 	router.HandleFunc("/instances/keyroll", api("instances.keyroll", InstancesKeyroll)).Methods("POST")
+	router.HandleFunc("/racks", api("rack.list", RegistryList)).Methods("GET")
 	router.HandleFunc("/registries", api("registry.list", RegistryList)).Methods("GET")
 	router.HandleFunc("/registries", api("registry.create", RegistryCreate)).Methods("POST")
 	// this should be DELETE /registries/{server} except for https://github.com/gorilla/mux/issues/132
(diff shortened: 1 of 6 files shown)


anthonyrisinger:
I like this ticket but I expected it to either list the racks from `~/.convox/auth` and/or call `DescribeStacks` and gather up the `Tags['Rack']` from the CF resources. The former would produce a list of racks known to the client and the latter would produce a region-bound list of racks known to the server.

api/controllers/routes.go
```diff
@@ -46,6 +46,7 @@ func NewRouter() (router *mux.Router) {
 	router.HandleFunc("/instances", api("instances.get", InstancesList)).Methods("GET")
 	router.HandleFunc("/instances/{id}", api("instance.delete", InstanceTerminate)).Methods("DELETE")
 	router.HandleFunc("/instances/keyroll", api("instances.keyroll", InstancesKeyroll)).Methods("POST")
+	router.HandleFunc("/racks", api("rack.list", RegistryList)).Methods("GET")
```
nzoschke: This is hooked up to the wrong handler

nzoschke:
Interesting expectations @anthonyrisinger.

There's another possibility which is all the racks registered in your Convox Console™, which is what @awsmsrc is laying the groundwork for.

anthonyrisinger:
Ah it looked just hard coded as a placeholder or something for now. "Expected" was only meant to mean "surprising" :-)

(whynotboth)? Dumping from `.convox/auth` would be pretty useful for us, if not connected to Console. I run `jq . ~/.convox/auth` so much I have an alias for it.

client/racks.go
```diff
@@ -0,0 +1,17 @@
+package client
+
+type Organization struct {
+	Id   string `json:"id"`
+	Name string `json:"name"`
+}
+
+type Rack struct {
```
nzoschke: We also have a struct that covers this in client.System. But I think that one is poorly named, so re-introducing a Rack struct is good.

nzoschke:
:+1:

nzoschke:
I understand your use case around reading .auth, and see how that could make more sense than this endpoint if you're not using our console @anthonyrisinger.

However we decided to go with the remote API that returns all the racks for now, and can re-visit any time later.

PR #469, opened 2016-03-21
Let CF implement SSL updates

anthonyrisinger:
Let CF perform the update instead of trying to do it manually. If all good, delete the old cert.

I just realized `UpdateStack` doesn't wait for the update, so a CF rollback could fail if the cert is deleted out-of-band... the delete goroutine maybe needs to wait for the stack update to complete before trying to drop the old cert? I should note this problem exists right now, this patch doesn't make things worse it just delegates the work to CF.

In the future it might be nice to just tag the cert (and maybe other resources) as orphaned and allow some future garbage collector (the monitor process? a cron process?) to clean it up. This would be nice for things like certificates because CF does not actually create them, therefore it does not clean them up. Same concept applies to S3 buckets that CF refuses to drop... they could be tagged as orphaned.

## Release Playbook
- [x] Rebase against master
- [x] Pass checks
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Publish release
- [ ] Release CLI

### DIFF
--- a/api/models/ssl.go
+++ b/api/models/ssl.go
@@ -16,9 +16,7 @@ import (
 
 	"github.com/convox/rack/Godeps/_workspace/src/github.com/aws/aws-sdk-go/aws"
 	"github.com/convox/rack/Godeps/_workspace/src/github.com/aws/aws-sdk-go/service/cloudformation"
-	"github.com/convox/rack/Godeps/_workspace/src/github.com/aws/aws-sdk-go/service/elb"
 	"github.com/convox/rack/Godeps/_workspace/src/github.com/aws/aws-sdk-go/service/iam"
-	"github.com/convox/rack/Godeps/_workspace/src/github.com/cenkalti/backoff"
 )
 
 type SSL struct {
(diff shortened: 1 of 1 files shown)


nzoschke:
Good catch on eliminating the manual ELB.SetLoadBalancerListenerSSLCertificate call.

I like the idea to tag and clean in a different thing.

What about not deleting the cert at all, since you recognized a potential rollback error in the current behavior?

anthonyrisinger:
Just leaving it works for me, will update.

I'll look into options for tagging arbitrary resources as orphaned. Also having a tag for unmanaged would prevent convox from adding orphan flag to resources it does not maintain (like a shared wildcard cert).

Is there a way to make CF automatically mark resources it unlinked but did not destroy? Or notify a gc process somewhere when done?

anthonyrisinger:
updated with comment.

nzoschke:
I don't think there's a way to have CF update resources on delete. I do think you can see from the CF api if resources were retained though....

nzoschke:
Closed on https://github.com/convox/rack/pull/488

PR #493, opened 2016-03-30
add mocking functionality to the test provider, eventually aws calls …

awsmsrc:


api/controllers/builds_test.go
```diff
@@ -0,0 +1,42 @@
...
+	}()
+
+	// setup expectations on current provider
+	testProvider.On("BuildDelete", "app-name", "build-id").Return(&testProvider.Build, nil)
+
+	// make request
+	body := test.HTTPBody("DELETE", "http://convox/apps/app-name/builds/build-id", nil)
+
+	// assert on expectations
+	testProvider.AssertExpectations(t)
```
awsmsrc: @ddollar fyi this is where the magic is. We can extract all calls to adoubleyouess from controllers and controller tests and utilize the provider interface

ddollar: nice

PR #529, opened 2016-04-08
Add support for passing existing IAM/ACM arn to convox ssl create/update commands

nduthoit:
## Release Playbook
- [ ] Rebase against master
- [ ] Pass checks
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Publish release
- [ ] Release CLI

### DIFF
--- a/api/controllers/ssl.go
+++ b/api/controllers/ssl.go
@@ -91,6 +91,7 @@ func SSLUpdate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	a := mux.Vars(r)["app"]
 	process := GetForm(r, "process")
 	port := GetForm(r, "port")
+	arn := GetForm(r, "arn")
 	chain := GetForm(r, "chain")
 	body := GetForm(r, "body")
 	key := GetForm(r, "key")
@@ -105,7 +106,7 @@ func SSLUpdate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 		return httperr.Errorf(403, "port must be numeric")
 	}
 
-	ssl, err := models.UpdateSSL(a, process, portn, body, key, chain)
+	ssl, err := models.UpdateSSL(a, process, portn, arn, body, key, chain)
 
 	if awsError(err) == "ValidationError" {
 		return httperr.Errorf(404, "%s", err)
--- a/api/controllers/ssl.go
+++ b/api/controllers/ssl.go
@@ -29,6 +29,7 @@ func SSLCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	a := mux.Vars(r)["app"]
 	process := GetForm(r, "process")
 	port := GetForm(r, "port")
+	arn := GetForm(r, "arn")
 	chain := GetForm(r, "chain")
 	body := GetForm(r, "body")
 	key := GetForm(r, "key")
@@ -44,7 +45,7 @@ func SSLCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 		return httperr.Errorf(403, "port must be numeric")
 	}
 
-	ssl, err := models.CreateSSL(a, process, portn, body, key, chain, (secure == "true"))
+	ssl, err := models.CreateSSL(a, process, portn, arn, body, key, chain, (secure == "true"))
 
 	if awsError(err) == "ValidationError" {
 		return httperr.Errorf(404, "%s", err)
--- a/api/controllers/ssl.go
+++ b/api/controllers/ssl.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/convox/rack/Godeps/_workspace/src/github.com/gorilla/mux"
 	"github.com/convox/rack/api/httperr"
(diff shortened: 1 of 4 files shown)


api/models/ssl.go
```diff
@@ -31,7 +31,7 @@ type SSL struct {
 
 type SSLs []SSL
 
-func CreateSSL(app, process string, port int, body, key string, chain string, secure bool) (*SSL, error) {
+func CreateSSL(app, process string, port int, arn string, body, key string, chain string, secure bool) (*SSL, error) {
```
mwarkentin: Wasn't clear to us immediately which args should have types attached to them (and why `body` doesn't have one).

api/models/ssl.go
```diff
@@ -74,11 +74,14 @@ func CreateSSL(app, process string, port int, body, key string, chain string, se
...
+	if arn == "" {
+		arn, err = uploadCert(a, process, port, body, key, chain)
 
-	if err != nil {
-		return nil, err
+		if err != nil {
+			return nil, err
+		}
 	}
+	// Check that the arn is valid?
```
mwarkentin: Could do some quick validation here that the string starts w/ `ARN::`, or just fall back on CloudFormation blowing up if you pass in an invalid ARN.

cmd/convox/ssl.go
```diff
@@ -256,18 +259,27 @@ func cmdSSLUpdate(c *cli.Context) {
...
+	case 3:
+		pub, err = ioutil.ReadFile(c.Args()[1])
 
-	if err != nil {
-		stdcli.Error(err)
-		return
-	}
+		if err != nil {
+			stdcli.Error(err)
+			return
```
mwarkentin: @nduthoit I think we messed up the formatting here.

api/models/ssl.go
nzoschke: Every arg does have a type. If an arg name doesn't have one immediately following it, the closest type to the right of it is its type.

It's some sugar so you can do (foo, bar, baz string) instead of (foo string, bar string, baz string).

So the idiomatic thing here would be to drop the ` string` after arn and key, since they're all the same type as chain.

cmd/convox/ssl.go
nduthoit: No @mwarkentin , the `if` statement is now within the `case 3:` block

api/models/ssl.go
nduthoit: Thanks @nzoschke, I've updated the function signatures

nzoschke:
Big +1 for supporting a pre-existing ARN since that opens up integration with other cert management strategies.

I'm starting to see everything though the "provider interface refactor" lens. I think a better interface for this would be to decouple a cert upload and cert update:

```go
SSLList(app string) (structs.SSLs, error) // list all IAM certificates that match an app either by cert name prefix or a resource tag
SSLCreate(app, body, key, chain string) (*structs.SSL, error) // upload a new IAM cert
SSLSet(app, process string, port int, secure bool, arn string) error // update app stack for new cert ARN
```

And make similar changes to the API, client and CLI.

```
$ convox ssl upload foo.pub foo.key
Added new SSL cert (arn:...)

$ convox ssl update arn:...
Updating app...
```

If you're interested in learning / helping with some provider interface stuff let me know and I can better flesh out a spec. I'm also happy accepting the simpler patch now and following up with provider interface later.

nduthoit:
@nzoschke agreed, decoupling cert upload and cert update makes sense. Happy to give it a shot once you have a more fleshed out spec.
Maybe we can use the current solution in the meanwhile to move ahead?

Thoughts on adding a check at the CLI level that the single argument passed is of the "arn:..." format? Just to avoid mistakenly falling in the arn case if only passing the public key for example.

nzoschke:
Yeah the current solution is a great for the interim.

Validation should happen in the controller so all API consumers get it. +1 for validating it looks like the right format.

nduthoit:
Sounds good, will add shortly.

mwarkentin:
@nzoschke Not familiar w/ the "provider interface refactor" pattern you mentioned - do you have a good link to an article / example?

My only question about your initial proposal - if I'm creating a listener using an existing cert, the verbiage is a bit strange: `convox ssl update web:80 arn:...`

`convox set ...` would seem to make sense both on an initial set up, and for updates.

`convox upload` would work in all cases as well.

mwarkentin:
+1 for changing commands in a second PR though, and getting this in as an interim solution to enable ACM.

api/models/ssl.go
mwarkentin: Should be easy w/ https://golang.org/pkg/strings/#HasPrefix

nduthoit:
I've rebased off the latest master

nduthoit:
@nzoschke @mwarkentin Updated with the check for valid ARN in the controller. Anything else missing?

api/controllers/ssl.go
```diff
@@ -90,6 +96,7 @@ func SSLUpdate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	a := mux.Vars(r)["app"]
 	process := GetForm(r, "process")
 	port := GetForm(r, "port")
+	arn := GetForm(r, "arn")
```
nzoschke: should this also validateARNFormat?

api/models/ssl.go
```diff
@@ -260,12 +263,15 @@ func UpdateSSL(app, process string, port int, body, key string, chain string) (*
...
+		// upload new cert
+		arn, err = uploadCert(a, process, port, body, key, chain)
 
-	if err != nil {
-		return nil, err
+		if err != nil {
+			return nil, err
+		}
 	}
+	// Check that the arn is valid?
```
nzoschke: can remove this comment

nzoschke:
This looks good to me. I can start working the release checklist soon

nzoschke:
:+1: Thanks! Going to test and release this on https://github.com/convox/rack/pull/530

PR #545, opened 2016-04-15
ProxyProtocol fix

### LINKED ISSUE #542: Proxy Protocol parameters reset on deployment
For an existing application, when I do this `convox apps params set AppPort443ProxyProtocol=Yes AppPort80ProxyProtocol=Yes` the parameters are set as expected, however when I do a `convox deploy` the parameters are changed back to ‘no’.

I have this in my docker-compose file:
```labels:
        - "com.convox.port.80.protocol=proxy"
        - "com.convox.port.443.protocol=proxy"

### PR DESCRIPTION
**Pull request by mattmanning**
- Call `mapping.Balancer` to get the port, rather than `mapping.Container`

- Drop the "com." from the labels. The format is now:

```
labels:
  - "convox.port.80.protocol=proxy"
```

using the string format or:

```
labels:
  convox.port.80.protocol: "proxy"
```

using the map format.

Fixes #542

## Release Playbook
- [x] Rebase against master
- [ ] Pass checks
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Publish release
- [ ] Release CLI

### DIFF
--- a/api/models/release.go
+++ b/api/models/release.go
@@ -181,10 +181,12 @@ func (r *Release) Promote() error {
 		for _, mapping := range entry.PortMappings() {
 			proxyParam := fmt.Sprintf("%sPort%sProxyProtocol", UpperName(entry.Name), mapping.Balancer)
 
-			switch entry.Label(fmt.Sprintf("convox.port.%s.protocol", mapping.Balancer)) {
-			case "proxy":
+			// "com.convox" backwards compatibility. This will be removed soon
+			oldLabel := entry.Label(fmt.Sprintf("com.convox.port.%s.protocol", mapping.Balancer))
+			newLabel := entry.Label(fmt.Sprintf("convox.port.%s.protocol", mapping.Balancer))
+			if (oldLabel == "proxy") || (newLabel == "proxy") {
 				app.Parameters[proxyParam] = "Yes"
-			default:
+			} else {
 				app.Parameters[proxyParam] = "No"
 			}
 		}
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ddollar** [comment]
There is another codepath that reads these labels that is used by start. Should drop the com from that one too.

**Comment by ddollar** [comment]
👍

**Comment by mattmanning** [comment]
After working on documentation and sleeping on this I kinda think removing the labels **should** revert the listeners back to the default config. Having users touch the app params directly feels like an abstraction leak.

What do you think @nzoschke @ddollar?

**Comment by nzoschke** [comment]
+1 for labels explicitly controlling the setting. Removing the label removes the proxy protocol setting.

**Comment by nzoschke** [comment]
I bet there are like two users of this feature right now, but changing the label will cause their setup to break. Can you still read the `com.convox` label for backwards compatibility and leave a comment that it can be removed in the future.

**Comment by mattmanning** [comment]
@nzoschke Added temporary backwards compatibility for deployments, but not for `convox start`. Seem reasonable?

**Comment by ddollar** [comment]
If it wasn't even working in production no need for the backwards-compatibility.

File: api/models/release.go [hunk 1]
```diff
@@ -181,10 +181,12 @@ func (r *Release) Promote() error {
 		for _, mapping := range entry.PortMappings() {
 			proxyParam := fmt.Sprintf("%sPort%sProxyProtocol", UpperName(entry.Name), mapping.Balancer)
 
-			switch entry.Label(fmt.Sprintf("com.convox.port.%s.protocol", mapping.Container)) {
```
ddollar: Let's keep the switch here. There will be other values for this label eventually. As mentioned in the other comment don't need the backwards compatibility since it wasn't even working before.

mattmanning:
@ddollar Reverted the backwards compatibility which also restores the switch. I had only removed it to make the backwards compatibility logic easier.

cliveza:
Thanks Gentleman, I took out the old version of the label and used the new one instead, so no need for backwards compatible code for me.

PR #522, opened 2016-04-07
amzn-ami-2016.03.a-amazon-ecs-optimized

nzoschke:
Follow up from https://github.com/convox/rack/pull/517. The new AMIs need some userdata changes to work properly.

PR #517, opened 2016-04-06
amzn-ami-2016.03.a-amazon-ecs-optimized

nzoschke:
Followup to https://github.com/convox/rack/pull/512, releasing with https://github.com/convox/rack/pull/509

PR #502, opened 2016-03-31
disable automatic updates of kernel* packages

robatwave:
## Release Playbook
- [ ] Rebase against master
- [ ] Pass checks
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update dev and testing racks
- [ ] Publish release
- [ ] Release CLI

Running 'yum update' upgrades the kernel and leaves the OS in an inconsistent state without a reboot and cleanup of old kernel packages. It also causes problems when InstanceBootCommand tries to install e.g. kernel-devel-$(uname -r) which will resolve to the older version of the kernel.

The proposed fix is to disable automatic updates of kernel* packages in yum's config.

This resolves #501

PR #550, opened 2016-04-16
amzn-ami-2016.03.a-amazon-ecs-optimized + cloud-init

bobzoller:
@nzoschke I took a pass at #522 by using cloud-init as you hinted in #517.

cloud-init's `bootcmd` runs _before_ `write_files`, which made for a much gnarlier patch.  `runcmd` however, runs after `write_files` but before services start, so it seems we can get away without `bootcmd`.

I chose to keep `InstanceBootCommand` and the `cfn-{init,signal}` in a separate file that runs after services are started, which preserves the existing behavior.

I could have gone a bit more minimal on the patch by leaving the swap and yum commands as bash, but I chose to leverage cloud-init's features instead.  Happy to reconsider that.

fixes #522 

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Edit release record in GitHub
- [ ] Publish release
- [ ] Release CLI

### DIFF
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -13,14 +13,14 @@
   },
   "Mappings": {
     "RegionConfig": {
-      "us-east-1": { "Ami": "ami-33b48a59" },
-      "us-west-1": { "Ami": "ami-26f78746" },
-      "us-west-2": { "Ami": "ami-65866a05" },
-      "eu-west-1": { "Ami": "ami-77ab1504" },
-      "eu-central-1": { "Ami": "ami-341efb5b" },
-      "ap-northeast-1": { "Ami": "ami-b3afa2dd" },
-      "ap-southeast-1": { "Ami": "ami-0cb0786f" },
-      "ap-southeast-2": { "Ami": "ami-cf6342ac" }
+      "us-east-1": { "Ami": "ami-67a3a90d" },
+      "us-west-1": { "Ami": "ami-b7d5a8d7" },
+      "us-west-2": { "Ami": "ami-c7a451a7" },
+      "eu-west-1": { "Ami": "ami-9c9819ef" },
+      "eu-central-1": { "Ami": "ami-9aeb0af5" },
+      "ap-northeast-1": { "Ami": "ami-7e4a5b10" },
+      "ap-southeast-1": { "Ami": "ami-be63a9dd" },
+      "ap-southeast-2": { "Ami": "ami-b8cbe8db" }
     }
   },
   "Outputs": {
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -865,12 +865,9 @@
             "swap:",
             "  filename: /swapfile",
             "  size: 5000000000",
-            "write_files:",
-            "- path: /opt/convox/runcmd.sh",
-            "  permissions: '0755'",
-            "  content: |",
-            { "Fn::Join" : [ "\n      ", [
-              "      #!/bin/bash",
+            "bootcmd:",
+            { "Fn::Join" : [ " ; ", [
+              "- /usr/bin/cloud-init-per once convox_bootcmd false || ( true",
               { "Fn::Join": [ "", [ "echo ECS_CLUSTER=", { "Ref": "Cluster" }, " >> /etc/ecs/ecs.config" ] ] },
               "echo ECS_ENGINE_AUTH_TYPE=docker >> /etc/ecs/ecs.config",
               { "Fn::If": [ "BlankCertificate",
@@ -885,20 +882,18 @@
               { "Fn::Join": [ "", [ "echo \"", { "Ref": "AWS::Region" }, "\" > /etc/convox/region" ] ] },
               { "Fn::Join": [ "", [ "echo \"", { "Ref": "Kinesis" }, "\" > /etc/convox/kinesis" ] ] },
               { "Fn::Join": [ "", [ "echo \"", { "Ref": "LogGroup" }, "\" > /etc/convox/log_group" ] ] },
-              "curl -s https://convox.s3.amazonaws.com/agent/0.67/convox.conf > /etc/init/convox.conf"
+              "curl -s https://convox.s3.amazonaws.com/agent/0.67/convox.conf > /etc/init/convox.conf",
+              ")"
             ] ] },
-            "- path: /var/lib/cloud/scripts/per-once/convox.sh",
-            "  permissions: '0755'",
-            "  content: |",
-            { "Fn::Join" : [ "\n      ", [
-              "      #!/bin/bash",
+            "runcmd:",
+            { "Fn::Join" : [ " ; ", [
+              "- /usr/bin/cloud-init-per once convox_runcmd false || ( true",
               { "Ref": "InstanceBootCommand" },
               { "Fn::Join": [ " ", [ "/opt/aws/bin/cfn-init", "-s", { "Ref": "AWS::StackName" }, "-r", "Instances", "--region", {"Ref":"AWS::Region"} ] ] },
               "sleep 30",
-              { "Fn::Join": [ " ", [ "/opt/aws/bin/cfn-signal", "--stack", { "Ref": "AWS::StackName" }, "--resource", "Instances", "--region", {"Ref":"AWS::Region"} ] ] }
-            ] ] },
-            "runcmd:",
-            "- [ cloud-init-per, once, convox_runcmd, /opt/convox/runcmd.sh ]"
+              { "Fn::Join": [ " ", [ "/opt/aws/bin/cfn-signal", "--stack", { "Ref": "AWS::StackName" }, "--resource", "Instances", "--region", {"Ref":"AWS::Region"} ] ] },
+              ")"
+            ] ] }
           ] ] }
         }
       }
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -858,6 +858,8 @@
           { "Fn::Join": [ "\n", [
             "#cloud-config",
             "repo_upgrade: all",
+            "repo_upgrade_exclude:",
+            "- kernel*",
             "packages:",
             "- aws-cfn-bootstrap",
             "swap:",
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -883,12 +883,12 @@
               { "Fn::Join": [ "", [ "echo \"", { "Ref": "Kinesis" }, "\" > /etc/convox/kinesis" ] ] },
               { "Fn::Join": [ "", [ "echo \"", { "Ref": "LogGroup" }, "\" > /etc/convox/log_group" ] ] },
               "curl -s https://convox.s3.amazonaws.com/agent/0.67/convox.conf > /etc/init/convox.conf",
+              { "Ref": "InstanceBootCommand" },
               ")"
             ] ] },
             "runcmd:",
             { "Fn::Join" : [ " ; ", [
               "- /usr/bin/cloud-init-per once convox_runcmd false || ( true",
-              { "Ref": "InstanceBootCommand" },
               { "Fn::Join": [ " ", [ "/opt/aws/bin/cfn-init", "-s", { "Ref": "AWS::StackName" }, "-r", "Instances", "--region", {"Ref":"AWS::Region"} ] ] },
               "sleep 30",
               { "Fn::Join": [ " ", [ "/opt/aws/bin/cfn-signal", "--stack", { "Ref": "AWS::StackName" }, "--resource", "Instances", "--region", {"Ref":"AWS::Region"} ] ] },
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -5,6 +5,7 @@
     "BlankAmi": { "Fn::Equals": [ { "Ref": "Ami" }, "" ] },
     "BlankCertificate": { "Fn::Equals": [ { "Ref": "Certificate" }, "" ] },
     "BlankDockerImageApi": { "Fn::Equals": [ { "Ref": "DockerImageApi" }, "" ] },
+    "BlankInstanceBootCommand": { "Fn::Equals": [ { "Ref": "InstanceBootCommand" }, "" ] },
     "BlankKey": { "Fn::Equals": [ { "Ref": "Key" }, "" ] },
     "BlankRegistryHost": { "Fn::Equals": [ { "Ref": "RegistryHost" }, "" ] },
     "Development": { "Fn::Equals": [ { "Ref": "Development" }, "Yes" ] },
@@ -883,7 +884,10 @@
               { "Fn::Join": [ "", [ "echo \"", { "Ref": "Kinesis" }, "\" > /etc/convox/kinesis" ] ] },
               { "Fn::Join": [ "", [ "echo \"", { "Ref": "LogGroup" }, "\" > /etc/convox/log_group" ] ] },
               "curl -s https://convox.s3.amazonaws.com/agent/0.67/convox.conf > /etc/init/convox.conf",
-              { "Ref": "InstanceBootCommand" },
+              { "Fn::If": [ "BlankInstanceBootCommand",
+                "true",
+                { "Ref": "InstanceBootCommand" }
+              ] },
               ")"
             ] ] },
             "runcmd:",
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -856,32 +856,47 @@
         "SecurityGroups": [ { "Ref": "SecurityGroup" } ],
         "UserData": { "Fn::Base64":
           { "Fn::Join": [ "\n", [
-            "#!/bin/bash",
-            "fallocate -l 5G /swapfile && chmod 0600 /swapfile && mkswap /swapfile && swapon /swapfile",
-            "echo 'exclude=kernel*' >> /etc/yum.conf",
-            "yum -y update",
-            "yum -y install aws-cfn-bootstrap",
-            { "Fn::Join": [ "", [ "echo ECS_CLUSTER=", { "Ref": "Cluster" }, " >> /etc/ecs/ecs.config" ] ] },
-            "echo ECS_ENGINE_AUTH_TYPE=docker >> /etc/ecs/ecs.config",
-            { "Fn::If": [ "BlankCertificate",
-              { "Fn::Join": [ "", [ "echo 'ECS_ENGINE_AUTH_DATA={\"index.docker.io\":{\"username\":\"\",\"password\":\"\",\"email\":\"\"},\"", { "Fn::Join": [ ":", [ { "Fn::GetAtt": [ "Balancer", "DNSName" ] }, "5000" ] ] }, "\":{\"username\":\"convox\",\"password\":\"", { "Ref": "Password" }, "\",\"email\":\"user@convox.io\"}}' >> /etc/ecs/ecs.config" ] ] },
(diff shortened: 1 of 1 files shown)


ddollar:
:+1: Have not actually booted this up in my Rack yet but this looks great. Thanks for the patch!

nzoschke:
This is a great patch, thank you. I'll work on testing and merging this this week.

bobzoller:
No problem!  I may push it to my racks and see if it helps with the
instance churn this week.  I'll let you know if I develop an opinion.

PS I left out the yum exclude 'kernel*` because I noticed the base AMI
excludes `kernel` (no star) already.  Easy to add back if both are desired.

mwarkentin:
@bobzoller The `kernel*` change was added here: https://github.com/convox/rack/pull/502

We found that it was causing issues installing sysdig without that line.

bobzoller:
Thanks for the pointer, @mwarkentin.  It does appear the built-in exclude of `kernel` predated this new ECS base image, so I'll guess adding `kernel*` is still necessary.  Added it back in 727bcd7.

nzoschke:
I rebased this and pushed it back to https://github.com/convox/rack/pull/522. Cutting a release...

nzoschke:
Starting to test this... Seeing some things that aren't quite right:

* convox/agent isn't running after boot
* `convox ps` returns a `ERROR: cannot connect to Docker endpoint`
* @mwarkentin reported "TypeError: Can not create provides table: disk I/O error" with InstanceBootCommand

nzoschke:
Seems like `start convox` got lost but that command also doesn't seem to work.

Also a note for myself: I lost agent 0.67 when I rebased on the ami branch https://github.com/convox/rack/pull/522/files

bobzoller:
`start convox` does work for me, but you're right I lost that command
(thought upstart would start it for us...)

bobzoller:
amazon uses `start on started ecs` to a slightly different end in [this doc](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/start_task_at_launch.html), and I suggest making that change to convox/agent's [/etc/init/convox.conf:3](https://github.com/convox/agent/blob/master/convox.conf#L3) as a backward-compatible fix.  no other changes to this PR are required.

bobzoller:
as for the `InstanceBootCommand` question, I chose run it _after_ services like docker have started to preserve the existing behavior, but it could be moved into the `runcmd` if we wanted to give folks a chance at the system before services start.

bobzoller:
rebased off master to restore agent 0.67

bobzoller:
`runcmd` is happening after services start.  I'll work on using `bootcmd` instead.

bobzoller:
@nzoschke was right -- `start on started docker` totally works.  the problem was cloud-init's `runcmd` fires after docker starts, so the agent never starts.  In 1067598 I found a not horrible way to move things into `bootcmd`.

verified my convox agent starts now.

mwarkentin:
@bobzoller How are you using your branch for your rack? Curious if I can help test this without bugging Noah for another release. 😄

bobzoller:
Manually updating the cloudformation using the AWS web UI and the
kernel.json from my branch.
On Apr 19, 2016 7:35 PM, "Michael Warkentin" <notifications@github.com>
wrote:

@bobzoller <https://github.com/bobzoller> How are you using your branch for
your rack? Curious if I can help test this without bugging Noah for another
release. 😄

—
You are receiving this because you were mentioned.
Reply to this email directly or view it on GitHub
<https://github.com/convox/rack/pull/550#issuecomment-212218825>

bobzoller:
@nzoschke I think this is ready for another look.  I did just move the `InstanceBootCommand` into the `bootcmd` section, so it has the potential to screw existing usages up (because services will not be started yet where they were before), but does seem like the original intent of the feature (and would likely resolve whatever disk issues @mwarkentin mentioned).  Up to you whether to include it.

mwarkentin:
I'd be happy to run some more tests if we can get another release cut.

nzoschke:
Ok taking another look at this with the intent of getting it released

mwarkentin:
@nzoschke We're still seeing problems with our `InstanceBootCommand` when I update our stack with this `kernel.json` - not really sure how to debug further at this point. Any chance we could schedule some time tomorrow to either work through it on Slack or maybe do a hangout/screenshare so I can show you what we've got happening?

nzoschke:
Yes, lets schedule some synchronous time... Hangout would be cool

nzoschke:
- [ ] Make final call on where InstanceBootCommand runs
- [ ] Go through branch release checklist

awsmsrc:
So we're going to cut a release *as is* and test.
My proposal regarding runcmd vs bootcmd is to have some sync time with @mwarkentin and see if we can get the script working where it is and if not add another hook for runcmd. That way we should definitely be able to solve it as well as having more functionality :)

nzoschke:
+1 leaving InstanceBootCommand in the cloudinit bootcmd section.

nzoschke:
Moving back to this PR... Here's my latest diff that is almost working. Some key changes:

* Every command is a a yaml line
* Call mkswap, because I can't tell if the swap stuff works with devices
* Get rid of the `cfn-init` call. That isn't needed unless we use CloudFormation::Init metadata
* Get rid of the `start convox` and `service docker restart`

Some known regressions with this strategy:

* Didn't port the conditional around BlankCertificate yet
* Didn't port InstanceBootCommand or InstanceRunCommand
* SHOWSTOPPER: /etc/cloud/cloud.cfg.d/90_ecs.cfg boot_cmd is not getting called, so xvdcz is not used

```bash
[root@ip-10-0-1-206 cloud.cfg.d]# lsblk
NAME                                                                                     MAJ:MIN   RM  SIZE RO TYPE MOUNTPOINT
xvda                                                                                     202:0      0    8G  0 disk 
└─xvda1                                                                                  202:1      0    8G  0 part /
xvdb                                                                                     202:16     0    5G  0 disk [SWAP]
loop0                                                                                      7:0      0  100G  0 loop 
└─docker-202:1-263276-pool                                                               253:0      0  100G  0 dm   
  ├─docker-202:1-263276-823fecfc72e33209b880cb4d61fb3a464211c6c8d906006ac5758cbf3042324c 253:1      0  100G  0 dm   
  ├─docker-202:1-263276-4c22d1ab5cb18c082cc23b80b94e9f2e047aebf7f224777de152c051ce730a3a 253:2      0  100G  0 dm   
  ├─docker-202:1-263276-11e28e31cb1eec551c60e0b3a1248f56426f31a72c5b111ad7752eae6155c830 253:3      0  100G  0 dm   
  └─docker-202:1-263276-2194103d28d18f4a81f7de0e60000228a490c9ac3575a52148bbec73c0c80541 253:4      0  100G  0 dm   
loop1                                                                                      7:1      0    2G  0 loop 
└─docker-202:1-263276-pool                                                               253:0      0  100G  0 dm   
  ├─docker-202:1-263276-823fecfc72e33209b880cb4d61fb3a464211c6c8d906006ac5758cbf3042324c 253:1      0  100G  0 dm   
  ├─docker-202:1-263276-4c22d1ab5cb18c082cc23b80b94e9f2e047aebf7f224777de152c051ce730a3a 253:2      0  100G  0 dm   
  ├─docker-202:1-263276-11e28e31cb1eec551c60e0b3a1248f56426f31a72c5b111ad7752eae6155c830 253:3      0  100G  0 dm   
  └─docker-202:1-263276-2194103d28d18f4a81f7de0e60000228a490c9ac3575a52148bbec73c0c80541 253:4      0  100G  0 dm   
xvdcz                                                                                    202:26368  0   30G  0 disk 
```

```
diff --git a/api/dist/kernel.json b/api/dist/kernel.json
index 86529be..26683c3 100644
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -875,55 +875,31 @@
         "PlacementTenancy" : { "Ref": "Tenancy" },
         "SecurityGroups": [ { "Ref": "SecurityGroup" } ],
         "UserData": { "Fn::Base64":
-          { "Fn::Join": [ "\n", [
-            "#cloud-config",
-            "repo_upgrade: all",
-            "repo_upgrade_exclude:",
-            "- kernel*",
-            "packages:",
-            "- aws-cfn-bootstrap",
-            "fs_setup:",
-            "    - device: /dev/xvdb",
-            "      filesystem: swap",
-            "mounts:",
-            "    - ['/dev/xvdb', 'none', 'swap', 'sw', '0', '0']",
-            "bootcmd:",
-            { "Fn::Join" : [ " ; ", [
-              "- /usr/bin/cloud-init-per once convox_bootcmd false || ( true",
-              { "Fn::Join": [ "", [ "echo ECS_CLUSTER=", { "Ref": "Cluster" }, " >> /etc/ecs/ecs.config" ] ] },
-              "echo ECS_ENGINE_AUTH_TYPE=docker >> /etc/ecs/ecs.config",
-              { "Fn::If": [ "BlankCertificate",
-                { "Fn::Join": [ "", [ "echo 'ECS_ENGINE_AUTH_DATA={\"index.docker.io\":{\"username\":\"\",\"password\":\"\",\"email\":\"\"},\"", { "Fn::Join": [ ":", [ { "Fn::GetAtt": [ "Balancer", "DNSName" ] }, "5000" ] ] }, "\":{\"username\":\"convox\",\"password\":\"", { "Ref": "Password" }, "\",\"email\":\"user@convox.io\"}}' >> /etc/ecs/ecs.config" ] ] },
-                { "Fn::Join": [ "", [ "echo 'ECS_ENGINE_AUTH_DATA={\"index.docker.io\":{\"username\":\"\",\"password\":\"\",\"email\":\"\"},\"", { "Fn::Join": [ ":", [ { "Ref": "RegistryHost" }, { "Ref": "RegistryPort" } ] ] }, "\":{\"username\":\"convox\",\"password\":\"", { "Ref": "Password" }, "\",\"email\":\"user@convox.io\"}}' >> /etc/ecs/ecs.config" ] ] }
-              ] },
-              { "Fn::If": [ "BlankCertificate",
-                { "Fn::Join": [ "", [ "echo 'OPTIONS=\"${OPTIONS} --log-opt max-file=2 --log-opt max-size=50m --insecure-registry=", { "Fn::Join": [ ":", [ { "Fn::GetAtt": [ "Balancer", "DNSName" ] }, "5000" ] ] }, " --host=unix:///var/run/docker.sock --host=0.0.0.0:2376\"' >> /etc/sysconfig/docker" ] ] },
-                "echo 'OPTIONS=\"${OPTIONS} --log-opt max-file=2 --log-opt max-size=50m --host=unix:///var/run/docker.sock  --host=0.0.0.0:2376\"' >> /etc/sysconfig/docker"
-              ] },
-              "mkdir -p /etc/convox",
-              { "Fn::Join": [ "", [ "echo \"", { "Ref": "AWS::Region" }, "\" > /etc/convox/region" ] ] },
-              { "Fn::Join": [ "", [ "echo \"", { "Ref": "ClientId" }, "\" > /etc/convox/client_id" ] ] },
-              { "Fn::Join": [ "", [ "echo \"", { "Ref": "Kinesis" }, "\" > /etc/convox/kinesis" ] ] },
-              { "Fn::Join": [ "", [ "echo \"", { "Ref": "LogGroup" }, "\" > /etc/convox/log_group" ] ] },
-              "curl -s https://convox.s3.amazonaws.com/agent/0.68/convox.conf > /etc/init/convox.conf",
-              { "Fn::If": [ "BlankInstanceBootCommand",
-                "true",
-                { "Ref": "InstanceBootCommand" }
-              ] },
-              ")"
-            ] ] },
-            "runcmd:",
-            { "Fn::Join" : [ " ; ", [
-              "- /usr/bin/cloud-init-per once convox_runcmd false || ( true",
-              { "Fn::If": [ "BlankInstanceBootCommand",
-                "true",
-                { "Ref": "InstanceRunCommand" }
-              ] },
-              { "Fn::Join": [ " ", [ "/opt/aws/bin/cfn-init", "-s", { "Ref": "AWS::StackName" }, "-r", "Instances", "--region", {"Ref":"AWS::Region"} ] ] },
-              "sleep 30",
-              { "Fn::Join": [ " ", [ "/opt/aws/bin/cfn-signal", "--stack", { "Ref": "AWS::StackName" }, "--resource", "Instances", "--region", {"Ref":"AWS::Region"} ] ] },
-              ")"
-            ] ] }
+          { "Fn::Join": [ "", [
+            "#cloud-config\n",
+            "repo_upgrade: all\n",
+            "repo_upgrade_exclude:\n",
+            "  - kernel*\n",
+            "packages:\n",
+            "  - aws-cfn-bootstrap\n",
+            "mounts:\n",
+            "  - ['/dev/xvdb', 'none', 'swap', 'sw', '0', '0']\n",
+            "bootcmd:\n",
+            "  - mkswap /dev/xvdb\n",
+            "  - swapon /dev/xvdb\n",
+            "  - echo ECS_CLUSTER=", { "Ref": "Cluster" }, " >> /etc/ecs/ecs.config\n",
+            "  - echo ECS_ENGINE_AUTH_TYPE=docker >> /etc/ecs/ecs.config\n",
+            "  - echo 'ECS_ENGINE_AUTH_DATA={\"index.docker.io\":{\"username\":\"\",\"password\":\"\",\"email\":\"\"},\"", { "Fn::GetAtt": [ "Balancer", "DNSName" ] }, ":5000\":{\"username\":\"convox\",\"password\":\"", { "Ref": "Password" }, "\",\"email\":\"user@convox.io\"}}' >> /etc/ecs/ecs.config\n",
+            "  - echo 'OPTIONS=\"${OPTIONS} --log-opt max-file=2 --log-opt max-size=50m --host=unix:///var/run/docker.sock  --host=0.0.0.0:2376\"' >> /etc/sysconfig/docker\n",
+            "  - mkdir -p /etc/convox\n",
+            "  - echo \"", { "Ref": "AWS::Region" }, "\" > /etc/convox/region\n",
+            "  - echo \"", { "Ref": "ClientId" }, "\" > /etc/convox/client_id\n",
+            "  - echo \"", { "Ref": "Kinesis" }, "\" > /etc/convox/kinesis\n",
+            "  - echo \"", { "Ref": "LogGroup" }, "\" > /etc/convox/log_group\n",
+            "  - curl -s https://convox.s3.amazonaws.com/agent/0.68/convox.conf > /etc/init/convox.conf\n",
+            "runcmd:\n",
+            "  - sleep 30\n",
+            "  - /opt/aws/bin/cfn-signal --region ", {"Ref": "AWS::Region"}, " --stack ", {"Ref": "AWS::StackName"}, " --resource Instances\n"
           ] ] }
         }
       }
```

nzoschke:
@bobzoller what was the inception of the strategy of using `/usr/bin/cloud-init-per once convox_bootcmd false` ?

I think the docker volume problem is that the userdata boot_cmd is clobbering this:

```bash
# cat 90_ecs.cfg 
#cloud-config
system_info:
  default_user:
    groups: [ "wheel", "docker" ]
bootcmd:
 - [ cloud-init-per, instance, docker_storage_setup, /usr/bin/docker-storage-setup ]
```

bobzoller:
oh interesting.  everything makes sense now.

I thought cloud-init would merge that list, but it appears to be a whole
topic: http://cloudinit.readthedocs.org/en/latest/topics/merging.html

bootcmds and runcmds fire on every boot by default, so I was preserving
existing behavior using cloud-init-per.  since every command would have to
be prefixed with cloud-init-per, I chose to do the whole
join-with-semicolon-for-a-giant-single-command approach.

hope that explains things, feel free to ping me if I can help more

nzoschke:
I have to say I'm really underwhelmed by the cloud-init docs. The intro is confusing. Is combining the default behavior? I'm not seeing the combining happen, do I need activate it somehow?

mwarkentin:
On the other link you posted AWS said they were running on a custom version of cloud-init - wonder if that was "pre-merge"?

bobzoller:
could be.  @nzoschke maybe you can bring it back to AWS as they suggested
the `bootcmd` in the first place, right?

On Mon, Apr 25, 2016 at 3:28 PM, Michael Warkentin <notifications@github.com

nzoschke:
Yes, I have an AWS support ticket open around this, pointing to this PR. We'll get it sorted soon enough I'm sure.

nzoschke:
David had a good idea... We could try to write a file next to `/etc/cloud/cloud.cfg.d/90_ecs.cfg` with CloudInit vs specifying userdata

bobzoller:
I _think_ you'll find you'd have to do it in a custom AMI -- `write_files`
is run after `bootcmd`.

nzoschke:
Ahh, you're probably right.

nzoschke:
I really don't want to build and maintain custom AMIs if at all possible.

ddollar:
Did you try this? `merge_how: 'dict(recurse_array,no_replace)+list(append)'`

http://jen20.com/2015/10/04/cloudconfig-merging.html

nzoschke:
I did not as I couldn't grok the cloudinit docs. That pointer looks pretty good, I'll give it a try soon.

nzoschke:
The merge thing didn't work on first try... :(

```diff
+          { "Fn::Join": [ "", [
+            "#cloud-config\n",
+            "merge_how: 'dict(recurse_array,no_replace)+list(append)'\n",
+            "repo_upgrade: all\n",
+            "repo_upgrade_exclude:\n",
```

nzoschke:
A couple other ideas...

copy ` - [ cloud-init-per, instance, docker_storage_setup, /usr/bin/docker-storage-setup ]` into the bootcmd

Consider using one of the other steps:

```
# The modules that run in the 'init' stage
cloud_init_modules:
  - migrator
  - rsyslog
  - bootcmd
  - write-files
  - write-metadata
  - growpart
  - resizefs
  - set-hostname
  - update-hostname
  - update-etc-hosts
  - resolv-conf
  - users-groups
  - ssh

# The modules that run in the 'config' stage
cloud_config_modules:
  - mounts
  - locale
  - set-passwords
  - yum-configure
  - yum-add-repo
  - package-update-upgrade-install
  - timezone
  - puppet
  - disable-ec2-metadata
  - runcmd

# The modules that run in the 'final' stage
cloud_final_modules:
  - scripts-per-once
  - scripts-per-boot
  - scripts-per-instance
  - scripts-user
  - ssh-authkey-fingerprints
  - keys-to-console
  - phone-home
  - final-message
  - power-state-change
```

nzoschke:
I do see hints about support for the merge strategies in `/usr/lib/python2.7/dist-packages/cloudinit/mergers`

```
[ec2-user@ip-10-0-2-76 mergers]$ ack append
m_str.py
23:        self._append = 'append' in opts
26:        return 'StringMerger: (append=%s)' % (self._append)
34:    # perform the following action, if appending we will
39:        if not self._append:

__init__.py
112:            parsed_mergers.append((name, opts))
133:        parsed_mergers.append((m_name, m_ops))
158:            mergers_to_be.append((mod_attr, m_ops))
163:        mergers.append(attr(root, opts))

m_list.py
20:MERGE_TYPES = ('append', 'prepend', DEF_MERGE_TYPE, 'no_replace')
65:        elif self._method == 'append':
```

nzoschke:
I do see hints of merging in the logs...

`Apr 27 00:33:01 cloud-init[2174]: cloud_config.py[DEBUG]: Merging by applying [('dict', ['recurse_array', 'no_replace']), ('list', ['append'])]`

nzoschke:
I think the cloud-init `write-files` hook happens early enough...

nzoschke:
FINALLY had some success by copying `            "  - [ cloud-init-per, instance, docker_storage_setup, /usr/bin/docker-storage-setup ]\n",` onto the end of our bootcmd list

nzoschke:
Closing this, and re-opening. Thanks again for the initial spike.

PR #556, opened 2016-04-19
Syslog

nzoschke:
With https://github.com/convox/syslog, which has been released.

## Demo

```bash
$ convox services create syslog --url tcp+tls://logs1.papertrailapp.com:11235
Creating syslog-8075 (syslog: url="tcp+tls://logs1.papertrailapp.com:11235")... CREATING

$ convox services info syslog-8075
Name    syslog-8075
Status  running
Exports
  URL: tcp+tls://logs1.papertrailapp.com:11235

$ convox services link syslog-8075 --app httpd
Linked syslog-8075 to httpd
```

screen shot 2016-04-18 at 6 53 34 pm

screen shot 2016-04-18 at 7 50 14 pm

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Edit release record in GitHub
- [ ] Publish release
- [ ] Release CLI

PR #557, opened 2016-04-19
Cron

mattmanning:
# Cron

Add cron jobs to your convox app.

## UX

Add a label in the format of `convox.cron.<name>=<cron schedule expression> <command>`.

Example:

```yaml
worker:
  labels:
    - convox.cron.myjob=* * * * ? rake background_job
```

When you deploy your app, Convox will create a Lambda Scheduled Event to run the command `rake background_job` in your `worker` process every 1 minute.

## How it works

If you have defined any cron jobs at all, convox will create a special Lambda function for you when you deploy. A zip file is uploaded to the app settings bucket at `functions/cron.zip` to back the function.

For each cron job a CloudWatchEvents Rule is set up with the given cron schedule. The Rule invokes the Lambda function with a payload that contains the app name, process name, and command.

## Open Questions

Checking for the lambda zip file every time seems kinda wrong. I tried inlining the function first, but you can only inline functions with `nodejs` as the runtime. I wanted to use `nodejs4.3` because AWS is [dropping support](https://aws.amazon.com/blogs/compute/node-js-4-3-2-runtime-now-available-on-lambda/) for creating functions with `nodejs` as the runtime in October.

Also, right now I'm creating one Lambda function per app, but I really only need one function per rack. If I add the function to the rack CF template, can I somehow get its name and arn in the app template so I can create the right rule mappings and permissions?

## Release Playbook
- [x] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Edit release record in GitHub
- [ ] Publish release
- [ ] Release CLI


api/models/cronjob.go
```diff
@@ -0,0 +1,107 @@
...
+func (cr *CronJob) ShortName() string {
+	return fmt.Sprintf("%s%s", strings.Title(cr.ManifestEntry.Name), strings.Title(cr.Name))
+}
+
+func (cr *CronJob) LongName() string {
+	app := cr.ManifestEntry.app
+	return fmt.Sprintf("%s-%s-%s", app.StackName(), cr.Process(), cr.Name)
+}
+
+func (cr *CronJob) UploadLambdaFunction() error {
```
nzoschke: This is an interesting implementation. When we manage lambda functions in other places (papertrail, syslog service, custome handler), we publish zips to our bucket and reference those URLs.

https://github.com/convox/syslog/blob/master/Makefile#L16

nzoschke: Another strategy with simpler .js source is to embed it directly in cloudformation. 

https://github.com/convox/rack/blob/master/api/dist/kernel.json#L324

mattmanning: @nzoschke Thanks for the tip. I was assuming the S3 bucket referred to [here](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-lambda-function-code.html) had to be in the same account, but maybe not. If I could just point to a zip file in a public bucket that would be fantastic.

The only unique thing in my current lambda function is the rack stack name, but I can probably configure the CloudWatch Event to send that, too.

See my comment above about inlining. As I read the docs, that's not going to be possible after October 2016.

ddollar:
Merged in #564

nzoschke:
👍 on shipping experimental features fast and getting feedback.

I have the same questions about managing the lambda source. But this strategy is good to experiment with.

Also same question about multiple lambda functions. Take a look at how the pending syslog stuff will work:

https://github.com/convox/rack/pull/556

Single lambda function, then go templating and cloudformation updates to add additional subscriptions on the lambda func. I think the same pattern could apply here.

PR #563, opened 2016-04-20
configure load balancers with labels

ddollar:
* Removes `ssl create` and `ssl delete` from the CLI and API
* Changes SSL update endpoint from `PUT /apps/:app/ssl` to `PUT /apps/:app/ssl/:process/:port`
* Adds SSL API endpoints to `manifest.yml`
* Removes the current `convox.port.###.protocol` behavior
* Port numbers in `convox.port` labels should refer to the front-end balancer port (first half of the pair)
* Adds three new labels:
  - `convox.port.###.protocol={http|https|tcp|tls}`
  - `convox.port.###.proxy={true|false}` (true if the container wants a PROXY header for remote IP information)
  - `convox.port.###.secure={true|false}` (true if the container is listening with https or tls)
* Makes `convox start` work appropriately with these labels
* If a secure protocol is chosen and no certificate exists on the application, a self-signed cert for the ELB hostname will be created and used
* Should be backwards compatible with existing SSL-enabled applications

#### Todo (non-blocking)
* Support the `secure` label in `convox start`

#### See Also
* https://github.com/convox/proxy/pull/1
* https://github.com/convox/convox.github.io/pull/125

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI


mattmanning:
I like this approach a lot. It makes sense to configure this stuff next to port configurations in `docker-compose.yml`.

Can we change the CLI command to `convox ssl upload` or something? Having just update there unchanged but removing the create and delete feels a little funny.

ddollar:
`convox ssl update` still makes sense since you can specify an ARN or key files. `upload` would only make sense for the latter.

api/manifest/manifest.go
```diff
@@ -774,11 +774,17 @@ func (me ManifestEntry) runAsync(m *Manifest, prefix, app, process string, cache
...
+			proxy := false
+
+			if me.Label(fmt.Sprintf("convox.port.%s.proxy")) == "true" {
+				proxy = true
+			}
+
 			rnd := RandomPort()
-			fmt.Println(prefix, special(fmt.Sprintf("proxy protocol enabled for %s:%s", host, container)))
-			go proxyPort(host, fmt.Sprintf("%s:%d", gateway, rnd))
+			fmt.Println(prefix, special(fmt.Sprintf("https proxy enabled for %s:%s", host, container)))
```
ddollar: Dont hardcode https

api/manifest/manifest.go
```diff
@@ -903,8 +909,17 @@ func exists(filename string) bool {
...
-	Execer("docker", "run", "-p", fmt.Sprintf("%s:%s", from, from), "convox/proxy", from, to, "proxy").Run()
+func proxyPort(protocol, from, to string, proxy bool) {
+	args := []string{"run", "-p", fmt.Sprintf("%s:%s", from, from), "convox/proxy", from, to, protocol}
+
+	if proxy {
+		args = append(args, "proxy")
+	}
+
+	cmd := Execer("docker", args...)
+	cmd.Stdout = os.Stdout
```
ddollar: Dont debug output

api/models/manifest_test.go
```diff
@@ -41,6 +41,7 @@ func Diff(t *testing.T, name, s1, s2 string) {
 
 	if len(diffs) > 0 {
 		t.Errorf("Unexpected results for %s:\n%s", name, strings.Join(diffs, "\n"))
+		fmt.Println(s2)
```
ddollar: debug output

api/models/templates/app.tmpl
```diff
@@ -252,17 +263,17 @@
...
         "AllowedValues": [ "Yes", "No" ]
       },
-      "{{ upper $balancer.ProcessName }}Port{{ .Balancer }}ProxyProtocol": {
+      "{{ upper $balancer.ProcessName }}Port{{ .Balancer }}Protocol": {
         "Type" : "String",
-        "Default" : "No",
+        "Default" : "tls",
         "Description" : "",
-        "AllowedValues": [ "Yes", "No" ]
+        "AllowedValues": [ "http", "https", "tcp", "tls", "tcp+proxy", "tls+proxy" ]
```
ddollar: remove tcp+proxy and tls+proxy

ddollar:
Merged in #564

PR #570, opened 2016-04-21
log events

nzoschke:
Simple visibility enhancement. `convox logs --filter EventSend` will show you an event log.


nzoschke:
Released in #575

PR #576, opened 2016-04-25
WIP: Improve YAML handling/validation for environment

nathany:
* Adding tests for valid and invalid map and array merges https://github.com/convox/rack/issues/481
* Forcing environment into a slice for later usage https://github.com/convox/rack/issues/112

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Edit [Rack release record](https://github.com/convox/rack/releases) and/or update [docs](https://github.com/convox/convox.github.io)
- [ ] Publish release
- [ ] Release CLI

PR #578, opened 2016-04-25
Handle environment section in map format

### LINKED ISSUE #112: Handle environment section in map format
If you have an environment section like:

```
environment: 
  FOO: bar
  BAZ: qux
```

it will cause failures with the following error:

```
invalid manifest: yaml: unmarshal errors: line 4: cannot unmarshal !!map into []string
```

nathany:
## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

fixes #112

this pulls in some code from libcompose (Apache license) but modifies it to work with gopkg.in/yaml.v2.

### DIFF
--- a/api/models/manifest.go
+++ b/api/models/manifest.go
@@ -23,7 +23,7 @@ type ManifestEntry struct {
 
 	Build      string                   `yaml:"build"`
 	Command    interface{}              `yaml:"command"`
-	Env        []string                 `yaml:"environment"`
+	Env        MapOrEqualSlice          `yaml:"environment"`
 	Exports    map[string]string        `yaml:"-"`
 	Image      string                   `yaml:"image"`
 	Labels     interface{}              `yaml:"labels"`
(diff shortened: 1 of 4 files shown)


awsmsrc:
We should consider duplicating this behavior in manifest/manifest.go as well, and possibly porting the tests from #576. As ideally we will be losing this model sooner rather then later and this will also prevent sending a bad environment in the first place.

I'm unsure how that would effect YAML map merging however.

api/models/manifest.go
```diff
@@ -532,3 +532,69 @@ func (me ManifestEntry) RegistryImage(app *App, buildId string) string {
 
 	return fmt.Sprintf("%s/%s-%s:%s", os.Getenv("REGISTRY_HOST"), app.Name, me.Name, buildId)
 }
+
+type MapOrEqualSlice []string
+
+// UnmarshalYAML implements the Unmarshaller interface.
+func (s *MapOrEqualSlice) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
nathany: this is the only interesting bit, changing the Unmarshaller interface from what libcompose does to what gopkg.in does.

ddollar:
:+1: Looks good. Thanks for adding the test, too!

mattmanning:
Merged in #584

PR #591, opened 2016-05-02
update init system and default manifest templates

ddollar:
## Release Playbook
- [x] Rebase against master
- [x] Release branch (20160513200733-init)
- [x] Pass CI (https://circleci.com/gh/convox/rack/762)
- [x] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Edit [Rack release record](https://github.com/convox/rack/releases) and/or update [docs](https://github.com/convox/convox.github.io)
- [ ] Publish release
- [ ] Release CLI

PR #628, opened 2016-05-13
Generate usable manifest for a project with Dockerfile & Procfile

prognostikos:
## Release Playbook
- [x] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

Before this change, the generated `docker-compose.yml` would not
generate sections for the services defined in the Procfile.

### DIFF
--- a/examples/dockerfile-procfile/Dockerfile
+++ b/examples/dockerfile-procfile/Dockerfile
@@ -1,0 +1,13 @@
+FROM gliderlabs/alpine:3.1
+
+RUN apk-install ca-certificates ruby ruby-dev
+
+RUN gem install sinatra
+
+ENV PORT 3000
+EXPOSE 3000
+
+WORKDIR /app
+COPY . /app
+
+CMD ["ruby", "web.rb"]
--- a/api/manifest/manifest.go
+++ b/api/manifest/manifest.go
@@ -1422,17 +1422,16 @@ func detectApplication(dir string) string {
 
 func initApplication(dir string) error {
 	wd, err := os.Getwd()
-	
+
 	if err != nil {
 		return err
 	}
-	
+
 	defer os.Chdir(wd)
-	
+
 	os.Chdir(dir)
-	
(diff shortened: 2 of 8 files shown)


prognostikos:
So this is kind of a minimal, hackish way of solving the problem I had, but it does work :)

Another option would be to re-work the switch statement in Init()

Open to suggestions - still very much a golang beginner.

prognostikos:
Hm, looking at the other open PRs it seems this will end up conflicting with #591- @ddollar wdyt?

ddollar:
Yes I think it will. #591 attempts to simplify the `convox init` process a bit. Would it be possible to rebase this work on top of that? I'd be happy to jump on a Hangout or something and assist.

prognostikos:
I'll give it a try and ping you if needed.

prognostikos:
This is ready for review.

examples/dockerfile-procfile/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM gliderlabs/alpine:3.1
```
matthewford: how about `ruby:2.3-alpine`?

prognostikos: @matthewford that's probably a good idea for another PR, and all of the other examples Dockerfiles could be updated as well. AFAIK, images are never built with these Dockerfiles - they're only used when testing various `convox init` scenarios.

ddollar:
I've had some problems with the alpine images and their non-standard libc. How about making these be based on ubuntu:16.04 like convox/rails

mattmanning:
Closing this in favor of https://github.com/convox/rack/pull/683

In that PR I have removed the use of alpine in the test and used our convox/ruby image instead.

Issue #41, opened 2015-09-18
Bad output for `convox env` on a new app


mattmanning:
```
$ convox apps create test3
Creating app test3... CREATING

$ convox env --app test3
<?xml version="1.0" encoding="UTF-8"?>
<ListBucketResult xmlns="http://s3.amazonaws.com/doc/2006-03-01/"><Name>env</Name><Prefix></Prefix><Marker></Marker><MaxKeys>1000</MaxKeys><IsTruncated>false</IsTruncated><Contents><Key>contact_us_for_a_quote</Key><LastModified>2014-03-03T15:07:22.000Z</LastModified><ETag>&quot;dc3796df991d4fa2b1ba3303dc8d4ec8&quot;</ETag><Size>17</Size><StorageClass>STANDARD</StorageClass></Contents><Contents><Key>index.html</Key><LastModified>2014-02-28T13:01:18.000Z</LastModified><ETag>&quot;d41d8cd98f00b204e9800998ecf8427e&quot;</ETag><Size>0</Size><StorageClass>STANDARD</StorageClass></Contents><Contents><Key>your_objects_here</Key><LastModified>2014-03-03T15:07:22.000Z</LastModified><ETag>&quot;d41d8cd98f00b204e9800998ecf8427e&quot;</ETag><Size>0</Size><StorageClass>STANDARD</StorageClass></Contents></ListBucketResult>
```

mattmanning:
This only happens when the app is the "creating" state.

joonathan:
Also, while app is still in creating state following could happen:

Updating environment... ERROR: AccessDenied: Access Denied
	status code: 403, request id:

nzoschke:
Confirmed that the XML output is still happening.

nzoschke:
Fixed on #631. Thanks @MiguelMoll !

Issue #601, opened 2016-05-05
`convox run` could have nicer output


beedub:
```
$ convox run RLVSHBOYZML bash
ERROR: InvalidParameterException: Task Definition can not be blank.
	status code: 400, request id: 769efe3c-1319-11e6-b48e-dd0e977921b0
```

I accidentally used the release instead of service name. It seems like convox should be checking if the service name exists before passing it through to ECS. In any case, it seems like the error message could be friendlier.

nzoschke:
A friendlier error message landed on #631. This was a CLI patch. Someday we can also fix this in the API.

PR #630, opened 2016-05-13
Validate the process ID passed to run

MiguelMoll:
When executing a `convox run` command, make sure the process ID is a
valid one.

I could see this exact logic implemented in `client/process.go` but this being my original method, I went with it.

Should solve #601
## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

### DIFF
--- a/cmd/convox/run.go
+++ b/cmd/convox/run.go
@@ -51,6 +51,11 @@ func cmdRun(c *cli.Context) {
 	}
 
 	ps := c.Args()[0]
+	err = validateProcessId(c, app, ps)
+	if err != nil {
+		stdcli.Error(err)
+		return
+	}
 
 	args := strings.Join(c.Args()[1:], " ")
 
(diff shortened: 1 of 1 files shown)


nzoschke:
I believe this fixes https://github.com/convox/rack/issues/41.

#335 will require some checking around the app name and the rack name

nzoschke:
This looks like it will improve the output for #601.

One side effect to be aware of is that the list processes API can take a while for very large apps. So now `convox run` could be much slower in this case. 

Ideally this validation would be handled on the API side to avoid excess process listing.

nzoschke:
merged on #631 631

PR #632, opened 2016-05-14
Validate syslog url scheme

### LINKED ISSUE #581: new syslog service url scheme is potentially confusing
I added a new syslog service - 

```convox services create syslog --url tcp+ssl://logs1.papertrailapp.com:12345```

It was created successfully but logs never made it to papertrail.

Turns out the url was incorrectly specified `tcp+ssl` vs. `tcp+tls`.

Ideally this would have validated the url was correct.
I have no idea if `tcp+ssl` is even a valid protocol definition and have never seen `tcp+tls` specified like this in a url.

### PR DESCRIPTION
**Pull request by prognostikos**
## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [x] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

Example:

    convox services create syslog --url tcp+ssl://example.com:11234
    ERROR: Invalid url scheme `tcp+ssl`. Allowed schemes are `udp`, `tcp`, `tcp+tls`.

Fixes #581

### DIFF
--- a/api/provider/aws/service.go
+++ b/api/provider/aws/service.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
 
@@ -315,6 +316,18 @@ func (p *AWSProvider) ServiceUnlinkSubscribe(a *structs.App, s *structs.Service)
 }
 
 func createSyslog(s *structs.Service) (*cloudformation.CreateStackInput, error) {
+	u, err := url.Parse(s.Parameters["Url"])
+	if err != nil {
+		return nil, err
+	}
+
+	switch u.Scheme {
+	case "udp", "tcp", "tcp+tls":
+		// proceed
+	default:
+		return nil, fmt.Errorf("Invalid url scheme `%s`. Allowed schemes are `udp`, `tcp`, `tcp+tls`.", u.Scheme)
+	}
+
 	formation, err := serviceFormation(s.Type, nil)
 	if err != nil {
 		return nil, err

### CODE REVIEW

**Comment by prognostikos** [comment]
As a convox & golang beginner, comments are welcome :)

Not sure if the cmd is the best place for validation.

Moving the validation into its own function would be cleaner.

I couldn't find tests for `cmd/convox/services` so not sure where they would go, but would like to add one for this.

As I said, comments/feedback welcome.

**Comment by prognostikos** [comment]
Cargo-culted some tests from the others in the cmd package.

One is skipped for now as the service names are randomly generated so I'm not sure how to mock the response. The ugly way would be to return a fixed value in the production code when run in the test env - there's probably a better way :)

**Comment by prognostikos** [comment]
Figured out that I can just pass the name in on create, so this is ready for review.

**Comment by nzoschke** [comment]
Thanks! This is a very thorough CLI patch.

However ideally validation takes place in the API / controllers. Then the CLI, web UI and anything using the client can all benefit from it.

Would you be interested in helping move the validation to that layer?

**Comment by prognostikos** [comment]
@nzoschke of course, that makes sense. I'll take a look at moving it to the API layer.

**Comment by prognostikos** [comment]
@nzoschke I've moved the code. Would like to add a test for it as well.

**Comment by nzoschke** [comment]
Thanks! I'll test this manually right now and if it's working I'll get it in a release.

The testing strategy is a bit in flux.

Up until recently we have been using httptest to do a mock API server. 

But with the provider interface, we have a easier path through the TestProviderRunner and github.com/stretchr/testify/mock. However I don't see a good example of that in this code base yet. I'm happy to help get one of these types of tests going.

**Comment by nzoschke** [comment]
```
$ convox services create syslog --url foo://
Creating syslog-5559 (syslog: url="foo://")... ERROR: Invalid url scheme `foo`. Allowed schemes are `udp`, `tcp`, `tcp+tls`.

$ convox services create syslog --url tcp+tls://logs4.papertrailapp.com:29446
Creating syslog-1119 (syslog: url="tcp+tls://logs4.papertrailapp.com:29446")... CREATING
```

👍

nzoschke:
Going to close this and test and release on #647. Would totally love a patch that tests this too but we can follow up on that any time @prognostikos.

PR #629, opened 2016-05-13
Allow for any number of whitespace characters between the COPY/ADD commands for file syncing

dbeard:
Was banging my head for a little bit why the file syncing wasn't working, and realized that the Dockerfile had an extra space between the local and remote files specified in the COPY command (e.g. COPY ./app  /app #Notice two spaces between them). Technically it's a valid Dockerfile, and thought this may help someone else who runs into a similar issue for some reason.

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [x] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Edit [Rack release record](https://github.com/convox/rack/releases) and/or update [docs](https://github.com/convox/convox.github.io)
- [ ] Publish release
- [ ] Release CLI

PR #635, opened 2016-05-15
Rails init

mattmanning:
This PR contains some changes to `convox init` to make getting started with Rails a little smoother.

* The generated `docker-compose.yml` not includes a `.:/app` volume mount
* `bin/web` is generated and removes `tmp/pids/server.pid` before starting Rails to avoid errors
*  `.dockerignore` and `config/initializers/convox.rb` are generated

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI


cmd/convox/templates/init/rails/config/initializers/convox.rb
```diff
@@ -0,0 +1,5 @@
+STDOUT.sync = true
```
ddollar: This file already comes from the convox/rails image

cmd/convox/templates/init/rails/bin/web
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
ddollar: This file comes from the convox/rails image. You should add your server.pid line to there.

cmd/convox/templates/init/rails/docker-compose.yml
```diff
@@ -6,3 +6,5 @@ web:
   ports:
     - 80:4000
     - 443:4001
+  volumes:
```
ddollar: We shouldn't do volumes like this by default until Docker for Mac is more widely available.

api/manifest/manifest.go
```diff
@@ -1448,6 +1448,20 @@ func initApplication(dir string) error {
 		return err
 	}
 
+	if kind == "rails" {
```
ddollar: If we have language-specific things they should be handled in the language Dockerfiles. bin/web and initializers/convox.rb come from convox/rails. .dockerignore should probably be global to all languages.

cmd/convox/templates/init/rails/config/initializers/convox.rb
mattmanning: When you volume mount the app directory the file from the image becomes unavailable because it isn't present on the host.

cmd/convox/templates/init/rails/bin/web
mattmanning: When you volume mount the app directory the file from the image becomes unavailable because it isn't present on the host.

I will also send a PR to the Docker image repo for this.

cmd/convox/templates/init/rails/docker-compose.yml
mattmanning: +1 for changing defaults on sync when we make this change

Happy to continue discussion for timing on this. Local Rais development without 2-way sync doesn't work very well. Looking forward to wider Docker for Mac availability here.

api/manifest/manifest.go
mattmanning: See comments above about files from images becoming unavailable with volume mounts.

This should totally be a switch. I'll make that change and refactor this a bit, but I think we still need to generate these files if volume mounting is our sync solution going forward.

cmd/convox/templates/init/rails/docker-compose.yml
nzoschke: I think we should confidently switch back to volumes ASAP. 

Docker for Mac is documented and downloadable as if its in GA: https://beta.docker.com/docs/mac/getting-started/

I'm testing how fast beta keys get delivered now. Rumor is it should be fast, though it's not yet totally automatic, i've been waiting for 30m now.

Volumes also work on every other way people are running Docker.

The convox sync stuff is very experimental, and people are having real problems with it out in the wild. https://github.com/convox/rack/pull/629

So I don't see any harm in this landing ASAP. Skate to where the puck is headed.

ddollar: Has anyone done any performance testing (even just anecdotal) with volumes on Docker for Mac? The existing solutions with volumes had real problems around performance that made them basically unusable. Are those problems gone?

ddollar: Volumes are not necessarily where the puck is headed. I still have real questions around dev/prod parity with volumes many of which are apparent by the files "lost" by mounting over them with volumes. Those files would suddenly reappear in production.

nzoschke: This `osxfs` stuff serious engineering and obviously reflects where Docker is putting all their focus: https://beta.docker.com/docs/mac/osxfs/

The docs are a lot more clear about the goals and hint at performance:

> osxfs provides a close-to-native user experience for bind mounting OS X file system trees into Docker containers.

But I do see your point.

* Bind mounts have side effects even if the mac/VM story was flawless.
* Volumes were simply unusable on the last iteration of these tools, and nobody was admitting it
* The osxfs stuff needs to be benchmarked

nzoschke: Can you offer a design sketch on how the archive and exec APIs can be used to do two-way sync for real world apps and development workflows?

nzoschke: Seems like you need to do inotify both on the host and inside the container, which is what osxfs is doing.

nzoschke: Maybe you can simplify things by copying files back out after a docker run or ctrl-c.

nzoschke: All said right now I'm still in favor of volume mounts and hving Convox init inject files into the project. 

Having important stuff already in /app/ in the base image causes so many side effects. Volume mounts Dont worj, and I don't have a good way to copy or customize the bin/web script or rails config file if it doesn't fit my app.

Convox inits job is to look at a code base, and make some guesses about what additional files it needs to run. In the real world every one of these guesses will need to be tweaked. Hard coding these into base images will make that hard.

ddollar: Docker exec a watcher that lists files that change. Docker cp them out when they show up in the list. The tech there should be trivial.

ddollar: Also just a note that just because Docker is investing energy into it doesn't mean it's the right call. See overlay networks.

ddollar: There are a plethora of other things that might mutate /app and get blown away. Assets precompile?

ddollar: They aren't hardcoded into images as they can be overwritten with a future COPY. They are just good defaults that the large majority of users should not need to tweak. For those that do we can provide a method to go power mode.

nzoschke: Ah good call on the COPY.

There's no reason why either strategy can't work. I think it's a matter of making the decision in a way that's actionable for the rails example to improve and for other language frameworks to get moving.

mattmanning: I've just done some overdue testing with Docker for Mac on a pretty heavy-duty Rails app and sadly found the volume performance to be terrible.

There's a pretty interesting thread on the topic in the Docker forums here: https://forums.docker.com/t/file-access-in-mounted-volumes-extremely-slow-cpu-bound/8076/15 

Seems like the osxfs performance is not really up to snuff yet. There are some alternatives presented in there and also some benchmark data.

The last word from Docker was 11days ago:

> Just an update here from the Docker camp...
We're still working on it! We've added some small performance boosts over the last few Beta's and performance is heading in the right direction. There are a few other fixes that should be heading in over the next few betas that will also provide a decent boost. The target is to be much faster than VirtualBox and we're confident we can do this before release.
>
> In the meantime the best thing you can do to help is to give us samples of workloads you feel are slow ideally in a a form that we can replicate e.g Dockerfiles, Compose Files etc... We can then analyse these and try to optimise as much as possible for each!
>
> Thanks for bearing with us and for trying the Docker for Mac beta

mattmanning:
Closing this for now since there's still so much uncertainty around Docker volumes.

I do think `convox init` should generate a `.dockerignore`, though, so I'll open a separate PR for that.

PR #652, opened 2016-05-18
Require at least 2 instances

kmonkeyjam:
Not sure if this is the right/useful fix or not.  Based on my conversation with @mwarkentin, there should be at least 1 more instance than the number of processes your app is scaled to.  Since by default when you create your app, it will be scaled to at least 1, it doesn't make sense to me that you can create a rack with < 2 instances.  If you do, you'll end up in a weird state where 2 containers end up launching on the same instance and you have to hand stop one of the tasks.  There's a more fundamental ecs weirdness that you don't get an error when updating a task when there is only one instance, and this same failure happens.

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

### DIFF
--- a/cmd/convox/install.go
+++ b/cmd/convox/install.go
@@ -81,7 +81,7 @@ func init() {
 			cli.IntFlag{
 				Name:  "instance-count",
 				Value: 3,
-				Usage: "number of EC2 instances",
+				Usage: "number of EC2 instances (must be greater than 1)",
 			},
 			cli.StringFlag{
 				Name:  "instance-type",
@@ -313,7 +313,11 @@ func cmdInstall(c *cli.Context) {
 	versionName := version.Version
 	formationUrl := fmt.Sprintf(FormationUrl, versionName)
 
-	instanceCount := fmt.Sprintf("%d", c.Int("instance-count"))
+	numInstances := c.Int("instance-count")
+	instanceCount := fmt.Sprintf("%d", numInstances)
+        if numInstances < 2 {
+		stdcli.QOSEventSend("cli-install", distinctId, stdcli.QOSEventProperties{Error: fmt.Errorf("instance-count must be greater than 1")})
+        }
 
 	fmt.Printf("Installing Convox (%s)...\n", versionName)
 


nzoschke:
This will be a great enhancement to the installer, thank you!

Yeah requiring a minimum of two instances seems very desirable for now. We could make other enhancements to make a deploy work with a single instance with some downtime, but the focusing on a zero-downtime architecture is best for now.

I think this should use `stdcli.Error(fmt.Errorf("instance-count must be greater than 1"))` to exit.

The QOS is new, so apologies for the lack of clarity on how to use it, but its mostly for getting unhandled errors into rollbar. In this case your patch cleanly validates a bad input, so it's not an exception.

There should be a corresponding validation in the API where you can scale a rack, but that could be handled in another issue / PR. https://github.com/convox/rack/blob/master/api/controllers/system.go#L43

nzoschke:
@kmonkeyjam I'll run with this and tweak the cli error and add the API validation too. Thank you so much for your testing, feedback, and patch!

nzoschke:
Closing this and opening a new PR for test / merge / release.

jordansexton:
> there should be at least 1 more instance than the number of processes your app is scaled to

Can anyone explain why this is? Requiring 2 instances for a tiny app seems heavy handed, and I haven't had any issues with needing to stop tasks when deploying.

mwarkentin:
Right now, at least with default settings, I don't think deploys will work if you don't have the extra instance (as the minimum healthy containers is 100%) - they'll basically just sit there waiting for another instance to place the container. Now that those params are configurable, it might not be an issue (though it still would be with the default settings).

jordansexton:
This is also a problem:
```
convox rack scale --count 0
ERROR: count must be greater than 1
```

There's no way to just scale a rack off now (also useful for staging environments). In short, this minimum of 2 only applies to production deployments.

jordansexton:
The params of the CFT are also validated, preventing it from being scaled down without the CLI.

screen shot 2016-06-03 at 4 44 41 pm

nzoschke:
@jordansexton really great feedback about staging environments, thanks.

There will always be a strong bias in Convox towards production environments. The goal is to deliver best practices out of the box: no single point of failure, multi-az placement, zero downtime deploys and fully automated deploys.

There should also be decreasingly less reasons to not follow the best practices in staging. 

What is your staging app footprint and price point? Would 2 t2.nanos make sense? They will cost < $9 / mo. on demand.

Turning off a staging or dev rack is a good technique. Another way to do this is to set up [Scheduled Scaling](http://docs.aws.amazon.com/autoscaling/latest/userguide/schedule_time.html). Manual modifications to the ASG will be honored independently of the rack InstanceCount parameter.

jordansexton:
@nzoschke Thanks for the detailed response, and that's a good point about scheduled scaling. It's something that could probably work for us if we tweak our workflow a bit.

However, would it be a problem to allow overriding the minimum if desired? I definitely understand the focus on production environments, but why not default to >= 2 instances, but still allow < 2 and just insert a descriptive (and perhaps scary) warning?

I suppose the default of 2 means that existing deployments (like our staging env) go up to 2 instances when upgraded, but it would be nice to be able to just scale the rack down to 1 or 0 rather than rely on scheduling. So far we've been able to use convox directly without even worrying about the AWS CLI or thinking about CFTs lately and it's been pretty great 😀

Anyway, I admit my reaction was in part due to surprise about the change, so please feel free to close this issue if the current behavior seems like the balance you want to strike.

nzoschke:
The new default does not automatically change your rack. You will get an error when you try to update:

> $ convox rack update
> ERROR: ValidationError: Parameter 'InstanceCount' must be a number not less than 2
>	status code: 400, request id: eefc5574-2cf6-11e6-abf1-0179c4f546cf

As for allowing 1 instance with a big warning, I'm still concerned about other side effects. 

One side effect I see is that we schedule 2 rack API containers for high availability of the control plane. Without conditionally scaling this down to 1 also, ECS will constantly be churning, and never detect and send notifications when any deployment is complete.

https://github.com/convox/rack/blob/master/api/dist/kernel.json#L1305

So I'd like to keep the default at 2 until someone presents a more comprehensive patch that addresses control plane, deployment and notifications challenges with 1 instance.

PR #648, opened 2016-05-17
use links to bind proxy container

ddollar:
This allows the proxy container to connect to the container it's proxying for.

See also: convox/proxy#3

## Quick CLI Release Playbook
- [ ] Code review
- [ ] Merge into master
- [ ] Release CLI

PR #661, opened 2016-05-19
WIP: Allow multiple apps in convox start with --link

ddollar:
This PR is rebased on top of #648

```
$ cd foo && convox start
web: starting

$ convox start --name bar --link foo
worker: starting (with $FOO_WEB_HOST)

$ convox start --link foo --link bar
monitor: starting (with $FOO_WEB_HOST and $BAR_WORKER_HOST)
```

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

### DIFF
--- a/api/manifest/manifest.go
+++ b/api/manifest/manifest.go
@@ -915,6 +915,9 @@ func proxyPort(protocol, from, to, link string, proxy bool) {
 		args = append(args, "proxy")
 	}
 
+	// wait for main container to come up
+	time.Sleep(1 * time.Second)
+
 	cmd := Execer("docker", args...)
 	// cmd.Stdout = os.Stdout
 	// cmd.Stderr = os.Stderr
--- a/api/manifest/manifest.go
+++ b/api/manifest/manifest.go
@@ -564,8 +564,14 @@ func (m *Manifest) Run(app string, cache bool) []error {
 
 	for i, name := range m.runOrder() {
 		sch := make(chan error)
-		go (*m)[name].runAsync(m, m.prefixForEntry(name, i), app, name, cache, ch, sch)
-		<-sch // block until started successfully
+		go (*m)[name].runAsync(m, m.prefixForEntry(name, i), app, name, cache, links, ch, sch)
+		select {
+		case <-sch:
+			continue
+		case err := <-ch:
+			fmt.Fprintf(os.Stderr, "ERROR: %s\n", err)
+			return []error{err}
+		}
 	}
 
 	errors := []error{}
--- a/api/manifest/manifest.go
+++ b/api/manifest/manifest.go
@@ -505,7 +505,7 @@ func (m *Manifest) Raw() ([]byte, error) {
 	return yaml.Marshal(m)
 }
 
-func (m *Manifest) Run(app string, cache bool) []error {
+func (m *Manifest) Run(app string, cache bool, links []string) []error {
 	ch := make(chan error)
 	sigch := make(chan os.Signal, 1)
 	signal.Notify(sigch, os.Interrupt, os.Kill)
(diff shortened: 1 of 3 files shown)


mwarkentin:
Took a quick read through - here are a few questions I had (maybe some more examples of how you'd actually interact with this would help):

* How do I make a request from my host too `foo` or `bar` app? Are those environment variables in the logs set on the host? What do their values look like?
* The actual linking looks like it would have to be done manually, and possibly in a specific order - I'm not sure how this would react to trying to link an app that's not currently running. My concern here is that devs work on subsets of our apps at once

### Dev 1

Working on: `a1`, `a2`, `a3`
Runs (?):

```
# shell1
cd a1 && convox start

# shell2
cd a2 && convox start --link a1

# shell3
cd a3 && convox start --link a1 --link a2
```

### Dev 2

Working on: `a1`, `a4`, `a5`, `a6`
Runs (?):

```
# shell1
cd a1 && convox start

# shell2
cd a4 && convox start --link a1

# shell3
cd a5 && convox start --link a1 --link a4

# shell4
cd a6 && convox start --link a1 --link a4 --link a5
```

This seems pretty complicated, manual, and error-prone - but I could be misunderstanding how it's meant to be used.

ddollar:
Yes, you have the basics correct. Working on some of the apps some of the time is a complicated proposition. My suggestion would be to build a shell script that run the `convox start`s for you. Perhaps in different tmux windows or terminal tabs?

ddollar:
Something like this:

```shell
#!/bin/bash

if [ -z $@ ]; then
  processes="all"
else
  processes=$@
fi

[ -z "$processes" ] && processes="all"

requested() {
  [[ "$processes" == "all" ]] && return 0
  for e in $processes; do [[ "$e" == "$1" ]] && return 0; done
  return 1
}

start() {
  app=$1
  shift
  args=$@

  requested $app && echo "starting $app: $args"
}

start bar
start baz
start foo --link bar --link baz
```

ddollar:
Only one application is accessible on the host. This is a limitation of Docker for Mac.

Any application started with `--link` does not expose its ports to the host. They are only available internally to other applications to access by `--link`ing to them.

mattmanning:
This is really cool. I worry that we're overloading the word "link" a little too much, but I can't think of a great alternative . . .

ddollar:
Are there fairly standard setups when working on different things? Perhaps you can create shell scripts to encode these:

```shell
$ cat start-mode1.sh
#!/bin/bash
term-new-window() { ... }
term-new-tab() { ... }

term-new-window
term-new-tab "cd ~/src/a1 && convox start --internal"
term-new-tab "cd ~/src/a2 && convox start --internal --link a1"
term-new-tab "cd ~/src/a3 && convox start --internal --link a1 --link a2"
```

mwarkentin:
Yeah, I don't really think that this is a workable solution for us - many of our apps have both web processes (need to be reachable from host) as well as rpc processes for inter-app communication (communicating between docker containers should work great).

With that said, I don't have any specific suggestions at this point - the best I've got right now is being able to specify a unique local port for dev (which gives us multi-app access from the host, fixes port conflicts) and specify a different port to be exposed on the ELB (which we want to be 443 for all of our apps so we can access over HTTPS without specifying a port in our dev/stage/production racks).

Perhaps that looks something like this in `docker-compose.yml`:

```yaml
web:
  build: .
  command: actions/web.sh
  environment:
    - DEVELOPMENT_MODE=true
    - WEB_CONCURRENCY=2
  labels:
    - convox.elb.port=443  # Replace 9001 with 443 on the ELB
    - convox.port.443.protocol=https
  ports:
    - 9001:80  # Unique port for local dev
  links:
    - redis
redis:
  image: convox/redis
  ports:
    - 9301
```

Looking at this some more, this actually looks like it could be a good solution for our apps rpc communication for local dev (which we haven't needed yet in the first two apps). Need to think on that some more - might not be an issue because we could use unique ports for those already (and I think communicate over the docker network with those ports), since we don't have the `443` requirement that we have for web traffic..

ddollar:
What about something like:

```
$ convox start
$ curl https://localhost:443/

$ convox start --port-shift 10000`
$ curl https://localhost:10443/
```

mwarkentin:
That could work... especially if we could encode that in our `docker-compose.yml` via label or similar...

ddollar:
Probably a `.convox/shift` file

ddollar:
Took a swing at that in #662

Issue #484, opened 2016-03-28
`convox start` host port conflicts


mattmanning:
One of our users has a `docker-compose.yml` that looks something like:

```
foo:
  ports:
    - 80:3000
bar:
  ports:
    - 80:3000
```

This works fine in production because foo and bar will each have their own ELB, but when running locally with `convox start` there is a conflict on host port 80.

How can we change our tooling to improve the local development experience in this case?

nzoschke:
I believe in the newest version of Docker Engine and Docker Compose you can use networks to get around this. You can make two networks for this app, and assign each process to an isolated network and access them with a different name/ip but same port.

nzoschke:
Here are some possible paths I see here.

We strongly support and encourage this in production. We spend time to make the development environment support the pattern.

We strongly don't support it at all. We add validation to `convox start` and `convox deploy` to not build the app in this case.

We take a middle ground. It's possible in production but not simple in development. You can use 2 docker-compose files, one for dev and one for deploy if you want this.

--

How does this work with docker-compose? Can we copy it's implementation? Docker networking should be flexible enough to make this work locally.

nzoschke:
Copying @mattmanning's additional information in #573:

I have a `docker-compose.yml` like this:

```yaml
web:
  links:
    - db1
    - db2
db1:
  ports:
    - 3306
db2:
  ports:
    - 3306
```

The DB image I'm using wants to listen on port 3306 (MySQL). This configuration should be fine in production because each will get its own load balancer. However, locally with `convox start` both containers are started with `3306:3306` as the port mapping, causing a conflict on host port 3306.

Since I'm using linking I don't really care what the host port is. I think this problem could be solved by choosing a random port for it in both cases.

mwarkentin:
Once you start needing to run multiple apps, you can run into the same issues locally:

`app1/docker-compose.yml`:

```yaml
web:
  - ports: 443:80
```

`app2/docker-compose.yml`L

```yaml
web:
  - ports: 443:80
```

You'll get a host port conflict on `443` if you try running both apps at once w/ `convox start`.

mwarkentin:
Here's a bit more information on our current set up that we're migrating to Convox, and what we're trying to achieve for local dev and in production.

### Local dev (current):

* We have ~15 apps with web processes configured in Vagrant. Each of these is configured to run on a unique port from `8000-8015`.
[...]

PR #662, opened 2016-05-19
Add --shift argument to convox start

ddollar:
Adds a `--shift` argument to `convox start` that allows the requested ports to be shifted. This allows multiple applications to run on one host without conflicts.

Another potential solution to #484
Rebased on top of #648 

```shell
$ convox start
$ curl https://localhost:443/

$ convox start --shift 10000
$ curl https://localhost:10443/

$ echo 20000 > .convox/shift
$ convox start
$ curl https://localhost:20443/
```

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

### DIFF
--- a/api/manifest/manifest.go
+++ b/api/manifest/manifest.go
@@ -915,6 +915,9 @@ func proxyPort(protocol, from, to, link string, proxy bool) {
 		args = append(args, "proxy")
 	}
 
+	// wait for main container to come up
+	time.Sleep(1 * time.Second)
+
 	cmd := Execer("docker", args...)
 	// cmd.Stdout = os.Stdout
 	// cmd.Stderr = os.Stderr
--- a/api/manifest/manifest.go
+++ b/api/manifest/manifest.go
@@ -734,13 +734,6 @@ func (me ManifestEntry) runAsync(m *Manifest, prefix, app, process string, cache
 		}
 	}
 
-	gateway, err := getDockerGateway()
-
-	if err != nil {
-		ch <- err
-		return
-	}
-
 	host := ""
 	container := ""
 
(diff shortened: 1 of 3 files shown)


mwarkentin:
I'll try this out tomorrow, thanks!

ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;KMieO63n&#x2F;174-port-conflicts"> Port Conflicts</a>

mwarkentin:
Something to think about - will this apply to all processes defined in `docker-compose.yml`? External facing only?

Perhaps `.shift` could be a set of key/value pairs so you have control over which processes this is applied to:

```
$ echo "web=20000" > .convox/shift
```

What's the thinking behind using the `.convox/shift` convention vs. something in `docker-compose.yml`?

ddollar:
For now I'd like to keep it simple and just shift everything, should make it easy to know what's being referenced at any point.

I'm still thinking about whether or not to add this to `docker-compose.yml`. It seems like it might be a per-developer-machine setting. Do you think everyone will have the exact same layout?

mwarkentin:
I'm pretty sure (at least in our case) that we'd like it to be consistent across developer environments (eg. our identity service is running on the same port on all dev machines). If it was done in `.convox/shift` we'd probably end up checking that into source control.

It seems cleaner to me to have it checked in to `docker-compose.yml`, but that could be dependent on the use case (in our case it would make sense).

ddollar:
After thinking about it changed this to read a label named `convox.start.shift=NNNNN`

ddollar:
Merging in #668

PR #715, opened 2016-06-02
change ssh port sg to use vpc cidr, not 0.0.0.0

beedub:
## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Edit [Rack release record](https://github.com/convox/rack/releases) and/or update [docs](https://github.com/convox/convox.github.io)
- [ ] Publish release
- [ ] Release CLI

I was hoping to test installing from scratch but currently blocked by https://github.com/convox/rack/issues/714. I tested manually post-install on private and public installs by manually updating the sg's. Related issue: https://github.com/convox/rack/issues/710

Issue #714, opened 2016-06-02
TEMPLATE_FILE too large for install


beedub:
I was following the development instructions and tried to specify `TEMPLATE_FILE` and received an error: `'templateBody' failed to satisfy constraint: Member must have length less than or equal to 51200`. Found a similar issue https://github.com/mapbox/cfn-config/issues/16

nzoschke:
The template file is 56561 bytes currently. 

We can upload it to S3 before installing though there is a bit of a challenge in picking a bucket to PUT it to. If you use the CloudFormation management console it creates a bucket like `cf-templates-7k121xl5kax6-us-east-1` in your account. 

The file minifies to 35k. Maybe that's a good enough hack?

PR #713, opened 2016-06-01
20160527

nzoschke:
- #703 CPU Percentage
- #705 Enable ECR for Ireland (eu-west-1)
- #707 Scale to -1 to Disable ELB and Service

## Multi-PR Release Playbook
- [x] Create a pre-release branch
- [x] Merge other PRs to include in the release
- [x] Release branch (20160601201943-20160527)
- [x] Pass CI (https://circleci.com/gh/convox/rack/792)
- [ ] Code review
- [x] Create release notes and docs PRs
- [ ] Merge into master
- [ ] Release master
- [ ] Pass CI
- [ ] Update dev and testing racks
- [ ] If rack or app formation changes, test upgrade → re-deploy existing app → downgrade cycle
- [ ] Create [Rack release record](https://github.com/convox/rack/releases) and/or update [docs](https://github.com/convox/convox.github.io)
- [ ] Publish release
- [ ] Release CLI

PR #716, opened 2016-06-02
change ssh port sg to use vpc cidr, not 0.0.0.0

### LINKED ISSUE #710: convox rack security group ssh
I noticed that the sg created for the rack instances allows ssh for `0.0.0.0/22`. It looks like `convox instance ssh` actually creates a websocket that streams through the api, so could the allowable range be more specific to an internal subnet (and thus more secure)? I tested this out on my rack and I was still able to use `convox instances ssh` into boxes

nzoschke:
Fixes #710, continuing from #715. From @beedub:

> I was hoping to test installing from scratch but currently blocked by #714. I tested manually post-install on private and public installs by manually updating the sg's. Related issue: #710

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

this should be safer

### DIFF
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -780,7 +780,7 @@
       "Properties": {
         "GroupDescription": "Instances",
         "SecurityGroupIngress": [
-          { "IpProtocol": "tcp", "FromPort": "22", "ToPort": "22", "CidrIp": "0.0.0.0/0" },
+          { "IpProtocol": "tcp", "FromPort": "22", "ToPort": "22", "CidrIp": { "Ref": "VPCCIDR" } },
           { "IpProtocol": "tcp", "FromPort": "0", "ToPort": "65535", "CidrIp": { "Ref": "VPCCIDR" } },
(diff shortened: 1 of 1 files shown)


beedub:
To add more context... it looks like `convox instances ssh` is proxying through the elb and the convox api, so there shouldn't be a need to make the instances internet-accessible in a normal, non-debug setup. Would love review and confirmation from anyone more knowledgable about convox and aws that there are no negative consequences from this change

nzoschke:
You can test applying this to a real rack with `convox rack update 20160602001833-ssh_sg`

nzoschke:
Convox should offer a VPC as secure as possible, so this change is definitely in line with that goal.

It is possible to specify a custom keypair and use that to SSH directly into the box. But in practice nobody is (or should be doing this) over `convox ssh`. Perhaps there should be docs somewhere that document this...

Anyway +1 on the patch. Going to test/release it on #713

PR #725, opened 2016-06-03
20160603

nzoschke:
- #716 change ssh port sg to use vpc cidr, not 0.0.0.0
- #704 Up ulimit for dockerd
- #719 Syslog bugfixes and Papertrail deprecation
- #721 Don't Reference ELB Hostname if Process and ELB is disabled
- #726 validate process name for `convox run` based on formation
- #727 compress kernel.json before using as TemplateBody
- #728 dont upgrade all packages; fixes a surprise update to docker 1.11

## Multi-PR Release Playbook
- [x] Create a pre-release branch
- [x] Merge other PRs to include in the release
- [x] Release branch (20160606200320-20160603)
- [x] Pass CI (https://circleci.com/gh/convox/rack/809)
- [x] Code review
- [ ] Create release notes and docs PRs
- [ ] Merge into master
- [ ] Release master
- [ ] Pass CI
- [ ] Update dev and testing racks
- [ ] If rack or app formation changes, test upgrade → re-deploy existing app → downgrade cycle
- [ ] Create [Rack release record](https://github.com/convox/rack/releases) and/or update [docs](https://github.com/convox/convox.github.io)
- [ ] Publish release
- [ ] Release CLI

PR #726, opened 2016-06-03
validate process name for `convox run` based on formation, not active…

nzoschke:
This retains the ability to do a `convox run` on a process type scaled to 0 or -1


mwarkentin:
:star:

nzoschke:
Testing / releasing on #725

PR #736, opened 2016-06-07
changed provisioned throughput for DynamoDB indexes

frsechet:
## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

A default throughput of 5 read / 5 writes for each index should be really more than enough for most users.

Objects saved in DDB are really not so big, and most users will not make many builds and releases per second. It shouldn't need to be 80 for the total (about $50/month), 5 units each max (it's in operations / second, depending on the size of the objects retrieved) should largely suffice, and those who are consistantly using more than that will probably be able to figure out that they are really doing so many builds and releases.

FYI, I have set my company's at 3 units each, and it is so low I can barely see that data in the cloudwatch alarms that will usually show "insufficient data" most of the time at about 50 builds/day, even when I tried hard to build and release a bunch of apps at about the same time.

### DIFF
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -1387,9 +1387,9 @@
           "IndexName": "app.created",
           "KeySchema": [ { "AttributeName": "app", "KeyType": "HASH" }, { "AttributeName": "created", "KeyType": "RANGE" } ],
           "Projection": { "ProjectionType": "ALL" },
-          "ProvisionedThroughput": { "ReadCapacityUnits": "20", "WriteCapacityUnits": "20" }
+          "ProvisionedThroughput": { "ReadCapacityUnits": "5", "WriteCapacityUnits": "5" }
         }],
-        "ProvisionedThroughput": { "ReadCapacityUnits": "20", "WriteCapacityUnits": "20" }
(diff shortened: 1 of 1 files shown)


mwarkentin:
How much will the monthly cost be after this change (estimate)?

frsechet:
Depends on the region of course, but roughly 4 times less :-)
About $10-12.
Then again, you can go in and edit it yourself in dynamodb, I wouldn't expect any issue even at 1 read / 1 write per index unless you are heavily using the platform (hundreds of builds per hour). For 99.999% of users, they are really overpaying with the current settings.

mwarkentin:
Neat. That is per rack, correct?

frsechet:
yes :-)

nzoschke:
Looks good to me. 20 does seem over-provisioned in hind sight

MiguelMoll:
Thanks! I'll be merging this into #739 for a release.

PR #738, opened 2016-06-07
Parameterize autoscale instance batch size

beedub:
## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

I had a fleet of 20 servers, but a rolling deploy 1-at-a-time took 30+ minutes. I would have been ok with batches of 5 servers being taken out of rotation.

Could there be weird behavior if a large batch size takes too many ECS instances out-of-rotation? Will ECS prevent the autoscale from succeeding? Despite these questions, it would be user-configurable with the safest option (1) as the default

### DIFF
--- a/api/dist/kernel.json
+++ b/api/dist/kernel.json
@@ -221,6 +221,12 @@
       "Description": "The type of the instances in the runtime cluster",
       "Type": "String"
     },
+    "InstanceUpdateBatchSize": {
+      "Default": "1",
+      "Description": "The number of instances to update in a batch",
+      "MinValue": "1",
+      "Type": "Number"
+    },
     "Key": {
       "Default": "",
       "Description": "SSH key name for access to cluster instances",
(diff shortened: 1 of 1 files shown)


ddollar:
We could also possibly add warnings around this and concurrency like we do rack scaling.

ddollar:
Removing from merge party for now, needs a better rebase

beedub:
> With a setting like InstanceUpdateBatchSize=5, are your critical services scaled up so that losing 5 web containers at a time would be ok?

I would expect that the user (me) would ensure that the batch size is set to an acceptable value. In my specific case, with 50 nodes, I'd look to set the param to 5

@ddollar just rebased and force pushed

PR #741, opened 2016-06-08
Add cpu per process

beedub:
## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

I had trouble getting fixtures and tests to work.

When running `make fixtures` according to `Development.md`, the fixtures generated looked like ones from the old `app.tmpl`, not the newer one. Running `make test` also outputted a slew of errors as well.


beedub:
I was able to fix my go env and get `make test` to pass locally. Should be ready for review!

nzoschke:
Nice!

I have some high level questions, as I've intentionally ignored what CPU shares can accomplish on ECS up until now for simplicity sake.

Can you offer some use cases for setting CPU shares differently among apps and process types?

How does settings CPU shares affect the runtime for processes? Do they get throttled or killed if they use too much? 

I believe the total number of CPU shares differs among instance type. Does the process type setting need to take this into account?

nzoschke:
Some related docs:

http://docs.aws.amazon.com/AmazonECS/latest/developerguide/task_definition_parameters.html#container_definition_environment
https://docs.docker.com/engine/reference/run/#cpu-share-constraint

beedub:
> Can you offer some use cases for setting CPU shares differently among apps and process types?

In a single app, we have a worker tier for cpu-intensive processes like image processing and a worker tier for everything else. We'd want to guarantee, for example, than the the image processing workers had atleast 2 cores available to it (represented through 2048 cpu units in ecs). Similarly, we'd want the worker tier to have atleast 1 core (represented by 1024 cpu units). It's also easy to imagine different types of apps running in one rack as well. For us, we have a main application for which each container spawns 4 workers, and we have newer microservices that only have 2 workers each. For the former, we'd like to dedicate atleast 2 cores worth of units to it, and the former 1 core.

> How does settings CPU shares affect the runtime for processes? Do they get throttled or killed if they use too much?

From what I've read and demos I've seen, they are simply throttled to that amount if there are competing processes on the box. They are *not* killed.

> I believe the total number of CPU shares differs among instance type. Does the process type setting need to take this into account?

The vCPU shares *are* different per instance type. There's a coarse check for the cpu setting in the PR that's similar to the check for the memory setting: the validation will check that the rack instance type has atleast enough overall cpu shares to satisfy the cpu setting. However, similar to memory constraints, this won't guarantee that the process can be placed. Eg. if you had one t1.small (has 1 vCPU/1024 cpu units) with the convox agent running with 256 cpu units, and tried to launch a process with 1024, then it may (will? I'll verify after this comment) not get placed.

I do think having this setting, and setting it to a sane default is important though. Without a cpu setting at all, one could launch a process accidentally that saturates cpu and adversely affect every other app and process running in the rack. Now that I think about it, I think it might be more intuitive to set the default to 512 (half a cpu core minimum) or 1024 (one core minimum).

beedub:
I tested the cpu shares on an ECS cluster. ECS will deny placing the task if there exists no capacity on any instances that fits the CPU requirement.

beedub:
I did a little bit more testing and research on cpu...

There are two aspects to consider: ecs' cpu units and docker cpu shares. They are almost similar but not exactly the same. While ecs' cpu units are a hard limit *and* relative value, docker cpu shares are only a relative value. This means that the docker daemon on the instance can handle way instances 

ECS translates cpu to docker `--cpu-shares` according to the following chart:

| cpu units | available cpu change | --cpu-shares value |
|-----------|-------------------------|--------------------|
| not specified / null / 0  | 0                       | 2                  |
| 1         | -1                       | 2                  |
| N > 1        | -N                     | N                  |

Currently, convox doesn't specify a cpu for its task definitions, so available cpu units on all instances is the max allowed and each task/process is launched with a docker cpu share of 2. As an example, if the cluster has 1 t2.small with 1 vCPU (1024 cpu units) and 3 running tasks each with share 2, then each task will be guaranteed 2/1024 cpu units and each will equally share the remaining unreserved cpu units (1024 - 2 * 3).

nzoschke:
Awesome research. Let me try to synthesize what you've found...

CPU shares are an important concept for apps, particularly for controlling how many cores a container can access / saturate.

On ECS, CPU shares are a hard limit for scheduling.

I don't quite get the Docker CPU shares stuff... My initial reaction is that it doesn't matter, just work with ECS.

---

> I do think having this setting, and setting it to a sane default is important though. Without a cpu setting at all, one could launch a process accidentally that saturates cpu and adversely affect every other app and process running in the rack.

I think the default is the main design decision here.

My Convox philosophy is sane defaults, then some simple knobs, but let power users do whatever they want with these knobs including shooting yourself in the foot.

I definitely understand your concern about a process gone wild.

But I think that saturating a CPU is actually a feature in a single-tenant system. Squeeze everything you can out of the boxes, and only tune things if you observe quality of service problems.

The main justification I see for this default is the ECS API and system itself. For every Task Definition, Memory is required and CPU is not.

The next justification is backwards compatibility. If we're not careful the next app deploy for all users will magically have some tight CPU limits. If the default is wrong, we should change it, but probably with a more careful migration strategy.

So what if the default was still 0, but your parameterization lands so you and everyone else can start playing with tuning it?

beedub:
After re-considering, I agree that leaving the cpu default as null makes
more sense than setting it to something arbitrary. As long as it's
configurable, I'm :+1:

mwarkentin:
I'm 👍 to that as well. Our main concern is when we start to put workers processing asynchronous tasks - we don't want those workers saturating CPU and causing performance issues on web requests, etc.

beedub:
@nzoschke updated

beedub:
> Limiting those workers specifically would be fine for us.

@mwarkentin to be clear, I think you meant not limiting those workers, but rather guaranteeing min cpu for your web tier

mwarkentin:
@beedub Guess I was thinking about this from the other way around.. but reserving CPU for specific processes (like web) makes sense from a scheduling perspective I suppose.

api/models/formation.go
```diff
@@ -118,6 +121,14 @@ func SetFormation(app, process string, count, memory int64) error {
 		params[fmt.Sprintf("%sMemory", UpperName(process))] = fmt.Sprintf("%d", memory)
 	}
 
+	if cpu > 0 {
+		if cpu > capacity.InstanceCPU {
+			return fmt.Errorf("requested cpu %d greater than instance size %d", cpu, capacity.InstanceCPU)
```
nzoschke: Great validation. Though this makes me wonder: what happens if I set this, then scale my rack instance sizes down a bunch. I think we could do some nominal validation on the rack scale API too.

api/models/templates/app.tmpl
```diff
@@ -150,12 +145,18 @@
...
+    "Blank{{ upper $k }}Cpu": { "Fn::Equals": [ { "Ref": "{{ upper $k }}Cpu" }, "-1" ] },
     "Disabled{{ upper $k }}DesiredCount": { "Fn::Not": [{ "Fn::Equals": [ {"Ref": "{{ upper $k }}DesiredCount" }, -1 ] }] },
   {{ end }}
 {{ end }}
 
 {{ define "process-params" }}
   {{ range $e := . }}
+    "{{ upper $e.Name }}Cpu": {
+      "Type": "Number",
+      "Default": "-1",
```
nzoschke: Can we do `Default: 0` and always pass it through? I think your logic is correct as-is but not needing a Condition would be simpler.

api/provider/aws/apps_test.go
```diff
@@ -46,13 +46,13 @@ func TestAppGet(t *testing.T) {
...
 		Parameters: map[string]string{
 			"WebMemory":              "256",
+			"WebCpu":                 "256",
 			"Release":                "RVFETUHHKKD",
 			"Subnets":                "subnet-13de3139,subnet-b5578fc3,subnet-21c13379",
 			"Private":                "Yes",
 			"WebPort80ProxyProtocol": "No",
 			"VPC":                  "vpc-f8006b9c",
 			"Cluster":              "convox-Cluster-1E4XJ0PQWNAYS",
-			"Cpu":                  "200",
```
nzoschke: Good catch! I now recall we played with an app-wide Cpu parameter before, but stopped doing that and went back to Cpu=0 everywhere.

api/provider/aws/capacity.go
```diff
@@ -34,7 +34,10 @@ func (p *AWSProvider) CapacityGet() (*structs.Capacity, error) {
 			if *resource.Name == "MEMORY" {
 				capacity.InstanceMemory = *resource.IntegerValue
 				capacity.ClusterMemory += *resource.IntegerValue
-				break
+			}
+			if *resource.Name == "CPU" {
```
nzoschke: I took this calculation out recently, and added pulling instance CPU metrics from CloudWatch. 

I think this change back could be good though... There are 2 CPU concepts. How much is reserved due to scheduling and how much is utilized according to the kernel.

I believe this is similar to the `Memory` vs `Size` values in the API.

nzoschke: Ah nice, just got to the ClusterCPU, InstanceCPU and ProcessCPU properties.

cmd/convox/scale.go
```diff
@@ -35,6 +39,7 @@ func cmdScale(c *cli.Context) error {
 	// initialize to invalid values that indicate no change
 	count := -2 // -1 is valid, indicates removing the process and ELB
 	memory := -1
+	cpu := -1
```
nzoschke: This makes total sense in this patch, but adding some more context while it's fresh... Recently we got bit by the client posting these magic values to indicate no change. A better design would be that the client doesn't post the field at all if it wasn't specified.

api/models/templates/app.tmpl
beedub: 👍

api/models/formation.go
beedub: I would assume that the service would throw the [`not enough cpu`](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/troubleshooting.html) error

beedub:
updated the app.tmpl's default from -1 to 0, which seemed like the only blocking change. I would test the behavior of scaling down the rack, but I can't seem to get `convox ps` to work against my local dev

cmd/convox/ps.go
```diff
@@ -52,18 +52,18 @@ func cmdPs(c *cli.Context) error {
 	}
 
 	if c.Bool("stats") {
-		t := stdcli.NewTable("ID", "NAME", "RELEASE", "SIZE", "CPU", "MEM", "STARTED", "COMMAND")
+		t := stdcli.NewTable("ID", "NAME", "RELEASE", "CPU", "SIZE", "CPU %", "MEM %", "STARTED", "COMMAND")
```
beedub: The `Size` term here seems a bit confusing but left alone

cmd/convox/scale.go
beedub: this seems relatively easy to do. I was imagining having a central mapping between count/cpu/mem "noop" vals somewhere and then referencing that here and in `SetFormation`. I would have made this change myself, but not sure how it would look in practice (new to go)

nzoschke:
Thanks for the default 0 change. We can carry this over the finish line on our end now. The next steps I see are:

- [ ] Development testing and verification
- [ ] Make a final design decision on CPU / Memory API fields. Will probably aim for minimal changes now
- [ ] Make a call on cleaning up the no-changn client/server pattern
- [ ] Release checklist

Thanks again @beedub for all the great research and development here!

nzoschke:
Testing this now, and I'm seeing that the `WebCpu` parameter is there, and the default of 0 is correct. But I caught that the value isn't actually getting set or updated in the TaskDefinition. 

This needs a change in the ECS Task Definition custom handler:

https://github.com/convox/rack/blob/master/api/cmd/formation/handler/ecs.go#L400

I'll make this change, putting notes here for future reference.

nzoschke:
https://github.com/convox/rack/pull/779/commits/251fc96813c0b6ac77b1f11699e8f0ce3a9b09a4 is testing well with setting the Cpu value on the Task Definition.

@beedub I noticed that these commits are from `Brian Hahn <bhahn@Brians-MacBook-Pro-2.local>`, perhaps you want to use a different email for better attribution?

beedub:
> you want to use a different email for better attribution?

Ah, thanks for the notification. Updated

nzoschke:
Caught another bug, it's valid to scale Cpu back to 0. 0875a752950c8acbc57d748b1b4676995174d959

nzoschke:
Thanks! Closing this and testing / releasing on #779. Won't release until next week though.

PR #743, opened 2016-06-08
Capture `convox run` output in a log

MiguelMoll:
The use of `tee` will allow the output to be displayed to STDOUT and logged to an artifact file.

Previous behavior, the output was being swallowed by `grep` and made it harder to debug what was going on.

### DIFF
--- a/ci/tests/example-app
+++ b/ci/tests/example-app
@@ -87,4 +87,4 @@ wait_for_app_deployment
 
 convox ps --app $APP_NAME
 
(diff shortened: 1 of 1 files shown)


nzoschke:
LGTM, there are a couple other places in example-app where we pipe to grep. Can you apply this pattern there too?

MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;a8kv016Q&#x2F;235-ci-shell-script-visibility"> CI Shell Script Visibility</a>

PR #747, opened 2016-06-09
Clarify AWS security groups for local rack development

MiguelMoll:
When PR #716 was merged, racks running locally were not able to SSH into the instances. With this change, any future modifications will be handled for local racks.

Figured opening all ports (0 - 65535) for the "My IP" source will help in the future. Thoughts?


nzoschke:
Looks good to me. 0 - 65535 was surprising at first but I do like the future proofing.

PR #754, opened 2016-06-10
Improving `convox ps` reliability

MiguelMoll:
Following changes are to make `convox ps` act a bit more reliable when encountering errors while gathering data. The goal is to log errors (if any) and fail gracefully.

### DIFF
--- a/api/models/process.go
+++ b/api/models/process.go
@@ -271,7 +271,7 @@ func ListProcesses(app string) ([]*Process, error) {
 }
 
 // ListPendingProcesses tries to get a list of all pending processes.
-// If unable to connect to a docker daemon, or error out for another reason, it will bypass that instance and continue to other daemons.
+// If unable to gather needed data it will continue to other processes.
 func ListPendingProcesses(app string) (Processes, error) {
 	// In AWS ECS, pending processes would present themselves during a deployment.
 	pss := Processes{}
--- a/api/models/process.go
+++ b/api/models/process.go
@@ -204,7 +204,6 @@ func ListProcesses(app string) ([]*Process, error) {
 		}
 
 		if ci == nil {
-			// Error out if it didn't find an ecs instance? Maybe just continue?
 			return nil, fmt.Errorf("could not find ECS instance")
 		}
 
@@ -218,7 +217,6 @@ func ListProcesses(app string) ([]*Process, error) {
 		}
 
 		if ec2i == nil {
-			// Error out if it didn't find an ec2 instance? Maybe just continue?
 			return nil, fmt.Errorf("could not find EC2 instance")
 		}
 
@@ -243,7 +241,7 @@ func ListProcesses(app string) ([]*Process, error) {
 		}
 	}
 
-	pss := make([]*Process, 0)
+	pss := Processes{}
 
 	for i := 0; i < num; i++ {
 		select {
@@ -255,23 +253,21 @@ func ListProcesses(app string) ([]*Process, error) {
 	pending, err := ListPendingProcesses(app)
 	if err != nil {
 		fmt.Printf("ns=kernel at=ListProcesses state=error message=\"unable to get pending processes: %s\"\n", err)
+	} else {
+		pss = append(pss, pending...)
 	}
-
-	pss = append(pss, pending...)
 
 	oneoff, err := ListOneoffProcesses(app)
-
 	if err != nil {
 		fmt.Printf("ns=kernel at=ListProcesses state=error message=\"unable to get one-off processes: %s\"\n", err)
+	} else {
+		pss = append(pss, oneoff...)
 	}
-
-	pss = append(pss, oneoff...)
 
 	return pss, nil
 }
 
 // ListPendingProcesses tries to get a list of all pending processes.
-// If unable to gather needed data it will continue to other processes.
 func ListPendingProcesses(app string) (Processes, error) {
 	// In AWS ECS, pending processes would present themselves during a deployment.
 	pss := Processes{}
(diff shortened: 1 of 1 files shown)


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;9Ejn8N6X&#x2F;217-interstate-convox-ps-broken-right-after-deploy"> interstate `convox ps` broken right after deploy</a>

api/models/process.go
```diff
@@ -217,6 +218,7 @@ func ListProcesses(app string) ([]*Process, error) {
 		}
 
 		if ec2i == nil {
+			// Error out if it didn't find an ec2 instance? Maybe just continue?
```
nzoschke: I think the behavior of a total error of ECS or EC2 API calls fail is ok for now. All the data points to the Docker parts being the main problem.

Though I think that the best design would be saying the ECS APIs are canonical and even EC2 data is nice to have.

api/models/process.go
```diff
@@ -247,32 +249,32 @@ func ListProcesses(app string) ([]*Process, error) {
...
-		case err := <-errch:
-			return nil, err
 		}
 	}
 
 	pending, err := ListPendingProcesses(app)
-
 	if err != nil {
-		return nil, err
+		fmt.Printf("ns=kernel at=ListProcesses state=error message=\"unable to get pending processes: %s\"\n", err)
```
nzoschke: Are there particular errors that you found down this path? As this is a pure ECS call, I hope that it's fairly reliable.

I also think that surfacing pending processes during a deploy is important visibility... Though I'm not sure if a total error would be any better than a process list without pending processes.

api/models/process.go
```diff
@@ -247,32 +249,32 @@ func ListProcesses(app string) ([]*Process, error) {
...
 	}
 
 	pss = append(pss, pending...)
 
 	oneoff, err := ListOneoffProcesses(app)
 
 	if err != nil {
-		return nil, err
+		fmt.Printf("ns=kernel at=ListProcesses state=error message=\"unable to get one-off processes: %s\"\n", err)
```
nzoschke: This looks right. This should definitely be a best effort addition to the process list and not cause a total error.

api/models/process.go
```diff
@@ -247,32 +249,32 @@ func ListProcesses(app string) ([]*Process, error) {
...
 
 	pss = append(pss, oneoff...)
 
 	return pss, nil
 }
 
-func ListPendingProcesses(app string) ([]*Process, error) {
-	pss := make([]*Process, 0)
+// ListPendingProcesses tries to get a list of all pending processes.
+// If unable to connect to a docker daemon, or error out for another reason, it will bypass that instance and continue to other daemons.
```
nzoschke: This comment isn't right. From my reading of the code ListPendingProcesses talks to CloudFormation to get all the ECS Service names for a stack, then talks to ECS to describe the services and their deployment status.

It doesn't talk to Docker or an individual instance directly.

api/models/process.go
MiguelMoll: No direct error was observed. It was strictly "failing gracefully" in the event that an ECS call doesn't work out.

api/models/process.go
```diff
@@ -294,9 +296,9 @@ func ListPendingProcesses(app string) ([]*Process, error) {
 			tres, err := ECS().DescribeTaskDefinition(&ecs.DescribeTaskDefinitionInput{
 				TaskDefinition: d.TaskDefinition,
 			})
-
 			if err != nil {
-				return nil, err
+				fmt.Printf("ns=kernel at=ListPendingProcesses state=error message=\"%s\"\n", err)
```
nzoschke: I see that line 282 can return an error. I think this particular function should pick one strategy or another in both places and either log and not error or error. If I had to make the call I'd say it can still error in both places.

api/models/process.go
```diff
@@ -326,6 +328,8 @@ func ListPendingProcesses(app string) ([]*Process, error) {
 	return pss, nil
 }
 
+// ListOneoffProcesses tries to get a list of all one-off processes.
+// If unable to connect to a docker daemon, it will bypass (instead of erroring out) and continue to other daemons.
```
nzoschke: This looks great.

api/models/process.go
```diff
@@ -422,19 +437,17 @@ func fetchProcess(app string, task ecs.Task, td ecs.TaskDefinition, cd ecs.Conta
...
 	}
 
 	ps.Host = ip
 
 	d, err := ps.Docker()
-
 	if err != nil {
-		errch <- fmt.Errorf("could not communicate with docker")
+		fmt.Printf("ns=kernel at=processes.list state=error message=\"%s\"\n", err)
+		psch <- ps
```
nzoschke: These look good to me. Should we take out the error channel stuff altogether?

api/models/process.go
```diff
@@ -531,15 +545,15 @@ func (p *Process) Docker() (*docker.Client, error) {
 	return Docker(fmt.Sprintf("http://%s:2376", p.Host))
 }
 
+// FetchStats attempts to gather stats from docker.
+// Does nothing if unable to connect to the daemon.
```
nzoschke: Looks good to me!

api/models/process.go
MiguelMoll: My line of thinking here was if `GetAppServices()` did fail there little else to do. If one ECS call failed, other could still pass. After looking at it again, I can see how line 282 could be changed to not bail out entirely so that running and one-off processes could still be displayed.

nzoschke:
Nice, this looks good and I'm +1 releasing as-is. In review I did see:

- [ ] Making a decision about ListPendingProcesses causing an error or not everywhere
- [ ] Cleaning up some unused stuff like the errch

In final review I think ListPendingProcesses should still error and the comment should be updated to reflect this, and the errch stuff is ok to leave in.

api/models/process.go
MiguelMoll: For now, indeed it seems ok. I was thinking along the lines if one ECS/EC2 call fails, there could still be others that working fine. Do we want to ignore those?

Issue #613, opened 2016-05-09
Configure ELB health-checks


dlanger:
Similar to #594 (in the vein of "more ELB configuration options"), it'd be nice if we could configure ELB health-checks. 

Right now, they default to a TCP connection to the exposed port. This works as a lowest-common-denominator, but misses out on a lot of the power of ELB health-checks. In an ideal world, I'd be able to pick an HTTP health-check and specific a specific path to use (which I could point at [`health_check`](https://github.com/ianheggie/health_check), [`django-watchman`](https://github.com/mwarkentin/django-watchman), etc.)

Open questions:
* Where do we specify the path to health-check? Probably a `docker-compose.yaml` label
* Should the health-check [type](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/elb-healthchecks.html) be individually selectable (thought a `docker-compose.yaml` label), or should it be coupled to the ELB protocol?

@nzoschke @ddollar Thoughts?

ddollar:
We can probably do this with more labels, similar to https://convox.com/docs/load-balancers/

shin-nien:
Also being able to configure the timeout, intervals and thresholds would be useful too.

I'm deploying a fairly heavy java app that takes a while to start up so at the moment I'm re-configuring the ELB once it's created so my app can start without getting killed.

nzoschke:
The configuration for every app stack (ELB, TaskDefinition, and Services) is in `app.tmpl`. The health check for each ELB is here:

https://github.com/convox/rack/blob/master/api/models/templates/app.tmpl#L357

This is a CloudFormation template that we run it through Go templating to make resources for every process type.

I think you'll want to Parameterize the Timeout or UnhealthyThreshold. 

You can follow the same pattern as `{{ upper $balancer.ProcessName }}Port{{ .Balancer }}Protocol`. It has a Parameter for every process type with a default value here:

https://github.com/convox/rack/blob/master/api/models/templates/app.tmpl#L282

Which can be used instead of the hard coded `3` here:

https://github.com/convox/rack/blob/master/api/models/templates/app.tmpl#L361

You can run the rack API on your laptop following the Development.md guide. Make sure to see the https://github.com/convox/rack/blob/master/Development.md#infrastructure-changes section. When you make changes to app.tmpl, you then need to run `make templates`, then wait for the api server to reload, then you can:

```
# login to development API with your changes
$ convox login localhost

# deploy the app to apply the new CF stack template
$ convox deploy

[...]

PR #755, opened 2016-06-12
[RFC] Add support for health check params

beedub:
This makes elb health check params configurable via convox labels. The added convox labels are:

1. convox.health_check.port
1. convox.health_check.path (applicable to http and https)
1. convox.health_check.protocol
1. convox.health_check.timeout
1. convox.health_check.healthy_threshold
1. convox.health_check.unhealthy_threshold
1. convox.health_check.interval

The port refers to the container port, which is mapped back to the random port.

##### Links
* http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-elb-health-check.html

##### Questions / Notes
* Is the design appropriate?
* I'm still a bit unclear about the rules for setting defaults in the cloudformation template according to convox. When does it make sense to specify a default?
* I have an inkling to add or update integration tests but could use some guidance here
* I'm new to go, so language-specific comments are appreciated!

##### Todos
- [x] Add validation that timeout must be less than interval

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI


mwarkentin:
Previous issue: #613

beedub:
(Oops, forgot to search for existing issues)

ddollar:
Overall looks good but a few comments:

Let's try to avoid too much length (and underscores) so probably convox.health.*

Is there some way we could describe all of this in one or two parameters?

Eg 

convox.health.endpoint=web:9000/foo
convox.health.timeout=60

And derive the rest from those. Would like to avoid getting too tied down to exact Amazon implementation detail in the docker compose or rack api if possible. 

For things that absolutely need to be exposed from down in the aws plumbing let's use cloudformation parameters.

beedub:
> Let's try to avoid too much length (and underscores) so probably convox.health.*

👍 

> Is there some way we could describe all of this in one or two parameters? And derive the rest from those. Would like to avoid getting too tied down to exact Amazon implementation detail in the docker compose or rack api if possible.

The protocol, port, and path could certainly be combined into `endpoint` but the other concepts/params are more-or-less common with the other two other major providers, GC and Azure, and not AMZN-specific

[GC health check params](https://cloud.google.com/compute/docs/load-balancing/health-checks) map essentially 1:1 with AWS', and so do [Azure's probes](https://msdn.microsoft.com/en-us/library/azure/Mt163574.aspx) except for the `healthy_threshold` concept (they do support the concept of `unhealthy_threshold`, which maps to Azure's `numberOfProbes`).

Is the concern strictly about making `docker-compose` consistent between local and cloud? or consistent between various clouds?

mwarkentin:
Also wondering what that would look like for TCP health checks? Although maybe the default configuration there is enough and it's the other params that might be interesting to tweak further (timeout, threshold, interval).

beedub:
I think @ddollar's example meant to say `convox.health.endpoint=tcp:9000/foo` as the label would already be under the process

mwarkentin:
Oh true, I was thinking that it would be more of a full url, but that makes sense.

nzoschke:
> Is the design appropriate?

The labels to parameters pattern is working out well, so yes I think this is a good design in general.

+1 to David's suggestion of shortening / combining these as much as possible. 

The things I see to tune are:

* Timeout
* Turn it into a path-based HTTP check
* Explicitly pick a port if I've mapped many

So I propose 3 labels: `port`, `path`, and `timeout`.

We already know what container ports should be alive from the port mapping and what protocols the container ports talks from the `convox.port.443.protocol=tcp|http` and `convox.port.443.secure=true` labels.

We can always pick a naive default of TCP to any port that is exposed.

Here are some other use cases I see...

### My app takes a long time to boot so I add a single new label:

```yaml
web:
  labels:
    - convox.health.timeout=120
  ports:
    - 80:5000
```

### My service exposes a "deep" health check endpoint over HTTP so I add a single new label:

```yaml
web:
  labels:
    - convox.health.path=/foo
  ports:
    - 80:5000
```

### I'm a total power user and want to assert health on my end-to-end encrypted service (not on the redirect port):

```yaml
web:
  labels:
    - convox.port.443.protocol=https
    - convox.port.443.secure=true
    - convox.health.path=/foo
    - convox.health.port=443
    - convox.health.timeout=120
  ports:
    - 80:5000
    - 443:5001
```

nzoschke:
> I'm still a bit unclear about the rules for setting defaults in the cloudformation template according to convox. When does it make sense to specify a default?

Most parameters need default values. The exceptions are:

* Convox might supply its own values on every stack update (i.e. rack/app Version parameter)
* The Parameter has a related Conditional that might toggle something off if the value is empty (i.e. Ami parameter and BlankAmi conditional)

nzoschke:
> I have an inkling to add or update integration tests but could use some guidance here

Adding some more YAML -> CF fixtures would be appropriate here. Add something like `custom-health.yml` that has the new labels to rack/api/models/fixtures, then type `make fixtures`. It should generate the cloudformation .json.

This will require manual verification but once the .json is considered correct but all of us it will help with regression testing and patches going forward.

mwarkentin:
Is multiple health checks a thing? It sounds like it from this:

> We can always pick a naive default of TCP to any port that is exposed.

Just wondering if that's a valid use case that can somehow be covered. Would it be a similar case to the protocol setting? eg. `convox.port.443.health.timeout=120`?

nzoschke:
No there is one health check per ELB

beedub:
A few comments about the proposed use-cases:

> My app takes a long time to boot so I add a single new label:
```yaml
web:
  labels:
    - convox.health.path=/foo
  ports:
    - 80:5000
```

The default port protocol is currently `tcp`, but specifying path would implicitly change the port's protocol to `http`, which creates the relationship that the health check has some ownership of the path's protocol (ie. the protocol isn't independent anymore). I'm not sure if this would be a practical problem long-term, but something to note.

So is there opposition to adding `healthy_threshold`, `unhealthy_threshold` and `interval` as well? They're common to the 3 major cloud providers (except for azure not supporting healthy_threshold). As these health check params would be additional labels, they would be only specified for the power users

mwarkentin:
I'm not sure if I like the health check setting updating the protocol for the port - but there could be some validation if you tried setting it without setting protocol to `http`/`https`?

> You tried to set an HTTP health check on a TCP listener. You can change the protocol by adding the following label to `docker-compose.yml`: `convox.port.443.protocol=https`

ddollar:
+1 to Noah's suggested labels

nzoschke:
From my playing with ELB it seems like you can specify an HTTP health check endpoint even if the ELB is in TCP mode. So I think the two are decoupled.

nzoschke:
@beedub a few reasons I have against parameterizing health_threshold, unhealthy_threshold and interval...

Good "developer experience" comes from removing complexity and optionality. From my experience, `boot timeout` is the concept everyone cares about and understands for their app, interval and threshold is implementation details.

There is a hard limit of max 60 parameters on every CF stack, and folks that have a bunch of process types are already hitting this. Adding a lot of new parameters could push more apps over the edge.

There's always a super dangerous power mode where you can tweak any settings not parameterized on AWS directly. While I don't recommend this in general, it's a great way for us to experiment before making a label / knob something we add to the platform.

beedub:
They're decoupled afaik as well. @nzoschke are you implying that the health check proto would also be decoupled at the convox layer?

eg. taking the aforementioned example again:

```
web:
  labels:
    - convox.health.path=/foo
  ports:
    - 80:5000
````

* ELB ingress would be `TCP:80`
* Instance ingress would be `TCP:5000`
* Health check would be `HTTP:5000/foo`

But in this example:

```
web:
  labels:
    - convox.port.443.protocol=https
    - convox.health.path=/foo
    - convox.health.port=443
    - convox.health.timeout=120
  ports:
    - 443:5001
```

* ELB ingress would be `HTTPS:443`
* Instance ingress would be `HTTPS:5001`
* Health check would be `HTTPS:5001/foo`

nzoschke:
I think its:

* TCP to the first port by default
* HTTP to the first port if there is a `.path` label
* HTTPS if there is a `.path` label and the first port or specified port also has a `protocol=https` label

So the `convox.port.443.protocol` and `convox.health.port=443` labels are coupled.

I do have a sense that this is pretty "clever", but I think this stuff is pretty tricky in general and we can have validation and feedback on `convox start` and good tests about yaml -> parameters.

beedub:
I'm getting a bit confused. The health check is between the LB and container instance, so shouldn't the `secure` label determine whether the health check protocol is `http` vs `https`?

nzoschke:
You are right, it's the secure label. I'll fix up previous comments for clarity.

nzoschke:
And you are correct in your last example. The health check will be `HTTP:5001/foo`

beedub:
To the same effect, shouldn't `convox.health.port` refer to the container port, not the LB ingress port?

nzoschke:
I think it could go either way since there's always a mapping. But yeah container port makes sense there too

Issue #379, opened 2016-02-24
Disallow operating on stacks that do not belong to us


anthonyrisinger:
Recently I used:

```
# convox apps params set --app=WRONGRACK InstanceCount=1
```

To manually scale a broken api build refusing to update itself. Unfortunately, I scaled the wrong rack! It would be nice if convox confirmed the target stack matches one of:

```
stack.Name == os.Getenv("RACK")
stack.Tags["Rack"] == os.Getenv("RACK")
```

to avoid operating on stacks not owned by the current rack.

nzoschke:
I strongly support this. Accidentally doing stuff to other stacks could be disastrous.

nzoschke:
The API that is serving a request knows it's own rack name from os.Getenv("RACK"). So the GetApp or GetRack calls should verify that they only access and operate on stacks that belong to the rack. This can be validated by the stack name itself, as well as the tags that stack has.

nzoschke:
The most important things to validate are anything that will change a stack like a CloudFormation StackUpdate or StackDelete call

MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;7sIdDSju&#x2F;251-disallow-operating-on-stacks-that-do-not-logged-into"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> Disallow operating on stacks that do not logged into</a>

MiguelMoll:
Merged with PR #770

PR #764, opened 2016-06-14
Allow a rack to only operate on itself

MiguelMoll:
This fix address the issue where more than one rack in the same AWS account
and region from affecting one another. Fixes issue #379.

### DIFF
--- a/api/models/app.go
+++ b/api/models/app.go
@@ -71,8 +71,15 @@ func GetApp(name string) (*App, error) {
 		app, err = getAppByStackName(name)
 	}
 
-	if app != nil && app.Tags["Rack"] != "" && app.Tags["Rack"] != os.Getenv("RACK") {
-		return nil, fmt.Errorf("no such app: %s", name)
+	if app != nil {
+		if app.Tags["Rack"] != "" && app.Tags["Rack"] != os.Getenv("RACK") {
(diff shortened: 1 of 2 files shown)


api/models/app.go
```diff
@@ -71,8 +71,15 @@ func GetApp(name string) (*App, error) {
...
 	}
 
-	if app != nil && app.Tags["Rack"] != "" && app.Tags["Rack"] != os.Getenv("RACK") {
-		return nil, fmt.Errorf("no such app: %s", name)
+	if app != nil {
+		if app.Tags["Rack"] != "" && app.Tags["Rack"] != os.Getenv("RACK") {
+			return nil, fmt.Errorf("no such app: %s", name)
+
+		} else if len(app.Tags) == 0 && name != os.Getenv("RACK") {
+			// This checks for a rack. An app with zero tags is a rack (this assumption should be addressed).
```
MiguelMoll: Nice intel. That is definitely an alternative way as long as only racks have that Dashboard key.

MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;7sIdDSju&#x2F;251-disallow-operating-on-stacks-that-user-isn-t-not-logged-into"> Disallow operating on stacks that user isn&#39;t not logged into</a>

PR #767, opened 2016-06-14
AppShow via the controller only returns an app

### LINKED ISSUE #633: Deploying to the rack name gives confusing error
If you run `convox deploy -a $rackname` everything looks like it tries to work and then you get a very confusing error message about an S3 bucket. We should validate the app name parameter here and return a helpful error message.

/cc @kmonkeyjam

MiguelMoll:
If asked for a rack via the AppShow method an error will be returned. Displays a nice error which fixes #633.

### DIFF
--- a/api/controllers/apps.go
+++ b/api/controllers/apps.go
@@ -17,7 +17,6 @@ import (
 
 func AppList(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	apps, err := models.ListApps()
-
 	if err != nil {
 		return httperr.Server(err)
 	}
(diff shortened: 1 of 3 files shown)


MiguelMoll:
Before (`dev` being the currect rack):
``` bash
⟩ convox deploy -a dev
Deploying misc
Creating tarball... OK
Uploading... ERROR: exit status 1
```

After:
``` bash
⟩ convox deploy -a dev
Deploying dev
ERROR: rack dev is not an app
```

Using a rack that's not the current one:
``` bash
⟩ convox deploy -a misc
Deploying misc
ERROR: invalid rack: misc
```

MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;VQIadVqD&#x2F;254-prevent-the-deployment-of-rack"> Prevent the deployment of rack</a>

api/controllers/apps.go
```diff
@@ -30,8 +29,11 @@ func AppList(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 func AppShow(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	app := mux.Vars(r)["app"]
 
-	a, err := models.GetApp(mux.Vars(r)["app"])
+	if app == os.Getenv("RACK") {
+		return httperr.Errorf(404, "rack %s is not an app", app)
```
nzoschke: Nice, I like the message and error code here.

nzoschke:
👍 looks good to me. Builds / deploys will fail on their app/status lookup, but the logs endpoint will still work.

We'll want to continue to tighten the validation of app and rack names -> stacks, but the most bang for the buck will be doing that as part of another provider interface refactor iteration.

PR #773, opened 2016-06-16
cleanup old builds after build creation

### LINKED ISSUE #361: Handle ECR 500 image limit
An ECR repository is limited to 500 images.  http://docs.aws.amazon.com/AmazonECR/latest/userguide/service_limits.html

Once a repo hits the limit you can no longer deploy the corresponding app. We should rotate out old images. This probably has implications for the build and release history too, since you won't be able to roll back to a release for which the images have been deleted.

awsmsrc:
fixes #361 

Currently ECR has a 500 image limit.

This causes builds and deployments to fail.

This approach may be too brute force, however we expect Amazon to remove this limit over time so want to avoid putting lots of engineering time into this hopefully temporary fix.

The idea here is to simply delete older builds, this will be done in batches on each build to remove old builds for existing users and will generally only delete the 51st oldest build from there.

Currently you can only access the 20 most recent bulids anyway so this should not break anyone's existing workflows.

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

### DIFF
--- a/api/controllers/builds.go
+++ b/api/controllers/builds.go
@@ -193,9 +193,9 @@ func BuildUpdate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 		if err != nil {
 			fmt.Println("Error listing builds for cleanup")
 		} else {
-			if len(bs) >= 4 {
+			if len(bs) >= 50 {
 				wg := new(sync.WaitGroup)
-				outDated := bs[4:]
+				outDated := bs[50:]
 				for _, b := range outDated {
 					wg.Add(1)
 					go func(buildId string, wg *sync.WaitGroup) {
--- a/api/controllers/builds.go
+++ b/api/controllers/builds.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -21,7 +22,21 @@ import (
 func BuildList(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	app := mux.Vars(r)["app"]
 
-	builds, err := provider.BuildList(app)
+	l := r.URL.Query().Get("limit")
+
+	var err error
+	var limit int
+
+	if l == "" {
+		limit = 20
+	} else {
+		limit, err = strconv.Atoi(l)
+		if err != nil {
+			return httperr.Errorf(400, err.Error())
+		}
+	}
+
+	builds, err := provider.BuildList(app, int64(limit))
 	if awsError(err) == "ValidationError" {
 		return httperr.Errorf(404, "no such app: %s", app)
 	}
--- a/api/controllers/builds.go
+++ b/api/controllers/builds.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/convox/rack/api/helpers"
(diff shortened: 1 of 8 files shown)


awsmsrc:
@miguel @nzoschke please review

api/controllers/builds.go
```diff
@@ -21,7 +22,21 @@ import (
 func BuildList(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	app := mux.Vars(r)["app"]
 
-	builds, err := provider.BuildList(app)
+	l := r.URL.Query().Get("limit")
+
+	var err error
+	var limit int
+
+	if l == "" {
```
nzoschke: This API change looks good to me for backwards compatibility.

api/provider/provider.go
```diff
@@ -20,7 +20,7 @@ type Provider interface {
 	BuildCreateTar(app string, src io.Reader, manifest, description string, cache bool) (*structs.Build, error)
 	BuildDelete(app, id string) (*structs.Build, error)
 	BuildGet(app, id string) (*structs.Build, error)
-	BuildList(app string) (structs.Builds, error)
+	BuildList(app string, limit int64) (structs.Builds, error)
```
nzoschke: Food for thought, a while back we discussed a ListOptions struct that would have limit and sort properties.

client/builds.go
```diff
@@ -35,6 +35,18 @@ func (c *Client) GetBuilds(app string) (Builds, error) {
 	return builds, nil
 }
 
+func (c *Client) GetBuildsWithLimit(app string, limit int) (Builds, error) {
+	var builds Builds
+
+	err := c.Get(fmt.Sprintf("/apps/%s/builds?limit=%d", app, limit), &builds)
+
```
nzoschke: whitespace

api/controllers/builds.go
awsmsrc: Yeah I would like an idomatic way to do pagination, limits, filters with a QueryOptions type but this was just a pragmatic solution

client/builds.go
awsmsrc: copied from the original func above, will fix both

nzoschke:
My first reaction is surprise to see this in `cmd/build/main.go` file. I consider that to be as simple of a shell around `git pull && docker build` as possible and to leave other concerns like background work up to the API.

The list 250 and delete anything over 200 is a clever strategy to get around APIs or calls to get the length of the collection and delete things in reverse.

One other big gotcha I see is that a build can produce many images (web, worker, redis and postgres). So 200 builds could be more than 500 images and trigger the limits. A couple solutions I see to this is to pick an even smaller max number of builds, or to introspect ECR to see how many images it has and use that as criteria to delete the oldest builds.

awsmsrc:
The reason I went with the cmd/build app was because it was the "lightest touch" to add a new and unproven strategy. That said I can't think of an objectively better place to put this functionality. Perhaps in the build.Create function?

Do you think 100 old builds is enough?

nzoschke:
In my anecdotal experience users care about:

* The last build in case of emergency rollback
* The last couple weeks of builds at most behind that

So I think 100 would be more than enough.

It still really bugs me that AWS is making us pick a number like this. I anticipate some day there will be unlimited images in ECR and we can remove this.

nzoschke:
This could go in https://github.com/convox/rack/blob/master/api/provider/aws/builds.go#L409

That is what the cmd/build calls back into with final build status and logs.

nzoschke:
👍 Thanks! Closing this and testing/releasing on #779. Won't release until next week though.

PR #774, opened 2016-06-16
Add some clarity to what the login password can accept

### LINKED ISSUE #709: convox login should be more clear about password vs api key
The cli language for `convox login` uses the term `password`, but it's actually an api key, which is actually the language used on the console website

MiguelMoll:
Small tweak that should fix #709. The usage prompt states that a Console API key is also accepted.

## Quick CLI Release Playbook
- [x] Code review
- [ ] Merge into master
- [ ] Release CLI

### DIFF
--- a/cmd/convox/login.go
+++ b/cmd/convox/login.go
@@ -29,7 +29,7 @@ func init() {
 		Flags: []cli.Flag{
 			cli.StringFlag{
 				Name:  "password, p",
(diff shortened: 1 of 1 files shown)


nzoschke:
I think a copy change is sufficient. But how about something shorter:
```
$ convox login --help
convox login: log into your convox rack

Usage:
  convox login [hostname]

Options:
   --password value, -p value	Rack Password or Console API Key. If not specified, prompt.
```

MiguelMoll:
I like it :+1:

PR #790, opened 2016-06-21
Security Group inbound rules

### LINKED ISSUE #720: Security Group inbound rules
A couple of the security groups created by Convox have inbound rules allowing 0.0.0.0/0 where I would have expected to see my VPC CIDR. For example, I have a linked process in my `docker-compose.yml` with a single port, correctly resulting in the creation of an internal load balancer. However, the associated security group inbound rules look as follows:

internal elb inbound

Looking at the [AWS documentation](http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/elb-security-groups.html#elb-vpc-security-groups), I would have expected to see my VPC CIDR instead of 0.0.0.0/0. The security group with tag value "BalancerSecurityGroup" also has some 0.0.0.0/0 rules. The security groups for the web process, redis service, postgres service, and the one with tag value "SecurityGroup" all look as I would have expected.

Happy to provide additional info if needed.

MiguelMoll:
These changes address a security group settings depending if an app needs an internal ELB or if the rack api is install via `--private-api`.

Not sure about hard coding the internal VPC cidr in `api/models/templates/app.tmpl` on line 338 (vs a e.g. parameter).

Fixes #720


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;0rL30zwZ&#x2F;272-security-group-inbound-rules"> Security Group inbound rules</a>

MiguelMoll:
Merged via #801

PR #803, opened 2016-06-23
Use current active release to fork from

### LINKED ISSUE #669: Convox run --release should have the proper environment
See the last few comments on #526

### LINKED ISSUE #793: Release created by environment change uses unexpected build
Currently when an environment variable is changed the `ForkRelease()` function forks from the latest release.

https://github.com/convox/rack/blob/master/api/models/app.go#L345-L361

So in the following scenario:

* User creates build but does not promote it
* User updates environment variable and promotes the new release
* App is now running on the unpromoted build

I think the right answer is to make ForkRelease() fork from the currently promoted Release. Anyone see any possible issues with that?

MiguelMoll:
Whenever `ForkRelease` is called it will now use the current release's build. Fixes #793

Also moved `ReleaseLatest` to the provider interface and added the Build ID to the output of `convox releases`.

Also in here, is a fix for `convox run --release`. If no releaseId is provided, `convox run` will use the last promoted release to run a process. Otherwise iterate over the previous revisions (starting with the
latest) looking for the releaseId specified. This makes sure the same environment variables are available. Closes #669

This does not resolve the issue where `convox run` can be used with a
release that hasn't been promoted yet. That will be tackled separately.

### DIFF
--- a/api/models/app.go
+++ b/api/models/app.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/convox/rack/api/helpers"
 	"github.com/convox/rack/api/provider"
+	"github.com/convox/rack/api/structs"
 	"github.com/convox/rack/client"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -335,7 +336,6 @@ func (a *App) UpdateParams(changes map[string]string) error {
 
 func (a *App) Formation() (string, error) {
 	data, err := buildTemplate("app", "app", Manifest{})
-
 	if err != nil {
 		return "", err
 	}
@@ -344,43 +344,25 @@ func (a *App) Formation() (string, error) {
 }
 
 func (a *App) ForkRelease() (*Release, error) {
-	release, err := a.LatestRelease()
-
+	release, err := provider.ReleaseLatest(a.Name)
 	if err != nil {
 		return nil, err
 	}
 
 	if release == nil {
-		r := NewRelease(a.Name)
-		release = &r
+		release = structs.NewRelease(a.Name)
 	}
 
 	release.Id = generateId("R", 10)
 	release.Created = time.Time{}
 
-	return release, nil
-}
-
-// FIXME: Port to provider interface
-func (a *App) LatestRelease() (*Release, error) {
-	releases, err := provider.ReleaseList(a.Name)
-	if err != nil {
-		return nil, err
-	}
-
-	if len(releases) == 0 {
-		return nil, nil
-	}
-
-	r := releases[0]
-
 	return &Release{
-		Id:       r.Id,
-		App:      r.App,
-		Build:    r.Build,
-		Env:      r.Env,
-		Manifest: r.Manifest,
-		Created:  r.Created,
+		Id:       release.Id,
+		App:      release.App,
+		Build:    release.Build,
+		Env:      release.Env,
+		Manifest: release.Manifest,
+		Created:  release.Created,
 	}, nil
 }
 
@@ -388,7 +370,6 @@ func (a *App) ExecAttached(pid, command string, height, width int, rw io.ReadWri
 	var ps Process
 
 	pss, err := ListProcesses(a.Name)
-
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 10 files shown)


nzoschke:
The refactor looks great.

I'd like to start getting a bit more prescriptive about including tests with patches. I think provider interface work is the obvious place to start since the TestProvider is there to offer easier testing strategies. 

Want to include some tests with this PR?

MiguelMoll:
Sure that sounds good to me. Also some output:
```
 convox releases
ID           CREATED             BUILD        STATUS
RLJCAORMDJT  3 minutes from now  BLOTHVWKJAE
RVLQLZVFOWF  3 minutes from now  BLOTHVWKJAE
RUZHEJXDDSV  10 minutes ago      BLOTHVWKJAE  active
RUYIJLIVPWA  10 minutes ago      BMMNUOMDCIE
RFQFSXCFKRF  11 minutes ago      BYCUTSQUQTQ
```

MiguelMoll:
Noah brought up:

> In reviewing this I found a case to think about... Setting 2 environment variables, one at a time.
> If we always fork from the current release, you can't progressively build up a new release.

ddollar:
I think the app's "current environment" is separate from the releases. We can pull the current environment from S3 rather than the latest or current release.

api/structs/release.go
```diff
@@ -19,3 +19,11 @@ func NewRelease(app string) *Release {
 		Id:  generateId("R", 10),
 	}
 }
+
+func (rs Releases) Latest() *Release {
```
ddollar: This could be susceptible to bugs in the future. There's no guarantee that this `Releases` struct is sorted appropriately.

MiguelMoll: Makes sense. Safer to check `Created` to get the latest one.

MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;ilsEvrhf&#x2F;277-release-created-by-environment-change-uses-unexpected-build"> Release created by environment change uses unexpected build</a>

MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Ue5ekEuk&#x2F;280-convox-run-release"> convox run --release</a>

api/models/app.go
```diff
@@ -476,31 +459,80 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
...
 		return err
 	}
 
-	var container *ecs.ContainerDefinition
-	for _, container = range task.TaskDefinition.ContainerDefinitions {
-		if *container.Name == process {
-			break
+	// If no releaseId is provided, use the last promoted release to run this process
+	// otherwise iterate over the previous revisions (starting with the latest) looking for the releaseId specified.
+	if releaseId == "" {
```
nzoschke: looks good to me to retain the current run behavior with no release specified

api/models/app.go
```diff
@@ -476,31 +459,80 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
...
+		for i := len(ts.TaskDefinitionArns) - 1; i >= 0; i-- {
+			t, err := ECS().DescribeTaskDefinition(&ecs.DescribeTaskDefinitionInput{
+				TaskDefinition: ts.TaskDefinitionArns[i],
+			})
+			if err != nil {
+				continue
+			}
+
+			if t == nil {
+				return fmt.Errorf("unable to retrieve task definition")
```
nzoschke: How about making this more user friendly... `fmt.Errorf("Can't find definition for release %s", releaseId)`

nzoschke: ah I'm not quite right here, this would be an internal ECS so I think your error ok

api/models/app.go
```diff
@@ -476,31 +459,80 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
...
+					}
+				}
+
+				if *container.Name == process && releaseMatch {
+					break
+				}
+				container = nil
+			}
+
+			if container != nil {
```
nzoschke: this logic is pretty hard to follow. would a label break be more clear?

https://www.goinggo.net/2013/11/label-breaks-in-go.html

MiguelMoll: I agree. Anything to help make it understandable works for me.

MiguelMoll: Added labels. I think it helps a bit for now.

MiguelMoll:
Merged via #820

Issue #532, opened 2016-04-09
Get just the URL of a service


mattmanning:
It would be nice to have a `--url` option on `convox services info`. That would enable easier scripting and command chaining like `convox env set WHATEVER_URL=$(convox services info whatever --url)`.

mwarkentin:
Another option could be to have a `--json` option, which returns a json blob instead of the plain text version you get normally - then you could parse out whatever fields you want w/ `jq` or some other program.

I suppose this might already be an option if you use the `convox api` command instead?

ddollar:
Yes, use `convox api` to get the JSON. Let's make this `convox services url <service-name>`

PR #825, opened 2016-06-29
Adding services url command to return only the url of a given service

adamenger:
This PR adds `convox services url <name>` and returns only the service.URL for the given service.

For example if I have two services:

```
⮀ convox services
NAME             TYPE    STATUS
fluentd-staging  syslog  running
s3-8346          s3      running
```

The first doesn't have a url and returns:
```
⮀ ~/go/bin/convox services url fluentd-staging
URL does not exist for fluentd-staging
```

The second does have a url and returns:
```
⮀ ~/go/bin/convox services url s3-8346
s3://asdf:asdf@reverb-staging-s3-8346
```

Fixes https://github.com/convox/rack/issues/532

## Release Playbook
- [ ] Rebase against master
- [ ] Release branch ()
- [ ] Pass CI ()
- [ ] Code review
- [ ] Merge into master
- [ ] Release master ()
- [ ] Pass CI ()
- [ ] Update staging rack
- [ ] Publish release
- [ ] Release CLI

### DIFF
--- a/cmd/convox/services.go
+++ b/cmd/convox/services.go
@@ -326,9 +326,9 @@ func cmdServiceURL(c *cli.Context) error {
 	if service.URL == "" {
 		nonexistentURLError := fmt.Errorf("URL does not exist for %s", service.Name)
 		return stdcli.ExitError(nonexistentURLError)
-	} else {
-		fmt.Printf("%s\n", service.URL)
 	}
+
+	fmt.Printf("%s\n", service.URL)
 
 	return nil
 }
--- a/cmd/convox/services.go
+++ b/cmd/convox/services.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"math/rand"
 	"net"
@@ -319,11 +320,13 @@ func cmdServiceURL(c *cli.Context) error {
 	}
 
 	if service.Status == "failed" {
-		fmt.Printf("Reason  %s\n", service.StatusReason)
+		service_failure_error := errors.New(fmt.Sprintf("Service failure for %s", service.StatusReason))
+		return stdcli.ExitError(service_failure_error)
 	}
 
 	if service.URL == "" {
-		fmt.Printf("URL does not exist for %s\n", service.Name)
+		nonexistent_url_error := errors.New(fmt.Sprintf("URL does not exist for %s", service.Name))
+		return stdcli.ExitError(nonexistent_url_error)
 	} else {
 		fmt.Printf("%s\n", service.URL)
 	}
--- a/cmd/convox/services.go
+++ b/cmd/convox/services.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"errors"
 	"fmt"
 	"math/rand"
 	"net"
@@ -320,12 +319,12 @@ func cmdServiceURL(c *cli.Context) error {
 	}
 
 	if service.Status == "failed" {
-		serviceFailureError := errors.New(fmt.Sprintf("Service failure for %s", service.StatusReason))
+		serviceFailureError := fmt.Errorf("Service failure for %s", service.StatusReason)
 		return stdcli.ExitError(serviceFailureError)
 	}
 
 	if service.URL == "" {
-		nonexistentURLError := errors.New(fmt.Sprintf("URL does not exist for %s", service.Name))
+		nonexistentURLError := fmt.Errorf("URL does not exist for %s", service.Name)
 		return stdcli.ExitError(nonexistentURLError)
 	} else {
 		fmt.Printf("%s\n", service.URL)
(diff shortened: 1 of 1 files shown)


cmd/convox/services.go
```diff
@@ -298,6 +305,32 @@ func cmdServiceInfo(c *cli.Context) error {
...
+	}
+
+	if service.Status == "failed" {
+		fmt.Printf("Reason  %s\n", service.StatusReason)
+	}
+
+	if service.URL == "" {
+		fmt.Printf("URL does not exist for %s\n", service.Name)
+	} else {
+    fmt.Printf("%s\n", service.URL)
```
ddollar: Looks like it might need a gofmt

adamenger: Yep totally. Thanks.

cmd/convox/services.go
```diff
@@ -298,6 +305,32 @@ func cmdServiceInfo(c *cli.Context) error {
...
+	if err != nil {
+		return stdcli.ExitError(err)
+	}
+
+	if service.Status == "failed" {
+		fmt.Printf("Reason  %s\n", service.StatusReason)
+	}
+
+	if service.URL == "" {
+		fmt.Printf("URL does not exist for %s\n", service.Name)
```
ddollar: Maybe use `stdcli.ExitError` here so we get the `ERROR:` prefix and non-zero exit code. Since this will probably be used in scripts I think the extra information would be helpful.

cmd/convox/services.go
```diff
@@ -298,6 +305,32 @@ func cmdServiceInfo(c *cli.Context) error {
...
+
+	name := c.Args()[0]
+
+	service, err := rackClient(c).GetService(name)
+	if err != nil {
+		return stdcli.ExitError(err)
+	}
+
+	if service.Status == "failed" {
+		fmt.Printf("Reason  %s\n", service.StatusReason)
```
ddollar: Probably make this an error of some sort with `stdcli.ExitError()` since there won't be a valid URL to show.

ddollar:
This looks great, thank you :) I've been wanting this command myself for a while. Made a couple of suggestions about ending in non-zero exit codes when a URL is not available.

adamenger:
Thank you for the feedback. I've submitted a patch to address those issues -- this definitely creates a better experience. Question: Is it okay to import "errors" here? I was following the patterns laid out in `cmdServiceInfo` which didn't return stdcli.ExitError() for a service status that resulted in a failure so that's why my initial implementation replicated that pattern.

adamenger:
Nevermind, hound reminded me of being able to use `fmt.Errorf` instead of `errors.New()`

cmd/convox/services.go
```diff
@@ -298,6 +305,34 @@ func cmdServiceInfo(c *cli.Context) error {
...
+
+	if service.Status == "failed" {
+		serviceFailureError := fmt.Errorf("Service failure for %s", service.StatusReason)
+		return stdcli.ExitError(serviceFailureError)
+	}
+
+	if service.URL == "" {
+		nonexistentURLError := fmt.Errorf("URL does not exist for %s", service.Name)
+		return stdcli.ExitError(nonexistentURLError)
+	} else {
```
adamenger: Is this something I should do here? @ddollar

ddollar: Yep, makes sense to me. The `if`s with a return inside are basically guard statements.

cmd/convox/services.go
```diff
@@ -298,6 +305,34 @@ func cmdServiceInfo(c *cli.Context) error {
...
+
+	name := c.Args()[0]
+
+	service, err := rackClient(c).GetService(name)
+	if err != nil {
+		return stdcli.ExitError(err)
+	}
+
+	if service.Status == "failed" {
+		serviceFailureError := fmt.Errorf("Service failure for %s", service.StatusReason)
```
ddollar: You can probably even one-liner these with:
`return stdcli.ExitError(fmt.Errorf(...))`

ddollar:
Looks good. I think this is a bit different than `services info` as it seems designed to be used in some sort of scripting. To me this means no formatting or preambles, making sure we are using stdin and stderr correctly, and good exit codes. Your changes are looking really good.

cmd/convox/services.go
adamenger: Oh I get it now. Was worried it would hit a point where the url was not defined but you're right, they're guarded by the prior returns.

Thanks.

cmd/convox/services.go
adamenger: Yep that's how I had that originally but I did was thinking of others who might wander in here. If you prefer the one liner, we can do that.

MiguelMoll:
Thanks for this 👍 . The change was merged via #844

PR #847, opened 2016-07-06
use aws cli to generate credentials if available

ddollar:
if the aws cli is installed and configured this will now try to read the credentials from your current profile if you haven't specified any as flags

### DIFF
--- a/cmd/convox/install.go
+++ b/cmd/convox/install.go
@@ -46,7 +46,7 @@ To generate a new set of AWS credentials go to:
 https://docs.convox.com/creating-an-iam-user`
 
 var (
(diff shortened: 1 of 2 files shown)


mwarkentin:
If you do end up doing something like this, we need to be able to pass the `--profile` option to our AWS CLI to specify the AWS account to be used (we have separate prod / dev accounts).

ddollar:
@mwarkentin you can also set the `AWS_DEFAULT_PROFILE` environment variable to change the current default

mwarkentin:
Oh true, that'd work. 👍

mwarkentin:
@ddollar Just realized something similar for `convox uninstall` might be useful as well? Easy to port there as well?

Issue #841, opened 2016-07-05
Verbose mode for convox sync process


mwarkentin:
We have an issue on one of our projects where files are continually synced in a loop - every second or so you get a new message about "downloaded 86 files". It'd be nice if there was an option which would cause that message to include the exact files being sync'd, so that we could track down what's going on easily.

PR #853, opened 2016-07-08
convox start sync working with single files

ddollar:


mwarkentin:
Any chance you can perhaps implement a verbose sync mode while you're in here for #841?

ddollar:
@mwarkentin good call, added

PR #858, opened 2016-07-11
Display the upload progress of the artifcat during a build

### LINKED ISSUE #543: CLI upload progress
When running `convox build` or `convox deploy` it would be nice to see live data upload progress.

MiguelMoll:
`convox build` needs better feedback during an upload. The goal here is to show the user an upload is making progress.

The feedback will include how much data has been uploaded and the total size of the file being uploaded.

Example output:
```
> convox build
Creating tarball... OK
Uploading... 58.85 KB / 58.85 KB  100.00 % 0
Getting build logs...
RUNNING: tar xz
RUNNING: docker pull httpd
Using default tag: latest
```

Fixes #543

### DIFF
--- a/client/builds.go
+++ b/client/builds.go
@@ -69,13 +69,13 @@ func (c *Client) CreateBuildIndex(app string, index Index, cache bool, manifest 
 	return &build, nil
 }
 
-// CreateBuildSource will create a new build from source. If progress of the uploaded is needed, see CreateBuildSourceP
+// CreateBuildSource will create a new build from source. If progress of the uploaded is needed, see CreateBuildSourceProgress
 func (c *Client) CreateBuildSource(app string, source []byte, cache bool, manifest string, description string) (*Build, error) {
-	return c.CreateBuildSourceP(app, source, cache, manifest, description, nil)
+	return c.CreateBuildSourceProgress(app, source, cache, manifest, description, nil)
 }
 
-// CreateBuildSourceP will create a new build from source with an optional callback to provide progress of the source being uploaded.
-func (c *Client) CreateBuildSourceP(app string, source []byte, cache bool, manifest string, description string, progressCallback func(s string)) (*Build, error) {
+// CreateBuildSourceProgress will create a new build from source with an optional callback to provide progress of the source being uploaded.
+func (c *Client) CreateBuildSourceProgress(app string, source []byte, cache bool, manifest string, description string, progressCallback func(s string)) (*Build, error) {
 	var build Build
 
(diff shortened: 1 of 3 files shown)


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;H42tPkAB&#x2F;332-upload-progress-during-builds"> Upload progress during builds</a>

client/builds.go
```diff
@@ -70,6 +70,10 @@ func (c *Client) CreateBuildIndex(app string, index Index, cache bool, manifest
 }
 
 func (c *Client) CreateBuildSource(app string, source []byte, cache bool, manifest string, description string) (*Build, error) {
+	return c.CreateBuildSourceP(app, source, cache, manifest, description, nil)
+}
+
+func (c *Client) CreateBuildSourceP(app string, source []byte, cache bool, manifest string, description string, progressCallback func(s string)) (*Build, error) {
```
ddollar: let's use a slightly more descriptive name as well if we're going to export this

mwarkentin:
@MiguelMoll Do you have a demo of what this looks like? 😄

MiguelMoll:
How did I forget?! I'll add it to the first post. Thanks!

client/builds.go
MiguelMoll: I thought about that but didn't want the name to get to unwieldy. I think a a doc comment will explain what it's for while keeping name in check.

cmd/convox/builds.go
```diff
@@ -623,6 +616,7 @@ func createTarball(base string) ([]byte, error) {
 }
 
 func finishBuild(c *cli.Context, app string, build *client.Build) (string, error) {
+	fmt.Println("Streaming build logs...")
```
ddollar: Let's remove this extra print

client/builds.go
ddollar: Let's go ahead and change the name on this so we don't have to look up docs to know what it means

MiguelMoll: Sounds good to me. Let me know if there any thoughts on `CreateBuildSourceProgress`.

cmd/convox/builds.go
MiguelMoll: Will do. I liked the print entry so we could know where in build process the CLI was. But I'm OK either way.

ddollar:
Looks like we have a few test failures now

MiguelMoll:
Missed a method call to rename :(

ddollar:
Merged in #882

PR #860, opened 2016-07-11
Logic to support `convox run` with releases that have yet to be promoted

### LINKED ISSUE #838: Regression -- Cannot run one-off processes with --release specified on processes with --count=-1
## Setup
```bash
$ convox scale -a <app>
NAME       DESIRED  RUNNING  CPU  MEMORY
<process>  -1        0        0    256
```

## Before
```bash
$ convox run <process> <command> -a <app> --release <release_id>
... runs the command ...
```
## After
```bash
$ convox run <process> <command> -a <app> --release <release_id>
ERROR: unable to find contaier for frontend
```

## Release that broke this behaviour
https://github.com/convox/rack/releases/tag/20160629165916

MiguelMoll:
The idea here is to combine settings from the most recently *promoted*
release and the most recently *created* release. With the most recently created release taking precedence.

Assumptions:
- An app would have to be promoted at least once.
- Using the most recently promoted release as a basis to build upon with the most recently created (but not promoted) release.

TODO:

- [x] Add support for running detached processes with a unpromoted release.

Fixes #838

### DIFF
--- a/api/models/app.go
+++ b/api/models/app.go
@@ -517,10 +517,12 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
 		}
 	}
 
-	ea := make([]string, 0)
+	var rawEnvs []string
 	for _, env := range container.Environment {
-		ea = append(ea, fmt.Sprintf("%s=%s", *env.Name, *env.Value))
+		rawEnvs = append(rawEnvs, fmt.Sprintf("%s=%s", *env.Name, *env.Value))
 	}
+	containerEnvs := structs.Environment{}
+	containerEnvs.LoadRaw(strings.Join(rawEnvs, "\n"))
 
 	// Update any environment variables that might be part of the unpromoted release.
 	if unpromotedRelease {
@@ -624,7 +626,7 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
 			AttachStdin:  true,
 			AttachStdout: true,
 			AttachStderr: true,
-			Env:          ea,
+			Env:          containerEnvs.List(),
 			OpenStdin:    true,
 			Tty:          true,
 			Cmd:          []string{"sh", "-c", command},
--- a/api/models/app.go
+++ b/api/models/app.go
@@ -459,7 +459,8 @@ func (a *App) ExecAttached(pid, command string, height, width int, rw io.ReadWri
 	return nil
 }
 
-func (a *App) RunAttached(process, command, releaseId string, height, width int, rw io.ReadWriter) error {
+// RunAttached runs a command in the foreground (e.g blocking) and writing the output from said command to rw.
+func (a *App) RunAttached(process, command, releaseID string, height, width int, rw io.ReadWriter) error {
 	//TODO: A lot of logic in here should be moved to the provider interface.
 
 	resources, err := a.Resources()
@@ -467,11 +468,11 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
 		return err
 	}
 
-	if releaseId == "" {
-		releaseId = a.Release
+	if releaseID == "" {
+		releaseID = a.Release
 	}
 
-	release, err := GetRelease(a.Name, releaseId)
+	release, err := GetRelease(a.Name, releaseID)
 	if err != nil {
 		return err
 	}
@@ -498,11 +499,11 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
 	}
 
 	// If the release ID provided does not equal the active one, some logic is needed to determine the next steps.
-	// - For a previous release, we iterate over the previous TaskDefinition revisions (starting with the latest) looking for the releaseId specified.
+	// - For a previous release, we iterate over the previous TaskDefinition revisions (starting with the latest) looking for the releaseID specified.
 	// - If the release has yet to be promoted, we use the most recent TaskDefinition with the provided release's environment.
-	if releaseId != a.Release {
+	if releaseID != a.Release {
 
-		_, releaseContainer, err := findAppDefinitions(process, releaseId, *task.TaskDefinition.Family, 20)
+		_, releaseContainer, err := findAppDefinitions(process, releaseID, *task.TaskDefinition.Family, 20)
 		if err != nil {
 			return nil
 		}
@@ -692,7 +693,8 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
 	return nil
 }
 
-func (a *App) RunDetached(process, command, releaseId string) error {
+// RunDetached runs a command in the background (e.g. non-blocking).
+func (a *App) RunDetached(process, command, releaseID string) error {
 	resources, err := a.Resources()
 	if err != nil {
 		return err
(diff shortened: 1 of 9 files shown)


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;bFCGG3GT&#x2F;313-fix-convox-run-work-with-a-unpromoted-release"> Fix convox run work with a &quot;unpromoted&quot; release</a>

api/models/app.go
```diff
@@ -520,30 +537,53 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
...
 				break
 			}
 		}
-	}
 
-	if container == nil {
-		return fmt.Errorf("unable to find contaier for %s", process)
+		// If container is nil, the release most likely hasn't been promoted and thus no TaskDefinition for it.
+		if releaseContainer == nil {
+			fmt.Printf("Unable to find contaier for %s. Basing container off of most recent release: %s.\n", process, a.Release)
```
nzoschke: typo on "contaier"

api/models/app.go
```diff
@@ -474,18 +486,22 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
...
+			break
 		}
+	}
 
-	} else {
+	if container == nil {
+		return fmt.Errorf("unable to find container for %s", process)
+	}
+
+	// If the release ID provided does not equal the active one, some logic is needed to determine the next steps.
```
nzoschke: This comment is very good! The logic described makes sense to me.

api/models/app.go
```diff
@@ -520,30 +537,53 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
...
 			}
 		}
-	}
 
-	if container == nil {
-		return fmt.Errorf("unable to find contaier for %s", process)
+		// If container is nil, the release most likely hasn't been promoted and thus no TaskDefinition for it.
+		if releaseContainer == nil {
+			fmt.Printf("Unable to find contaier for %s. Basing container off of most recent release: %s.\n", process, a.Release)
+			unpromotedRelease = true
```
nzoschke: is there a way to determine this before looping through all the task definitions?

api/models/app.go
```diff
@@ -474,18 +486,22 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
...
+	if container == nil {
+		return fmt.Errorf("unable to find container for %s", process)
+	}
+
+	// If the release ID provided does not equal the active one, some logic is needed to determine the next steps.
+	// - For a previous release, we iterate over the previous TaskDefinition revisions (starting with the latest) looking for the releaseId specified.
+	// - If the release has yet to be promoted, we use the most recent TaskDefinition with the provided release's environment.
+	if releaseId != a.Release {
+
+		var releaseContainer *ecs.ContainerDefinition
```
nzoschke: some apps could have hundreds of TaskDefinitions from previous deploys. Does this take too long to search? Will the search hit rate limits?

api/models/app.go
```diff
@@ -520,30 +537,53 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
...
 	}
 
-	release, err := GetRelease(a.Name, releaseId)
-	if err != nil {
-		return err
+	// Update any environment variables that might be part of the unpromoted release.
+	if unpromotedRelease {
+		releaseEnv := strings.Split(release.Env, "\n")
+
+	ReleaseEnvCheck:
```
nzoschke: This loop is hard for me to follow. From your past comments I was expecting this to be straighforward. The unpromoted release env is applied over the current release env. Finally RELEASE would be explicitly set to the unpromoted release id.

api/models/helpers.go
```diff
@@ -389,3 +389,12 @@ func UpperName(name string) string {
 
 	return us
 }
+
+func StringInSlice(str string, slice []string) bool {
```
nzoschke: Are these needed? I don't see them referenced elsewhere in the patch

api/models/app.go
MiguelMoll: I was thinking the same before, perhaps with using the dynamo table as a reference but nothing stood out at me. 

Thought about using the created timestamp and comparing that to the active timestamp. Wasn't sure if it was reliable enough.

api/models/app.go
MiguelMoll: That's the gist of the logic, yes. I'll take another look at it to see what else can be improved. It was probably a little hack keeping the environment variables as strings instead of converting them to real maps to make it easier to work with.

api/models/helpers.go
MiguelMoll: I used them during initial design phase and thought they were nifty to have around. But think "less code is more" wins here.

api/models/app.go
MiguelMoll: This can indeed take a while, especially if a user calls `convox run` with a very old release. The go-ecs API doesn't provide a "paging" functionality like it's other calls for describing task definitions (which is understandable since we want to describe one revision).

Two options top of mind:

1. Sleep for a second after _n_ number of task definitions. (arbitrarily start with 15?).
2. Limit `convox run` to use only one of the last _n_ task definitions. (again 15 might work).

These options help with rate limits. There's still the issue of a run taking too long. Especially if it's towards the bottom of the list. More recent releases would be towards the top of the list and are iterated over first.

I'm leaning towards option 2. As is, to determine if a release hasn't been promoted, we have to iterate the task definitions revisions. If it's not found there then it's an "unpromoted" release. It would be a bad user experience having to wait that iteration if there are hundreds of revisions to iterate first.

nzoschke: This reminds me of the recent builds cleanup change. We now only retain the last 50 builds and only list 20 by default.  (We also have a rate limit problem in the build list/cleanup here, which is why its on my mind for this review).

So +1 for limiting it to N most recent releases. We only return 20 in the ReleaseList API so that seems like a reasonable number.

nzoschke:
The strategy of creating a new TaskDefinition is cool. Will this TaskDefinition be used by subsequent runs?

nzoschke:
Will the temp TaskDefinition cause problems if/when the release is eventually promoted?

MiguelMoll:
Good question! I had planned to clean up the temp task (that is mark it as "inactive") but notice that  subsequent runs will use the same task definition. Saw it as an extra bonus so I kept it as is.

If/when the release is promoted, there won't be any issues as CloudFormation will create a new revision to the task definition and the release will use that new one. So would any runs that use that release.

ddollar:
Let's correct the Hound violations then should be ready to merge

MiguelMoll:
The `rawEnvs` change makes sense to me but I used `releaseId` (lowercase d) as that's the same naming scheme used throughout the code. I was favoring consistency. 

Thoughts on renaming existing ones or keeping it as is?

ddollar:
Feel free to rename them all. Let's try to keep the linter happy as much as possible.

MiguelMoll:
I like that stance. Will use it going forward. Should be good to go.

ddollar:
Merged in #891

PR #869, opened 2016-07-15
Adding FluentD service

adamenger:
This patch adds a service for shipping logs to a fluentd collector. You will need to fork [this repo](https://github.com/reverbdotcom/convox-fluentd), build and deploy the fluentd.zip to the convox s3 bucket in order for it to work. 

Setting up a FluentD service:
```
⮀convox services create fluentd --name fluentd-native --url tcp://10.0.0.234:24224
Creating fluentd-native (fluentd: name="fluentd-native" url="tcp://10.0.0.234:24224")...
CREATING
```

Linking an app:
```
⮀convox services link fluentd-native --app hendrix
Linked fluentd-native to hendrix
```

Unlinking:
```
⮀ ~/go/bin/convox services unlink fluentd-native --app hendrix
Unlinked fluentd-native from hendrix
```

Deleting the service:
```
⮀ ~/go/bin/convox services delete fluentd-native
Deleting fluentd-native... DELETING
```

I'm having one issue and that's with updating the service:
```
⮀ ~/go/bin/convox services update fluentd-native --url tcp://10.0.0.235:24224
Updating fluentd-native (url="tcp://10.0.0.235:24224")...ERROR: Asset templates/service/fluentd.tmpl not found
```

I've followed all of the patterns that I could find for the syslog service, and it appears that service type has the same issue when updating:
```
⮀ convox services update syslog-4305 --url tcp://1.2.5.4:4150
Updating syslog-4305 (url="tcp://1.2.5.4:4150")...ERROR: Asset templates/service/syslog.tmpl not found
```

**Update**: The syslog update controller has been patched here(https://github.com/convox/rack/pull/870), so if this is accepted I can update this PR to follow this pattern as well.

Please let me know if there are any questions.

PR #870, opened 2016-07-15
Fixing updates for logger services

adamenger:
Currently the syslog service fails with `Updating syslog-4305 (url="tcp://1.2.5.4:4150")...ERROR: Asset templates/service/syslog.tmpl not found` when running `convox services update syslog-4305 --url tcp://1.2.5.4:4150`

This patch adds that template to the proper location, adds a `service-logger` model and an UpdateLogger method. Feedback greatly appreciated.

When/if https://github.com/convox/rack/pull/869 is merged, updating that service to follow this pattern will be straightforward.


ddollar:
Merged in #882

PR #873, opened 2016-07-18
Add django support to init

ianmalott:
When `convox init` is run in a directory with a `manage.py`, assume it's a django project and generate helpful starting points for `.dockerignore`, `Dockerfile`, and `docker-compose.yml`.


ddollar:
Merged in #882

PR #874, opened 2016-07-18
Gather build logs in separate goroutine.

MiguelMoll:
Sometimes, `client.Logs()` would block. That has to be looked into. But in the mean time, this alternate approach won't lock up the CLI during a build/deploy.

This PR is more like a RFC on the approach taken. Definitely still a WIP. For example, if the Logs() call does block in the goroutine, that goroutine will stay alive in the background until it's cleaned up (docker timeout perhaps?).

### DIFF
--- a/api/controllers/builds.go
+++ b/api/controllers/builds.go
@@ -295,7 +295,7 @@ func BuildLogs(ws *websocket.Conn) *httperr.Error {
 	go keepAlive(ws, quit)
 	go func() {
 		fmt.Println("############################ BEFORE client.Logs")
-		err := client.Logs(docker.LogsOptions{
+		e := client.Logs(docker.LogsOptions{
 			Container:    fmt.Sprintf("build-%s", build),
 			Follow:       true,
 			Stdout:       true,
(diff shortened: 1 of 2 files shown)


ddollar:
You could build a `time.After` channel and add that to your select for timeout/cleanup.

MiguelMoll:
Definitely. I was thinking more about what a reasonable timeout value would look like. 15 minutes? 30?

MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;RqWQO4WB&#x2F;335-hanging-builds"> Hanging builds</a>

nzoschke:
My first guess is that a global build timeout should be closer to 45m or 60m to start. We have some users with apps with lots of python dependencies. If there is a Docker cache miss this takes a long time to rebuild.

ddollar:
Yep, I would say long. An hour seems right to me. I'd even be fine with something like 4h just as a safety check for runaway processes but letting people do weird silly things if they want.

ddollar:
Merged in #882

PR #885, opened 2016-07-20
[WIP] One manifest to rule them all

awsmsrc:


api/models/app.go
```diff
@@ -546,13 +554,13 @@ func (a *App) RunAttached(process, command, releaseId string, height, width int,
 		return err
 	}
 
-	manifest, err := LoadManifest(release.Manifest, a)
+	manifest, err := manifest.Load([]byte(release.Manifest))
```
awsmsrc: why are we loading this when it is passed in?

api/models/templates/app.tmpl
```diff
@@ -323,14 +324,15 @@
       },
       "{{ upper $balancer.ProcessName }}Port{{ .Balancer }}BalancerName": {
         "Condition": "Enabled{{ upper $balancer.ProcessName }}",
-        "Value": {{ $balancer.LoadBalancerName }}
+        "Value": {{ $balancer.LoadBalancerName true $app.Name }}
```
awsmsrc: expicitly setting this to true may not be correct

manifest/balancer.go
```diff
@@ -0,0 +1,220 @@
...
+		}
+		// ELB name must be 32 chars or less
+		if len(prefix) > 32-len(suffix) {
+			prefix = prefix[:32-len(suffix)]
+		}
+		return template.HTML(`"` + prefix + suffix + `"`)
+	}
+
+	// Unbound apps use legacy StackName or StackName-ProcessName format
+	//TODO
```
awsmsrc: What does this mean??

PR #886, opened 2016-07-20
warn about multiple missing envs if more then one

awsmsrc:
cc @ddollar

### DIFF
--- a/manifest/run.go
+++ b/manifest/run.go
@@ -81,13 +81,18 @@ func (r *Run) Start() error {
 			links[key] = true
 		}
 
+		missingEnv := []string{}
 		for key, val := range s.Environment {
 			eok := val != ""
 			_, exok := existing[key]
 			_, lok := links[key]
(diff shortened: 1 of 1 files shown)


manifest/run.go
```diff
@@ -81,14 +81,19 @@ func (r *Run) Start() error {
...
 			_, exok := existing[key]
 			_, lok := links[key]
 			if !eok && !exok && !lok {
-				return fmt.Errorf("env expected: %s", key)
+				missingEnv = append(missingEnv, key)
 			}
 		}
+
+		if len(missingEnv) > 0 {
+			return fmt.Errorf("env expected: %s", strings.Join(missingEnv, ", "))
```
mwarkentin: Maybe `env(s)`?

awsmsrc: some portion of environment is missing vs env variables are missing. 
English is a funny old language :)

Happy to go with the majority here as I have no preference

PR #890, opened 2016-07-20
Pad the uploading progress message to clear some text

MiguelMoll:
This approach seems the most pragmatic. Other option is to mess around with ASCII/Unicode escape characters to clear the end of the line:

```
fmt.Printf("\r\033[KUploading... %s       ", strings.TrimSpace(s))
```
It works but could run into issues with a different  character encoding.


ddollar:
Merged in #891

Issue #897, opened 2016-07-21
Files syncing to unexpected  dir in container


ianmalott:
When running `convox start`, files created in `backend/` get synced to the container at `/'/usr/src/app'/backend` instead of `/usr/src/app/backend`.

---

Project structure
```
.
├── backend
│   ├── Dockerfile
│   ├── foo
│   ├── package.json
│   └── server.js
├── docker-compose.yml
└── frontend
    ├── Dockerfile
    ├── package.json
    └── server.js
```

docker-compose.yml
```
version: '2'
services:
  backend:
    build: ./backend
    ports:
      - 8001:8000

  frontend:
    build: ./frontend
    ports:
      - 80:8000
```

Dockerfile for both backend/ and frontend/
```
FROM nodesource/node:6.2

ADD package.json package.json
RUN npm install --loglevel error
ADD . .

EXPOSE 8000

CMD ["npm", "start"]
```

awsmsrc:
This is fixed

PR #906, opened 2016-07-22
WIP - BUG 897

awsmsrc:
So one issue here was that there were ' characters being passed as part of the path.
This PR already fixes that however now you can see that we need to run the path the sync should being ran from as we are getting extra nesting.

cc @ianmalott


mwarkentin:
Ref: https://github.com/convox/rack/issues/897

PR #908, opened 2016-07-22
WIP - 827: Parse `tty` param

### LINKED ISSUE #827: docker-compose run equivalent?
- [docker-compose run docs](https://docs.docker.com/compose/reference/run/)

Our use case is running an interactive debugger in one of our web or rpc processes - which doesn’t work with `convox start` as it’s not an interactive terminal. We’ve got a hacky way to handle it for our web processes detailed below), however this doesn’t work for our rpc processes as we would have to modify our calling code as well to point to the new port (would probably get laughed out of the room if I asked devs to do this 😄).

What we’d like is something like (ignore the terrible naming, there’s already a `convox run` command..) - `convox runlocal rpc —service-ports python manage.py runserver 0.0.0.0:8000`

This would run the command above inside the `rpc` process, launch all of our other required containers in the background, and map the ports.
### Hacky McHackface

docker-compose.yml

``` yaml
web:
  build: .
  command: /app/bin/web
  ports:
    - 10008:80
    - 8008:8000  # this is the extra port for running the debug process interactively on
```

Makefile

```
start: ## Start application
    @convox start
debug: ## Launch a interactive runserver process
    @open https://localhost:8008  # Launch the browser to the correct port, so that devs don't have to remember which port is which
    docker exec -it notifications-web python manage.py runserver 0.0.0.0:8000  # Launch the dev server w/ debugger in interactive terminal
```

To use the interactive debugger, I would run `make start` in one tab, then `make debug` in another - so I'm running my debug process _in addition to the normal web process_ on a different port.

Ideally (as far as my short amount of thinking goes), what I'd do instead:
- Remove the `8008` port definition from my `docker-compose.yml`
- Replace `docker exec -it notifications-web python manage.py runserver 0.0.0.0:8000` with `convox runlocal web —service-ports` (I'm not sure if I could do this to run my default command, or if I'd need to specify)
  - This would launch an interactive terminal with my web process running on it's usual port, which means I could continue to use my existing domain mapping, etc.

mwarkentin:
Fixes #827 

Not working yet..

```
$ convox start

...

build  │ Successfully built 5793a497a8f7
web    │ running: docker run -i --rm --name pronetwork-web --tty -e DEBUG=1 -e GEOIP_PATH=/usr/local/share/GeoIP -e UWSGI_PY_AUTORELOAD=1 -e COMPRESS_OFFLINE=0 -e CONVOX=1 -e CHUNNEL_URI_IDENTITY=33.33.33.100:3500 -p 80 convox-a088a6d827 sh -c /app/bin/web
web    │ the input device is not a TTY
```

### DIFF
--- a/manifest/process.go
+++ b/manifest/process.go
@@ -19,6 +19,10 @@ func NewProcess(app string, s Service) Process {
 	args = append(args, "--rm")
 	args = append(args, "--name", name)
 
+	if s.Tty {
+		args = append(args, "--tty")
+	}
+
 	if s.Entrypoint != "" {
 		args = append(args, "--entrypoint", s.Entrypoint)
 	}
--- a/manifest/service.go
+++ b/manifest/service.go
@@ -26,6 +26,7 @@ type Service struct {
 	Networks    []string    `yaml:"networks,omitempty"`
 	Ports       Ports       `yaml:"ports,omitempty"`
 	Privileged  bool        `yaml:"privileged,omitempty"`
+	Tty         bool        `yaml:"tty,omitempty"`
 	Volumes     []string    `yaml:"volumes,omitempty"`
 }
 


mwarkentin:
@ddollar Did you make any other tweaks besides the args to `docker` when you gave me the `convox` bin for interactive debugger testing?

ddollar:
Going to close this for now but will solve #827 with a pty-based solution

PR #910, opened 2016-07-23
Adding support for relative paths when parsing docker-compose.yml

adamenger:
In its current form if I have a docker-compose.yml that looks like:
```
version: "2"
services:
  postgres:
    image: postgres:9.5
    volumes:
      - ./dbdata:/var/lib/postgresql/data
```

I will get this error:
```
postgres │ running: docker run -i --rm --name lessons-postgres -v ./dbdata:/var/lib/postgresql/data convox-5ec422cf51
postgres │ docker: Error response from daemon: create ./dbdata: "./dbdata" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]" are allowed.
```

Docker expects absolute paths when resolving volume paths. This patch adds support for relative paths when mounting volumes by testing if the path is absolute and if it's not, we resolve the path using `Abs` so we can proceed booting the container cleanly:

```
postgres │ running: docker run -i --rm --name lessons-postgres -v /Users/adam/dev/lessons/dbdata:/var/lib/postgresql/data convox-5ec422cf51
```

### DIFF
--- a/manifest/process.go
+++ b/manifest/process.go
@@ -1,6 +1,10 @@
 package manifest
 
-import "fmt"
+import (
+	"fmt"
+	"path/filepath"
+	"strings"
+)
 
 type Process struct {
 	Name string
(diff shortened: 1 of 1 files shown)


adamenger:
On second thought this is seemingly causing issues with sync. Need to rethink this.

Issue #905, opened 2016-07-22
Upgrading to 20160720211125 errors out


adamenger:
Doing a `convox rack update` on my rack results in this error in CloudFormation:

```
    ELB cannot be attached to multiple subnets in the same AZ.
```

frankleng:
+1

ddollar:
I reverted some changes around the exisiting-VPC stuff and cut a new release. Please try again and let me know if you still see this.

frankleng:
@ddollar thanks. confirm fixed

PR #911, opened 2016-07-23
Fix existing VPC installation

beedub:
The EFS volume stuff didn't have its VPC conditional, which was making installations fail.

The lingering issue is https://github.com/convox/rack/issues/905. Just looking at the changes, I can't obviously tell why it resulted in the error above.


beedub:
@ddollar update to allow only an existing VPC, no existing subnets

beedub:
Obviate by https://github.com/convox/rack/pull/946

PR #912, opened 2016-07-24
allow customization of ELB IdleTimeout

bobzoller:
We noticed a significant number of 504s being served by our Convox-managed ELBs.  Turns out this is the well-known issue that manifests when the ELB `IdleTimeout` is greater than the backend's socket timeout.  For our nodejs-based apps, the default socket timeout is 2 minutes.  Though ELBs usually default to a 60 second timeout, Convox sets it to 3600 seconds.

>  To ensure that the load balancer is responsible for closing the connections to your back-end instance, make sure that the value you set for the keep-alive time is greater than the idle timeout setting on your load balancer.

http://docs.aws.amazon.com/ElasticLoadBalancing/latest/DeveloperGuide/config-idle-timeout.html

A higher timeout typically means better performance, but it seems like this parameter should be tunable, as different situations call for different values.  To that end, this PR allows for a new docker label `convox.idle_timeout`, which must be a number.  If it is not provided, the existing default of `3600` will be used.

Feedback welcome.  I saw two existing patterns to build on: the custom health check and the `SecurityGroup` parameter.  I chose the former because I prefer the declarative docker label style.


bobzoller:
rebased

api/models/manifest.go
```diff
@@ -297,6 +297,23 @@ func (mb ManifestBalancer) HealthInterval() (string, error) {
 	return "5", nil
 }
 
+// IdleTimeout The amount of time to allow the balancer to keep idle connections open. This should be
+// greater than the keep-alive timeout on your back-end, so that the balancer is responsible for
+// closing connections
+func (mb ManifestBalancer) IdleTimeout() (string, error) {
+	if timeout := mb.Entry.Label("convox.idle_timeout"); timeout != "" {
```
ddollar: Let's call this `convox.idle.timeout`

bobzoller: :+1: changed

bobzoller:
@ddollar back to you

ddollar:
We recently made a fairly large refactor to the manifest handling stuff so this will unfortunately need to be rebased. Let me know if it is unclear how to proceed.

bobzoller:
@ddollar rebased, but I wasn't sure what to do with the test I added (in `api/models/fixtures/idle_timeout.{yml,json}`, with a reference in the defunct `api/models/manifest_test.go`)

bobzoller:
@ddollar figured out how to test the parsing of `convox.idle.timeout`, but feels like I should be testing the CF app template too.  LMK.

ddollar:
Carrying forward in #980, thank you!

PR #915, opened 2016-07-25
Bug fixes for `convox run`

MiguelMoll:
Two `convox run` issues are fixed here:

- The CLI returns a proper error message when an app is not found with `--detach`.
- Fixed a silent error when specifying a release that doesn't have any environment settings. (Thanks to @beedub for the debug help!)

### DIFF
--- a/api/structs/environment.go
+++ b/api/structs/environment.go
@@ -56,6 +56,10 @@ func (e Environment) Raw() string {
 
 // LoadRaw reads a raw string (key/values separated by a newline) to load environment variables
 func (e Environment) LoadRaw(raw string) {
+	if raw == "" {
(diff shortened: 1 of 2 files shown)


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;v5cdXuwj&#x2F;390-bug-fix-for-convox-run"> bug fix for convox run</a>

beedub:
shouldn't the title read "run command handles error if env is empty"?

MiguelMoll:
I created this PR before we found out the root cause of the env issue. The title is in reference to a bug I found along the way. I'll update it. :+1:

PR #920, opened 2016-07-26
Consolidate app params

beedub:
We exceeded the limit of app params (63 > 60). This PR attempts to reduce the number of cf params generated.

1. Remove the health params altogether
1. Conditionally include Proxy and Secure params only if they're specifically set to true in their corresponding label

So in total, each balancer would have between 2-4 fewer cf params generated for it


api/models/templates/app.tmpl
```diff
@@ -287,24 +249,28 @@
         "Default" : "{{ index $balancer.Randoms .Balancer }}",
         "Description" : ""
       },
+      {{ if .Secure }}
```
ddollar: I think these conditionals will be problematic if the parameter is not there at all. We should either fully de-parameterize or make them unconditional I think.

api/models/templates/app.tmpl
```diff
@@ -386,68 +352,59 @@
...
-                          {
-                            "Fn::If": [
-                              "Balancer{{ upper $balancer.ProcessName }}Port{{ .HealthPort }}Secure",
-                              {"Fn::FindInMap": ["PortProtocol", "http", "SecureInstanceProtocol"]},
-                              {"Fn::FindInMap": ["PortProtocol", "http", "InstanceProtocol"]}
-                            ]
-                          }
-                        ]
-                      },
+                      {{ if eq .HealthPath "" }}
```
ddollar: what is the difference in this branch?

api/models/templates/app.tmpl
beedub: Yeah you're right. I'll fully de-param

beedub: updated

api/models/templates/app.tmpl
beedub: difference is port protocol.

healthpath set = tcp
healthpath not set = http

PR #921, opened 2016-07-26
Consolidate app params #2

beedub:
(This is a redux of https://github.com/convox/rack/pull/920 based on comments)

This PR reduces the number of app params generated by `2 * <LB COUNT> + 2 * <LB PORT MAPPING COUNT>` by removing the base lb health params, and per-port proxy and secure params.

@ddollar


ddollar:
Closing in favor of #956 we can work this back in in pieces

PR #922, opened 2016-07-26
Add deployment min/max options

beedub:
This will allow each process to specify the min and max number of
containers to run during deployment


api/models/manifest.go
```diff
@@ -537,6 +537,23 @@ func (e ManifestEntry) MountableVolumes() []MountableVolume {
 	return volumes
 }
 
+// DeploymentMinimumPercent returns the min percent of containers that are allowed during deployment
+func (me ManifestEntry) DeploymentMinimumPercent() string {
+	if min := me.Label("convox.deployment.min_percent"); min != "" {
```
ddollar: Let's use `convox.deployment.minimum` for this

beedub: updated

ddollar:
This looks good but similar concern as the others with removing the params. We should add a check to make sure that the labels match the old params if they were set.

api/models/release.go
```diff
@@ -194,6 +194,19 @@ func (r *Release) Promote() error {
 			}
 		}
 
+		// Deployment min/max per service was moved from CF param to manifest label
+		// Verify that deployment min and max are backwards compatible
+		deploymentOpts := []string{ "minimum", "maximum" }
+		for opt := range deploymentOpts {
+			label := fmt.Sprintf("convox.deployment.%s", opt)
+			if labelVal, ok := entry.Labels[label]; ok {
+				cfVal, ok := app.Parameters[label]
```
beedub: @ddollar this should return the existing values in the template right?

ddollar: I think you want `app.Parameters["DeploymentMinimum"]` etc

beedub: you're right. updated

ddollar:
Carrying forward in #974

PR #924, opened 2016-07-26
Return error if there is a port conflict with the host

awsmsrc:
Error out if a port conflict with the host is found


awsmsrc:
closing in favor of #941

PR #937, opened 2016-07-28
Fix nat gateway updates

beedub:
There is currently a bug with nat gateway updates. The custom NAT gateway currently handles updates by creating a new NAT gateway after which CF deletes the old NAT gateway. This is problematic because the new NAT gateway will sometimes fail creation because the EIP association still exists with the old NAT gateway.

This changes the update to delete the NAT gateway first, and the create a new NAT gateway. Deletion has also been updated to wait for the EIP association to be deleted in addition.

Also not sure the best way to test this in convox, but I confirmed the behavior/issue above with boto and aws console


beedub:
this is apparently obviated https://github.com/convox/rack/pull/946, which removes the custom nat gateways for CF's stock ones

PR #951, opened 2016-07-28
Fixing regressions

MiguelMoll:
- [x]  Pretty json in the template
- [x] Service command could be a string or array
- [x] If a docker-compose file includes an environment variable but no default value, don't include it
- [x] Fixture ports should be deterministic for testing purposes 

- [x] Fix missing env vars:
Examples: "APP": "httpd",
                  "AWS_REGION": "us-test-2",
                  "PROCESS": "main",
                  "RACK": "convox-test"


MiguelMoll:
Deploy using latest commit looking good:
```
APP	httpd
RACK	devrack
AWS_REGION	us-east-1
RELEASE	RQGUYDGKULA
PROCESS	web
```

PR #956, opened 2016-08-01
Remove health params

beedub:
Reducing unnecessary parameters so that we can have more processes.

This will remove `2 x <LB COUNT>` params

@ddollar


api/models/release.go
```diff
@@ -182,17 +182,7 @@ func (r *Release) Promote() error {
 		return err
 	}
 
-	healthOptions := []string{"port", "path", "timeout"}
```
beedub: port was made into a param, but never used by the app/cf template

Issue #573, opened 2016-04-22
Local dev host port conflict using internal load balancers


mattmanning:
I have a `docker-compose.yml` like this:

```yaml
web:
  links:
    - db1
    - db2
db1:
  ports:
    - 3306
db2:
  ports:
    - 3306
```

The DB image I'm using wants to listen on port 3306 (MySQL). This configuration should be fine in production because each will get its own load balancer. However, locally with `convox start` both containers are started with `3306:3306` as the port mapping, causing a conflict on host port 3306.

Since I'm using linking I don't really care what the host port is. I think this problem could be solved by choosing a random port for it in both cases.

nzoschke:
@mattmanning there is an existing issue about port conflicts in convox start. Can you close this and add it to the discussion here: https://github.com/convox/rack/issues/484

nzoschke:
Closing in favor of https://github.com/convox/rack/issues/484

mattmanning:
Reopening this as it was not solved by https://github.com/convox/rack/pull/662 which closed https://github.com/convox/rack/issues/484

mwarkentin:
@mattmanning Just curious - both of those dbs have "private" port mapping - does this actually cause a host conflict? I wouldn't expect it to be mapping those ports on the host.. 

I could have a misunderstanding of how that works - just curious. :)

PR #960, opened 2016-08-01
Random internal balancer port

awsmsrc:
@mattmanning can we chat through this one today


manifest/process.go
```diff
@@ -37,6 +37,16 @@ func NewProcess(app string, s Service, m Manifest) Process {
 		}
 	}
 
+	for i, p := range s.Ports {
+		// Assign random ports to "internal" balancers
+		if !p.Public {
+			p.Public = true
```
ddollar: Why do we flip this to public? Does this affect the creation of internal load balancers on AWS?

awsmsrc: Nope (as far as I’m aware), process is only used by start.
The fact that this is in the manifest package makes this a little confusion.

I’m a little confused as to the goal of the original issue now as private services should never collide anyway so will discuss this one with @mattmanning when he’s around

ddollar: We should treat manifest as if it is used by both local and production as the eventual goal should be to make that completely true.

Create a docker-compose.yml with `postgres1` and `postgres2` process types and have them both listen on `- 5432` to demonstrate the problem.

awsmsrc: I agree, process/run/sync should be in a run package or even the cli package.
As always the refactor here is a step change.

I have just been using the rails example and done exactly that (double postgres).
The original issue is here: https://github.com/convox/rack/issues/573 it has been closed but I believe the problem persists for certain use cases?? cc @mattmanning

mattmanning: Yes #573 has the description for this problem. It was closed previously, but I reopened it because the problems still wasn't solved.

The goal here is to make `convox start` work when you have more than one internal load balancer listening on the same port.

This isn't a problem in production because each process type gets its own ELB.

It was broken with `convox start` because when we did the port mapping we'd just use a host port that matched the internal port. This caused port conflicts on the host if you had more than one process using the same internal port.

The solution here (as I understand it) was to use random host ports with `convox start` to avoid conflicts.

ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;BT9QqhL3&#x2F;436-random-internal-balancer-port"> random internal balancer port</a>

manifest/process.go
awsmsrc: @mattmanning with a private balancer we no longer bind to the host at all so this doesn't actually break.

This does mean you can't access these services from the host but that should be fine?

Could you test master against a repo you know has this issue?

I would like to check this isn't an issue

ddollar:
What's the status here?

ddollar:
Reopen this if it's still valid

Issue #425, opened 2016-03-04
Convox rack password exposed in stack parameters


mwarkentin:
While poking around in AWS, I was a bit surprised to see the rack password exposed in the CloudFormation parameters list:

![](https://mwarkentin-snaps.s3.amazonaws.com/CloudFormation_Management_Console_2016-03-04_17-20-24.png)

According to [the docs](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html), this parameter could be marked as `NoEcho` to prevent the password from being displayed:

> Important
For sensitive parameter values (such as passwords), set the NoEcho property to true. That way, whenever anyone describes your stack, the parameter value is shown as asterisks (*****).

After discussing w/ @ddollar, he mentioned that this was the current "password recovery" flow for users. Another option might be to follow the standard playbook for website passwords - never "recover", only set a new one. Not sure if there are any other implications to think about.

Slack logs:

```

mwarkentin [5:26 PM] 
Gotcha - thoughts on the Cloudformation Param? Should creds in templates be marked as NoEcho?

[5:27] 
Or is that how the password gets pulled out for the Grid UI, etc.

ddollar [5:28 PM] 
Maybe noecho will have to think about that. Those Parameters are how people recover the password now since its randomly generated

[5:28] 
You can also lock down the cloudformation describe stacks api so you can't read that

mwarkentin [5:30 PM] 
Cool - maybe instead of recovering the password you just set a new one (if you have access to do that)? Similar to a normal web password recovery flow? I don’t know what the implications of that might be. :simple_smile:

ddollar [5:31 PM] 
Definitely worth considering. If you wouldn't mind please file an issue on convox/rack so we can discuss and track
```

nzoschke:
+1 for never recover, always set a new one.

ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;tlpghS9X&#x2F;385-plain-text-rack-password"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> plain text rack password</a>

ddollar:
Fixed in #969

PR #969, opened 2016-08-02
Use NoEcho to mask retrieved CF passwords with ****

ianmalott:


mwarkentin:
Resolves #425

api/provider/aws/helpers.go
```diff
@@ -273,11 +273,16 @@ func (p *AWSProvider) stackUpdate(name string, templateUrl string, changes map[s
...
 		fp, err := formationParameters(string(body))
 
-		// remove params that don't exist in the template
 		for key := range params {
+			// remove params that don't exist in the template
 			if _, ok := fp[key]; !ok {
 				delete(params, key)
 			}
+
+			// exclude Password from params to avoid updating password to "*****"
```
ddollar: Is there some easy way to catch this higher and eliminate all `NoEcho` parameters from being carried forward into this map in the first place?

ddollar:
Does this affect Console and its creation of Racks in any way?

ianmalott:
@ddollar I got some clarification about Console and rack creation from @mattmanning:

> When console installs a Rack it needs to store the Rack password so it can proxy API calls to it. I would assume David is asking if that will be broken by the Password NoEcho setting.

> I'm guessing that since there's no --password option for the CLI installer that console fetches the password from CloudFormation after the install.

> But would need to confirm.

> If that's the case, we might want to have Console set the Password when it installs so it doesn't have to be fetched back.

Not sure how you'd want to proceed from here, but it seems like this is blocked by some Console and CLI changes.

ddollar:
Would it be possible to test this `UsePreviousValue` logic somehow?

ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;EAYn8PFe&#x2F;421-change-password-parameter-to-noecho-on-rack"> Change Password parameter to NoEcho on Rack</a>

ddollar:
Is this only blocking on tests? If so let's rebase and get it in.

Have we verified that it won't break Console rack installation?

ianmalott:
@ddollar: yep, only blocked on tests (and review of refactoring). Wasn't sure how to test a console-based rack installation, but I tested the following from the CLI:

* `convox install --version <noecho-release>` with `--password` and without
* `convox rack update` to a release with `"NoEcho": "false"` for Password param
* `convox rack params set Password=foobarbaz`

mwarkentin:
@ianmalott @ddollar Might be worth adding a small bit of documentation around what you need to do now if you've lost your password?

ianmalott:
@mwarkentin Thanks for the suggestion. @nzoschke has a doc on the way that describes how to set a new Rack API Key/Password: https://github.com/convox/site/pull/229

Issue #403, opened 2016-02-29
More flexible local environment configuration with .env


mwarkentin:
The current way that things work when you run `convox start` seems like it could be improved by allowing `environment` values that are set in `docker-compose.yml` to be overridden by `.env`.

This is based on the following assumptions:

* `docker-compose.yml` is checked into git as it defines your local and convox environments
* `.env` is not checked into git so that you can store credentials there which you would like to keep out of your VCS history

The main use case for this change would be to provide sane defaults for your application in `docker-compose.yml`, but allow developers to override any of them locally without dealing with git issues

### Example

I have a small application which has one configuration variable: `S3_BUCKET`

By default, this will work out of the box for our devs as I've created a shared S3 bucket called `acme-devs-shared`. I've checked the following `docker-compose.yml` file into our github repo:

```yml
main:
  build: .
  environment:
    - S3_BUCKET=acme-devs-shared
  ports:
    - 80:80
```

I have `.env` added to my `.gitignore` file so that we don't accidentally check in any secret credentials there in the future:

```
.env
```

This means that another developer can check out the repo, run `convox start`, and things will work out of the box. However if one of the devs wanted to test out a new feature and use her own S3 bucket, she would need to modify `docker-compose.yml` to either:

Allow the value to be set in `.env`:

```yml
# docker-compose.yml

environment:
    - S3_BUCKET

# .env
S3_BUCKET=my-custom-bucket
```

or just setting it in `docker-compose.yml` directly:

```yml
environment:
    - S3_BUCKET=my-custom-bucket
```

In both cases, the dev would need to modify a file which is checked into version control in order to be able to customize their local environment, which means you run into the possibility of accidentally checking it in, dealing with conflicts when pulling, etc.

It seems ideal to be able to override a set value with `.env` to get around these issues.

The main downside seems like it would be the digression from how [docker-compose](https://docs.docker.com/compose/compose-file/#env-file) works:

> Environment variables specified in `environment` override these values.

mwarkentin:
In https://github.com/docker/compose/issues/2380, they mention using `extends:` to allow for customization outside of git - maybe this is another way to handle it. Does convox support using `extends` within `docker-compose.yml`?

mwarkentin:
Closed by https://github.com/convox/rack/pull/981.

PR #977, opened 2016-08-04
allow user env to override docker-compose.yml

ddollar:
…ntries like FOO=bar


awsmsrc:
Is this the way docker-compose does it?

mwarkentin:
Fixes https://github.com/convox/rack/issues/403

PR #978, opened 2016-08-04
Cleanup after app deletion

MiguelMoll:
Fixes here include for when an app is deleted:

- Delete an app's bucket in batches of 1000 objects to play nice with AWS limits
- Delete builds and releases in batches of 25
- Force delete the app's ECR repo as it might be not be deleted by CloudFormation since there could be images yet to be deleted.

### DIFF
--- a/api/provider/provider.go
+++ b/api/provider/provider.go
@@ -44,7 +44,7 @@ type Provider interface {
 
 	LogStream(app string, w io.Writer, opts structs.LogStreamOptions) error
 
-	ReleaseBatchDelete(app, buildID string) error
+	ReleaseDelete(app, buildID string) error
 	ReleaseGet(app, id string) (*structs.Release, error)
 	ReleaseList(app string, limit int64) (structs.Releases, error)
 	ReleasePromote(app, id string) (*structs.Release, error)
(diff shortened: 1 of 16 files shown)


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;BSaCymsg&#x2F;384-releases-from-old-app-show-up-in-new-app-with-same-name"> releases from old app show up in new app with same name</a>

api/provider/provider.go
```diff
@@ -43,9 +44,9 @@ type Provider interface {
 
 	LogStream(app string, w io.Writer, opts structs.LogStreamOptions) error
 
-	ReleaseDelete(app, id string) (*structs.Release, error)
+	ReleaseBatchDelete(app, buildID string) error
```
ddollar: Can we get away without exposing `ReleaseBatchDelete` in the provider interface?

api/structs/app.go
```diff
@@ -28,3 +30,23 @@ func (a *App) IsBound() bool {
 	// Tags are present but "Name" tag is not, so we have an unbound app.
 	return false
 }
+
+// StackName returns the app's stack if the app is bound. Otherwise returns the short name.
+func (a *App) StackName() string {
+	if a.IsBound() {
```
ddollar: I believe all of the Bound logic should be gone now

api/provider/provider.go
MiguelMoll: We could, but the reason I added it was to keep the the logic for deleting builds and releases separate. That is two calls have to be made for deleting builds and it's releases.

An alternate route I though of is to pass a bool flag to `BuildDelete()` to indicate if the releases should be deleted as well.

ddollar: Would prefer to see it left as `ReleaseDelete` and the controller can call it multiple times. There should not be very many releases per build in most situations.

PR #988, opened 2016-08-04
add Internal param to racks and apps to force creation of internal elbs

ddollar:


mattmanning:
👍

mwarkentin:
What's the difference between this and `ports: 8000`?

ddollar:
If you set `Internal` to `Yes` you will get internal load balancers even if you specify things like `80:5000`

mwarkentin:
Still not sure I'm seeing the use case - is it just to be able to specify a different port to access on the internal LB than the port inside the container? I guess that makes sense to be able to access internal LBs over HTTP(S) without adding a port. 👍

ddollar:
Right, you should still be able to do this with an internal load balancer if you are building intranet-ish apps:

```
labels:
  - convox.port.443.protocol=tls
ports:
  - 80:5000
  - 443:5000
```

PR #1000, opened 2016-08-08
don't pull image if it exists

awsmsrc:
@ddollar this should do the trick

### DIFF
--- a/manifest/build.go
+++ b/manifest/build.go
@@ -2,12 +2,21 @@ package manifest
 
 import (
 	"fmt"
+	"log"
 	"time"
 )
 
 func (m *Manifest) Build(dir, appName string, s Stream, cache bool) error {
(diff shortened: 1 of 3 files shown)


ddollar:
Can we add some quick tests around this?

manifest/build.go
```diff
@@ -2,13 +2,22 @@ package manifest
...
 )
 
 func (m *Manifest) Build(dir, appName string, s Stream, cache bool) error {
 	pulls := map[string]string{}
 	builds := []Service{}
 
+	defer func() {
+		if r := recover(); r != nil {
+			time.Sleep(5 * time.Second)
+			fmt.Println("Recovered in f")
```
ddollar: Debug print?

awsmsrc:
Tests added

PR #1001, opened 2016-08-08
error out if link to non existant service or if service doesn't expos…

### LINKED ISSUE #990: Links without ports defined
A few changes to the behavior of `convox start` with regard to link variables:

- [ ] Print a warning (or even error) if a service defines a link to a service with no defined ports.
- [ ] If just a warning, do not set link env vars when linking to a service with no defined ports.

Changes to `convox build`:

- [ ] Print a warning (or error, same as above) if the docker-compose.yml tries to link to a service with no defined ports.

### LINKED ISSUE #994: validate heath.check.timeout value
The max health check timeout allowed by AWS is 60 seconds. When parsing the `convox.health.timeout` label, an error should be thrown for values greater than 60.

awsmsrc:
…e ports 

fixes #990 
fixes #994

### DIFF
--- a/manifest/fixtures/invalid-cron.yml
+++ b/manifest/fixtures/invalid-cron.yml
@@ -1,0 +1,11 @@
+version: "2"
+services:
+  web:
+    build: .
+    command: bin/web
+    dockerfile: Dockerfile.dev
+    entrypoint: /sbin/init
+    labels:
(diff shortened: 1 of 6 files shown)


ddollar:
Can we add some quick tests around this?

awsmsrc:
tests added cc @MiguelMoll @ddollar

PR #1004, opened 2016-08-08
dont require a label set to tcp

ddollar:


ddollar:
Tried to timebox some tests for this but ended up having to blow up most of the Promote() function. Bailing for now

PR #1011, opened 2016-08-09
use public subnets for rack unless balancer is internal

ddollar:


api/models/templates/app.tmpl
```diff
@@ -382,18 +382,16 @@
...
-          { "Ref": "Subnets" }
-        ] },
+        {{ if eq .Scheme "internal" }}
+          "Scheme": "internal",
+          "Subnets": { "Fn::If": [ "Private",
+            { "Ref": "SubnetsPrivate" },
+            { "Ref": "Subnets" }
+          ] },
+        {{ else }}
+          "Scheme": { "Fn::If": [ "Internal", "internal", { "Ref": "AWS::NoValue" } ] },
```
beedub: isn't this unnecessary given that scheme == internal is checked above?

ddollar: There is the CF parameter "Internal" and the local check of .Scheme == "internal" (the latter is set when you use only a single port pair)

api/models/templates/app.tmpl
```diff
@@ -382,18 +382,16 @@
...
+          "Subnets": { "Fn::If": [ "Internal", { "Ref": "SubnetsPrivate" }, { "Ref": "Subnets" } ] },
```
beedub: same vein as above, couldn't this be simplified to `"Subnets": { "Ref": "Subnets" }`?

api/models/templates/app.tmpl
beedub: ah gotcha

PR #1015, opened 2016-08-10
reuse identical builds

ddollar:


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;98wfA4SQ&#x2F;452-reuse-identical-builds-via-tags"> reuse identical builds via tags</a>

awsmsrc:
makes sense, this will speed builds up a lot
:+1:

manifest/build.go
```diff
@@ -21,7 +21,16 @@ func (m *Manifest) Build(dir, appName string, s Stream, cache bool) error {
 		}
 	}
 
+	buildCache := map[string]string{}
+
 	for _, service := range builds {
+		if bc, ok := buildCache[service.Build.Hash()]; ok {
```
awsmsrc: I'm not convinced Hash() is necessary 

map[manifest.Build]string{} would probably work

ddollar: I think you can't use things with a map as a key. See the "Key types" section of https://blog.golang.org/go-maps-in-action

That being said I think this isnt properly hashing the args map.

PR #1016, opened 2016-08-10
use convox/golang

ddollar:


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;uB8EpMCw&#x2F;453-use-convox-golang"> use convox&#x2F;golang</a>

mwarkentin:
Switching off of alpine? Any specific reasons why? :)

PR #1029, opened 2016-08-10
Latest version is not an error

mwarkentin:
I don't think having an up-to-date rack should be considered an error:

```
$ convox rack releases
VERSION                         UPDATED       STATUS
20160810023912                  2 hours ago   active
20160809175449-subnet-behavior  23 hours ago
20160808193602                  1 day ago
ERROR: current version "20160810023912" is latest
```

Note: Not tested.

### DIFF
--- a/vendor/github.com/convox/version/version.go
+++ b/vendor/github.com/convox/version/version.go
@@ -184,7 +184,7 @@ func (vs Versions) Next(curr string) (string, error) {
 		return published, nil
 	}
 
-	return "", fmt.Printf("current version %q is latest", curr)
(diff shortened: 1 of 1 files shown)


mwarkentin:
@ddollar @nzoschke After discussing this with @mattmanning in Slack:

* This is a vendored file, so the actual change needs to be made upstream
* Can I get some feedback on whether this change actually makes sense? It seems like the calling function expects either a version number, or an error - do you think that logic makes sense as it stands? Should we potentially be returning the latest version (even if it's the same as the current), and doing the comparison afterwards?

ddollar:
Closing to track as issue #1135

PR #1030, opened 2016-08-10
Fix convox/proxy not being attached to custom network

### LINKED ISSUE #1020: Local proxy does not get attached to correct docker network
`convox start` does not currently pass `--net` parameter when starting the `convox/proxy` containers, which results in the docker failing to link containers. In addition, the failed `docker run` commands for the proxy containers don't get logged, so there's no indication of failure, apart from the proxies being missing.

bladealslayer:
Fixes #1020.

### DIFF
--- a/manifest/service.go
+++ b/manifest/service.go
@@ -107,6 +107,7 @@ func (s *Service) Proxies(app string) []Proxy {
 				Balancer:  p.Balancer,
 				Container: p.Container,
 				Host:      fmt.Sprintf("%s-%s", app, s.Name),
+				Network:   s.NetworkName(),
 			}
 
 			s.Ports[i].Balancer = 0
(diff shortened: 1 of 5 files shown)


manifest/service.go
```diff
@@ -82,11 +82,19 @@ func (s *Service) Proxies(app string) []Proxy {
 		if p.External() {
 			name := fmt.Sprintf("%s-%s-proxy-%d", app, s.Name, p.Balancer)
 
+			networkName := ""
```
ddollar: Maybe a short comment above this logic talking about its intent? Are we always getting the last network defined?

bladealslayer: Sure, will add that. I used same logic as for launching the service containers themselves: https://github.com/convox/rack/blob/master/manifest/process.go#L56 It picks the last defined network, yes. According to documentation, `convox` supports only single network, no? https://convox.com/docs/docker-compose-file/

bladealslayer:
@ddollar I refactored this a bit to reuse the network name lookup logic. Does it look ok to you?

ddollar:
Looking good. Please rebase when you can. Would it be possible to add some tests around this NetworkName?

bladealslayer:
@ddollar rebased and added tests.

PR #1041, opened 2016-08-11
Handle manifests with repeated images

mattmanning:
When a manifest used the same image for multiple processes we were only creating one image tag, which led to errors when we tried to later tag and push into ECR (image would not be found). 

This PR maps images to a slice of tags instead of a single tag, then loops over those when running the actual `docker tag` commands.

### DIFF
--- a/manifest/build.go
+++ b/manifest/build.go
@@ -6,7 +6,7 @@ import (
 )
 
 func (m *Manifest) Build(dir, appName string, s Stream, cache bool) error {
-	pulls := map[string]string{}
+	pulls := map[string][]string{}
 	builds := []Service{}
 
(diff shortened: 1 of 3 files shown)


ddollar:
:+1: just as a git style comment combining your last two commits would be better for `git bisect` purposes. Ideally each commit should compile/run validly on its own.

mattmanning:
@ddollar Good point. Just rewrote the commits accordingly.

mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;tB414nSG&#x2F;486-handle-manifests-with-repeated-images"> Handle manifests with repeated images</a>

PR #1043, opened 2016-08-12
Make stopping of local containers more graceful

### LINKED ISSUE #1042: Possible data corruption with convox start
`convox start` uses `docker kill` to stop containers, which doesn't allow processes to exit cleanly. This can result in data corruption when data volumes are used.

bladealslayer:
"docker kill" doesn't allow processes to exit cleanly, which can result in data corruption
when data volumes are used. "docker stop" first sends a SIGTERM and then SIGKILL
only if process didn't exit within a timeout.

Resolves #1042.

### DIFF
--- a/manifest/run.go
+++ b/manifest/run.go
@@ -175,13 +175,17 @@ func (r *Run) Wait() error {
 }
 
 func (r *Run) Stop() {
-	for _, p := range r.processes {
-		Docker("kill", p.Name).Run()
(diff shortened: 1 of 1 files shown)


ddollar:
Please rebase this and we'll get it into our next release. Thanks!

bladealslayer:
@ddollar rebased!

PR #1045, opened 2016-08-12
[WIP] Provisional ALB support

ddollar:
Adds provisional support for ALBs on a per-app level. To use, add a `convox.router.path` label to your `docker-compose.yml`:

```
web:
  labels:
    - convox.router.path=/*
  ports:
    - 80
```
- https://github.com/convox-examples/router/blob/master/docker-compose.yml
- http://convox-router-567724791.us-east-1.elb.amazonaws.com/
- http://convox-router-567724791.us-east-1.elb.amazonaws.com/test/

**KNOWN LIMITATIONS**
- Does not yet support SSL
- Router priority is currently assigned by the start order of the docker-compose.yml (dependency order based on links then alphabetical by process name)


dlanger:
For us, the most attractive feature of ALBs is that they enable running more than one copy of a container on a single instance. This greatly simplifies capacity planning and makes it easier to think of an ECS cluster as a pool of compute resources, and means I no longer have to do separate tracking to ensure that `max(number of copies of a single container I'd like to run) <= number of instances`.

ddollar:
So one thing to note is that running those processes on multiple instances is a lot more fault-tolerant. If you stack multiple processes on a single instance then it's a lot easier to take it down with the loss of a single instance.

mwarkentin:
@ddollar @dlanger was more talking about if we wanted to scale up to 6 processes, that we could do that at 2x3 instances or whatever, not running a whole bunch on 1 instance. Like he mentioned, it simplifies our resource calculations on one dimension.. another nice benefits is that we could see some cost savings on things like Datadog and ThreatStack where things are billed on a per-instance basis.

scottmessinger:
Hi! @ddollar asked for me to leave a comment in this PR with my use case for ALB. We're deploying an Elixir app that uses websockets. From what I understand, [ELB doesn't support sticky sessions with websockets](https://medium.com/@Philmod/load-balancing-websockets-on-ec2-1da94584a5e9#.4rbtj2b17). Also, there seem to be some complexity in simply setting up websockets on ELB from the scattering of blog posts I've read. ALB is attractive because, from what I understand, it would just work. (Btw, I could be off here -- I'm coming from Heroku so I haven't had to get into the weeds with AWS deployments)

idyll:
That's crazy @scottmessinger -- we are in the exact same position. Some Elixir (other stuff still in Ruby) and we really need the ALB support to remove the need for ProxyProtocol.

We are actively porting more and more ruby to elixir so the pain of missing ALB support is getting worse for us.

mattmanning:
Closing in favor or https://github.com/convox/rack/issues/1373

jkurz:
So since this PR is closed, is convox still behind this blog post? https://convox.com/blog/alb/

It points people to use this branch to get ALB features going. Maybe an update should be in order, saying that the real solution is still in the works and this is for testing and should not be used in production. Unless I am wrong and convox is saying this could be used in production?

ddollar:
I would not use this PR in production mainly because it has branched from master and has not received later updates, such as security-related AMI updates.

We will likely still add support for ALB but still have open questions as to the cost/benefit tradeoff that we are working through in #1373

jkurz:
I understand @ddollar just wanted to clarify, as it was not clear coming from the blog post. The blog post reads like this is going to be the solution.

ddollar:
Thanks for asking! Hopefully these comments will help clear that up.

PR #1046, opened 2016-08-12
replace a service during a role change

ddollar:
need to replace services if their roles change


awsmsrc:
Side note to this PR but based on the work I was doing we really need to make these handler functions testable

ddollar:
Yes, I would really like to test Lambda JSON input -> AWS API requests

PR #1064, opened 2016-08-15
Explicit build call for `monitor` added

MiguelMoll:
Previous method wasn't building `monitor`


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;yZs7Z817&#x2F;499-small-bug-not-building-monitor-cmd"> small bug not building monitor cmd</a>

PR #1070, opened 2016-08-16
Adding conditional to check for an app thats linked to a service. 

adamenger:
This will disallow a user to delete an app that has an existing service link. Right now if you delete an app that's linked to a service it breaks amazon which requires manual modification of the cloudformation template for the service.


api/controllers/apps.go
```diff
@@ -100,6 +100,23 @@ func AppDelete(rw http.ResponseWriter, r *http.Request) *httperr.Error {
...
+		return httperr.Server(err)
+	}
+	for _, serviceName := range services {
+		service, err := provider.ServiceGet(serviceName.Name)
+		if err != nil {
+			return httperr.Server(err)
+		}
+		for _, linkedApp := range service.Apps {
+			if linkedApp.Name == app.Name {
+				return httperr.Errorf(412, "%s is linked to the service %s. You must delete all service links before deleting the app %s", app, service.Name, app)
```
mwarkentin: This will return on the first service it finds, correct? Does it make sense to do so, or to find the full list of linked services and then return a message enumerating all of them? Could save an annoying cycle of delete app, delete service, delete app, delete service.. but maybe it's not common to have too many linked services on the other hand.

adamenger: That's a great point. I imagine over time it would become commonplace for apps to have multiple linked services. I'll defer to @ddollar for the preferred user experience but populating an array and returning that won't be difficult.

adamenger:
Could use some guidance debugging this:
```
--- FAIL: TestAppDelete (0.01s)
	Error Trace:	apps_test.go:174
	Error:		Expected nil, but got: &json.UnmarshalTypeError{Value:"string", Type:(*reflect.rtype)(0xc44200), Offset:88}
```

api/controllers/apps.go
ddollar: Go ahead and return one error but make it `%s is linked to these services and cannot be deleted: service1, service2, service`

adamenger: OK, thanks for the feedback. Patch will be pushed momentarily.

adamenger:
Interesting: 

```
--- FAIL: TestAppDelete (0.00s)
panic: 
assert: mock: I don't know what to return because the method call was unexpected.
	Either do Mock.On("ServiceList").Return(...) first, or remove the ServiceList() call.
	This method was unexpected:
		ServiceList()
```

Thoughts on how I could mock this?

ddollar:
Closing this for now because of age but will create an issue.

PR #1088, opened 2016-08-17
WIP: Use `UsePreviousValue` in UpdateParams

ianmalott:
When NoEcho was turned on for the CF Password param, we updated one relevant code path (`convox rack update`), but didn't update the second (`convox rack params set`). This updates the second code path so that all stack update paths should now use `UsePreviousValue` to play nicely with NoEcho params.

### DIFF
--- a/test/aws_cycles.go
+++ b/test/aws_cycles.go
@@ -1,6 +1,7 @@
 package test
 
 import (
+	"fmt"
 	"net/http/httptest"
 	"os"
 	"strings"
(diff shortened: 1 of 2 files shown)


nzoschke:
LGTM. We'll want to test this well and get the console patch out with it.

test/aws_cycles.go
```diff
@@ -191,8 +192,22 @@ func GetItemAppReleaseCycle(appName string) awsutil.Cycle {
 }
 
 func UpdateAppStackCycle(stackName, count, memory string) awsutil.Cycle {
+	const countPreviousValue string = "1"
```
nzoschke: Small go feedback. This could be something simpler like `previousCount := "1"`, and I think in general with Go simpler code wins.

PR #1104, opened 2016-08-19
only randomize the instance ports for old apps

ddollar:


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;xZQUVsak&#x2F;520-service-handlers-followup"> Service Handlers Followup</a>

api/models/release.go
```diff
@@ -285,11 +287,12 @@ func (r *Release) Promote() error {
 		}
 	}
 
-	// randomize the instance ports
-	// TODO: only do this for an old app
-	for key := range app.Parameters {
-		if strings.HasSuffix(key, "Host") {
-			app.Parameters[key] = strconv.Itoa(rand.Intn(50000) + 10000)
+	// randomize the instance ports for older apps so we can upgrade smoothly
+	if oldVersion <= "20160818013241" {
```
beedub: this seems like it should be `<`, not `<=`, as `20160818013241` includes the migration

PR #1108, opened 2016-08-19
Fix services info, url, and delete

ddollar:


provider/aws/service.go
```diff
@@ -149,11 +149,36 @@ func (p *AWSProvider) ServiceGet(name string) (*structs.Service, error) {
...
+			s.Exports["URL"] = parsedUrl.Query().Get("endpoint")
+		}
+	}
+
 	// Populate linked apps
 	for k, _ := range s.Outputs {
 		if strings.HasSuffix(k, "Link") {
 			n := DashName(k)
-			app := n[:len(n)-5]
+			app := n[:len(n)-4]
```
nzoschke: was `-5` a bug?

ddollar: As far as I can tell

bladealslayer: This actually seems to leave a trailing dash now. What was the problem with `-5`?

PR #1109, opened 2016-08-20
Fix List container instances everywhere

nzoschke:
Released as 20160824024411-list-container-instances for testing. `convox instances` and `convox ps` are very slow.


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;op7VvDZL&#x2F;560-fix-for-instance-list-with-pagination"> fix for instance list with pagination</a>

PR #1112, opened 2016-08-21
Updating FluentD service w/ fixes

adamenger:
Patching the string log line decoding function which failed to parse on a json object with a string. Added some tests to prevent this from happening in the future.


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;4NCYG2ud&#x2F;545-fix-for-fluentd-service-lambda"> Fix for fluentd service lambda</a>

Issue #1111, opened 2016-08-20
Commands that depend on updating the stack don't handle the potential of a rollback


beedub:
Promoting a release doesn't (anymore?) guarantee that it will be successful anymore due to the change to use `AWS::ECS::Service` instead of the custom service handler.

The `AWS::ECS::Service` resource depends on the service to stabilize completely (ie. desired count of new deployment must be met) or the update will fail and the whole cloudformation update will rollback, essentially meaning the promotion will rollback as well, all unbeknownst to the end-user. 

Several of my deployments have already silently rolled back, and I've been forced to watch cloudformation and the actual ecs services to see if the deployment is actually going to succeed in time and is ultimately successful. 

Some notes:
- In my latest rollback, the cloudformation logs indicate that CF started the rollback 11 minutes after the service update was first initiated.
- The critical factor for the deployments is capacity. If there's 0 capacity, then the services cannot obviously stabilize. If there's only a tiny bit of capacity, but the services themselves have very restrictive deployment settings, then the services will update slowly and may be subject to the 11 minute timeout.
- Lowering the deployment minimum from 100 seems now to be a good idea to reduce the chance that services will not relaunch in the timeout window

beedub:
Another place where this happens is `convox scale`

beedub:
AWS support reply on timeout:

> The internal team has got back to us saying there is a hard timeout for ECS service stabilization of 10 minutes. The ECS service has 10 minutes to be up and running successfully before CloudFormation automatically times out. This does not reset based on progress of the deployment.

Further:

> Currently works are on in order to increase this to 15 minutes (or even 30) as it's been agreed on, that the current timeout of 10 minutes is a bit too short.

I asked for an ETA on these changes to be released.

In the meantime, this would be helpful to document on convox's deployment docs.

ddollar:
This is now documented on https://convox.com/docs/rolling-updates/

PR #1113, opened 2016-08-21
More helpful build error msg

beedub:
When the stack rolls back, a build cannot be created until the rollback completes. The previous err msg was generic: `ERROR: unable to build app: my-app`. Now the build err msg will include the status.

(This has become an issue because of https://github.com/convox/rack/issues/1111)

### DIFF
--- a/cmd/convox/builds.go
+++ b/cmd/convox/builds.go
@@ -158,12 +158,8 @@ func cmdBuildsCreate(c *cli.Context) error {
 		return stdcli.ExitError(err)
 	}
 
-	switch a.Status {
-	case "creating":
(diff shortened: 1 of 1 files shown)


ddollar:
I can't think of any reason we need to block builds if an app is updating.

beedub:
That makes sense. Was just copying over what was previously there. Do
builds depend on stack creation as well?

ddollar:
They probably need the initial stack creation to be done so the registry is around.

beedub:
@ddollar updated

ddollar:
Rolled in to #1231

PR #1122, opened 2016-08-23
Convox path sweep

MiguelMoll:
Add a `/convox/` path to resources of type `AWS::IAM::*` where allowed.

This change will cause the creation of new instances as a change was made to the Instance Profile/Role.


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;1B06nRWZ&#x2F;537-change-convox-path-for-aws-iam-types"> Change convox path for AWS::IAM::* types</a>

ddollar:
Have you tested these changes for service disruptions with rack API and/or running apps?

MiguelMoll:
Apps are be fine as long as they're scaled to two or more processes. Rack API would experience a brief downtime as the creation of a new ECS removes the exiting EC2 instances from the Rack ELB.

PR #1124, opened 2016-08-23
Update postgres version to latest 9.5.2

lisajluo:


contra:
Shouldn't there be a way to update existing instances from the CLI?

PR #1132, opened 2016-08-24
Validate parameter is not an empty string for S3 and SNS

MiguelMoll:


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;Nd2ToAL6&#x2F;551-s3-and-sns-parameter-check"> s3 and sns parameter check</a>

PR #1137, opened 2016-08-25
Normalize DescribeInstances calls

nzoschke:
Follow up to #1109. Use a valid EC2 DescribeInstances call that will return more than >100 instances.


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;op7VvDZL&#x2F;560-fix-for-instance-list-with-pagination"> fix for instance list with pagination</a>

PR #1144, opened 2016-08-25
point to correct fluentd lambda code

ddollar:
Related to #1112


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;mFgG9m6M&#x2F;578-fluentd-s3-location"> fluentd s3 location</a>

PR #1145, opened 2016-08-25
Build export and import

MiguelMoll:
Goal is to simplify build workflows by taking the build from RackA AppA to RackB AppB. That way the same image can be used throughout different environments including production.

As a first pass, the following CLI commands are:

Export will create a local artifact from stdout
```
 convox builds export -a appname BLTKJOQFMHA > build.tgz
```

Import will take a build artifact and create a new build from it in the alternate rack/app
```
convox builds import -a appname < build.tgz
```
Idea is to `convox switch` to a new rack in between export and import.


MiguelMoll:
This PR brings in an update aws-go-sdk. As such, I'll update the tests to accommodate the changes. The update was to take advantage (but not limited to) of the new ECS calls.

Testing strategy for the import/export feature is still a WIP. Thoughts welcomed.

MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;sghqRlOv&#x2F;493-convox-builds-export-import"> convox builds export &#x2F; import</a>

ddollar:
Needs to be rehomed on master now

api/controllers/builds.go
```diff
@@ -153,6 +158,104 @@ func BuildDelete(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	return RenderJson(rw, build)
 }
 
+// BuildImport imports a build for an app
+func BuildImport(rw http.ResponseWriter, r *http.Request) *httperr.Error {
```
ddollar: Almost all of this controller should be in the BuildImport provider method. The controller should have no knowledge of Docker, just handing the file upload straight off to `provider.BuildImport(app, reader)`

api/controllers/routes.go
```diff
@@ -20,10 +20,12 @@ func NewRouter() (router *mux.Router) {
 	router.HandleFunc("/apps/{app}", api("app.delete", AppDelete)).Methods("DELETE")
 	router.HandleFunc("/apps/{app}/builds", api("build.list", BuildList)).Methods("GET")
 	router.HandleFunc("/apps/{app}/builds", api("build.create", BuildCreate)).Methods("POST")
+	router.HandleFunc("/apps/{app}/builds/import", api("build.import", BuildImport)).Methods("POST")
```
ddollar: Can this be part of POST /apps/{app}/builds ?

The former already has a few different input types (tarball, repo url, incremental index)

api/controllers/routes.go
```diff
@@ -20,10 +20,12 @@ func NewRouter() (router *mux.Router) {
...
 	router.HandleFunc("/apps/{app}/builds/{build}", api("build.get", BuildGet)).Methods("GET")
 	router.HandleFunc("/apps/{app}/builds/{build}", api("build.update", BuildUpdate)).Methods("PUT")
 	router.HandleFunc("/apps/{app}/builds/{build}", api("build.delete", BuildDelete)).Methods("DELETE")
 	router.HandleFunc("/apps/{app}/builds/{build}/copy", api("build.copy", BuildCopy)).Methods("POST")
+	router.HandleFunc("/apps/{app}/builds/{build}/export", api("build.export", BuildExport)).Methods("GET")
```
ddollar: How about `GET /apps/{app}/builds/{build}.tgz` ?

api/structs/app.go
```diff
@@ -14,6 +14,12 @@ type App struct {
 
 type Apps []App
 
+// AppRepository defines an image repository for an App
```
ddollar: Probably don't need to expose repos outside of the provider

provider/provider.go
```diff
@@ -12,13 +12,16 @@ import (
 type Provider interface {
 	AppGet(name string) (*structs.App, error)
 	AppDelete(name string) error
+	AppRepository(name string) (*structs.AppRepository, error)
```
ddollar: Don't need to expose repositories from the provider interface

api/controllers/routes.go
MiguelMoll: Definitely could. I was thinking of having Import return a release so that the newly imported build could be released. Thoughts on that since a code path could return a release struct vs the others returning a build. Seems a little off since the function is called `BuildCreate`.

ddollar: All of the build paths return a Build object which itself has a Release attribute (that's how things like `convox build --promote` work)

MiguelMoll: So determined to return a release struct, that went over my head. Good to go.

ddollar:
Needs controller and provider tests covering the new functionality.

mwarkentin:
Would you be able to do something like `convox builds export -a appname BLTKJOQFMHA --rack stage > convox builds import -a appname --rack prod`?

MiguelMoll:
@mwarkentin That's one option. :+1:

PR #1147, opened 2016-08-26
Lessen chance of throttling during builds

nzoschke:
Before this patch GetBuild() always calls:
- CloudFormation DescribeStack on the app
- DynamoDB Query for the build ID
- S3 Get for the build logs

GetBuild is also called every 1 second during a build, to poll build status for the build log websocket to know when to disconnect.

This seems like a probable cause for throttling during builds.

This patch makes GetBuild() only call DynamoDB which we can poll more reliably, and introduces GetBuildLogs() which uses CF and S3 and is only needed on a `convox builds info`.

### DIFF
--- a/provider/provider.go
+++ b/provider/provider.go
@@ -19,7 +19,7 @@ type Provider interface {
 	BuildCreateTar(app string, src io.Reader, manifest, description string, cache bool) (*structs.Build, error)
 	BuildDelete(app, id string) (*structs.Build, error)
 	BuildGet(app, id string) (*structs.Build, error)
-	BuildGetLogs(app, id string) (string, error)
+	BuildLogs(app, id string) (string, error)
 	BuildList(app string, limit int64) (structs.Builds, error)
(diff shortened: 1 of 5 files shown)


provider/provider.go
```diff
@@ -19,6 +19,7 @@ type Provider interface {
 	BuildCreateTar(app string, src io.Reader, manifest, description string, cache bool) (*structs.Build, error)
 	BuildDelete(app, id string) (*structs.Build, error)
 	BuildGet(app, id string) (*structs.Build, error)
+	BuildGetLogs(app, id string) (string, error)
```
ddollar: Let's call this just `BuildLogs`

ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;CZMvPKDB&#x2F;568-goodeggs-aws-throttling-exceptions"> Goodeggs: AWS Throttling Exceptions</a>

PR #1156, opened 2016-08-27
Handle process names with hyphens

mattmanning:
When process names have hyphens they can get added to the Rule and LambdaPermission items in the CF template, causing a template error.

This change drops them. It also allows dashes in cronjob names.

This doc change should be deployed when this is released https://github.com/convox/site/pull/235

### DIFF
--- a/api/models/fixtures/cron_labels.yml
+++ b/api/models/fixtures/cron_labels.yml
@@ -7,4 +7,4 @@ main:
 really-long-process-type-name:
   build: .
   labels:
-    - convox.cron.really-long-cron-job-name=0 * * * ? bin/myjob
+    - convox.cron.ReallyLongCronJobName=0 * * * ? bin/myjob
--- a/api/models/fixtures/cron_labels.yml
+++ b/api/models/fixtures/cron_labels.yml
@@ -7,4 +7,4 @@ main:
 really-long-process-type-name:
   build: .
   labels:
-    - convox.cron.ReallyLongCronJobName=0 * * * ? bin/myjob
+    - convox.cron.really-long-cron-job-name=0 * * * ? bin/myjob
--- a/api/models/cronjob.go
+++ b/api/models/cronjob.go
@@ -4,6 +4,7 @@ import (
 	"crypto/sha256"
 	"encoding/base32"
 	"fmt"
+	"regexp"
 	"strings"
 
 	"github.com/convox/rack/manifest"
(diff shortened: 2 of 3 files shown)


mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;pkNxSMdm&#x2F;586-don-t-break-cron-when-process-names-have-hyphens"> Don&#39;t break cron when process names have hyphens</a>

ddollar:
Should this really be removal of non-alpha characters?

mattmanning:
Yeah I guess it should actually remove all non-alphanumerics. Should maybe still titleize before removing to make things as human-readable as possible.

api/models/fixtures/cron_labels.yml
```diff
@@ -7,4 +7,4 @@ main:
 really-long-process-type-name:
   build: .
   labels:
-    - convox.cron.really-long-cron-job-name=0 * * * ? bin/myjob
```
ddollar: How does this patch behave when the cron job name has dashes in it?

mattmanning: @ddollar We specify that dashes are not valid in cron names.

https://convox.com/docs/scheduled-tasks/ 

> Task names may not be reused within the same application process and must be alphanumeric

The reason we did that is because if we stripped dashes you could have a job called `MyJob` and one called `my-job` and they would conflict. Now that I think about it, though, that same problem can happen with processes.

`my-process` and `my_process` could currently conflict. Will think about this some more . . .

ddollar: Do we actually prevent the dashes from being used in the first place?

mattmanning: Only via template error propagating up:

```
ERROR: Template format error: Resource name WebJob-With-DashRule is non alphanumeric.
```

mattmanning: Thinking now we should allow anything in a cron job name that is a valid Docker label key:

> * Keys should only consist of lower-cased alphanumeric characters, dots and dashes (for example, [a-z0-9-.]).
> * Keys should start and end with an alpha numeric character.
> * Keys may not contain consecutive dots or dashes.

https://docs.docker.com/engine/userguide/labels-custom-metadata/

mattmanning: Hmm there was some existing validation, but we were allowing dashes:

https://github.com/convox/rack/blob/master/manifest/manifest.go#L93-L111

mattmanning: Just made the change to allow dashes in cronjob names for now.

Allowing dots in cronjob names doesn't seem like a good idea because we're using dots to signify namespaces.

I think there's a separate PR to be made to enforce starting and ending with an alphanumeric characters and disallowing consecutive dots and dashes across all of our labels.

PR #1159, opened 2016-08-27
Only fail the build on ExitErrors

ddollar:
https://golang.org/pkg/os/exec/#Cmd.Wait

Should fix the "No child processes" race some users have been reporting.


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;XsGxTyOy&#x2F;589-fix-build-no-child-processes-wait-error"> fix build: no child processes wait error</a>

PR #1165, opened 2016-08-29
Block underscore in service name

MiguelMoll:
Block the use of underscores in a service name. While a simple replace of an underscore with a dash would solve the issue with respect to tagging, it opens another path of potential issues with Cloudformation stack and parameters.

```
Promoting RBCMCAUEVTI... ERROR: Template format error: Parameter name Web_apiPort8080Certificate is non alphanumeric.
```

### DIFF
--- a/manifest/fixtures/underscore_service.yml
+++ b/manifest/fixtures/underscore_service.yml
@@ -1,0 +1,4 @@
+web_api:
+  image: httpd
+  ports:
+    - 80:80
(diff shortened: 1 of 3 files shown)


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;4S7EIQ4T&#x2F;565-unable-to-use-underscore-in-service-name-in-compose-file"> Unable to use underscore in service name in compose file</a>

PR #1170, opened 2016-08-30
Update ELB name when switching to PrivateApi

mattmanning:
This avoids a CloudFormation error that occurs when trying to replace a custom-named resource.


mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;GQOaOJwl&#x2F;609-update-elb-name-when-switching-to-privateapi"> Update ELB name when switching to PrivateApi</a>

PR #1173, opened 2016-08-31
Fix autoscaler process width calculation

ddollar:


nzoschke:
The 5s tick concerns me about rate limiting and throttling.

ddollar:
Oh my bad that was just for debugging, will fix

api/workers/autoscale.go
```diff
@@ -72,10 +75,7 @@ func autoscaleRack() {
 
 	err = models.Provider().SystemSave(*system)
 	if err != nil {
-		log.Log("fn=system.Save err=%q", err)
+		log.Error(err)
 		return
 	}
-
-	// log for humans
-	fmt.Printf("who=\"convox/monitor\" what=\"autoscaled instance count to %d\" why=\"a service wants %s processes behind a load balancer\"\n", system.Count)
```
nzoschke: These have been helpful for debugging... The goal is to have a simple rack logs filter we can document so users can understand some internal changes that are executed and why

nzoschke:
There is an issue around this. Can you also review the memory calculation while you are in here? I think we want to add one extra instance because we can't always perfectly pack for memory

ddollar:
The +1 for extra room has already been in for quite some time.

ddollar:
This PR also now takes CPU into account for scaling and displays the autoscaling reason.

To search for autoscale events you can `--filter="autoscaleRack change="`

api/workers/autoscale.go
ddollar: This message was misleading. Autoscale can happen for more reasons than just process count. It has been replaced by better logs.

ddollar:
Just a note of order. Line-based comments are better if possible because you can "clear" them when the code changes.

mwarkentin:
Might be nice to have a note in the [AutoScale docs](https://convox.com/docs/scaling/) about how to search for autoscale events?

ianmalott:
@mwarkentin 👍 adding a note now

mwarkentin:
Thanks!

PR #1182, opened 2016-09-02
Extract app name from log group

MiguelMoll:
Generally a length of 5 works. A catch is when an app ends in a number (e.g. httpd80). This would create a link output value of `Httpd80Link`. When an app name is retrieved from that, dashName() returns `foo80link`. Removing the last 5 characters results in `foo8`.

**UPDATE**: As a stop gap, extract app name from log group. This method helps avoid some edge cases but also could cause an issue if a user has two apps (following the pattern of foo80 and foo-80) they'll only be able to link to one or another.

### DIFF
--- a/provider/aws/service.go
+++ b/provider/aws/service.go
@@ -199,10 +199,12 @@ func (p *AWSProvider) ServiceGet(name string) (*structs.Service, error) {
 	}
 
 	// Populate linked apps
-	for k, _ := range s.Outputs {
-		if strings.HasSuffix(k, "Link") {
-			n := dashName(k)
-			app := n[:len(n)-5] // 5 == len("-link")
+	for key, value := range s.Outputs {
+		if strings.HasSuffix(key, "Link") {
+			// Extract app name from log group
+			index := strings.Index(value, "-LogGroup")
+			r := strings.NewReplacer(fmt.Sprintf("%s-", p.Rack), "", value[index:], "")
+			app := r.Replace(value)
 
 			a, err := p.AppGet(app)
 			if err != nil {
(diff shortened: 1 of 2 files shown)


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;fWE8GquS&#x2F;648-syslog-unlinking-deleting-not-working"> Syslog unlinking &#x2F; deleting not working</a>

ddollar:
This change seems like it won't help for the other case you mentioned since the len("-Link") is static

MiguelMoll:
Right, I'm trying to see a way to fix that example without breaking the naming scheme. That is, we want to avoid adding a `-` when the app name doesn't have one and vice versa.

ddollar:
Can we fix the de-title-caser to add the dash on the number boundary?

bladealslayer:
Won't changing `dashName` result in breaking current apps named in the form `AppNN`? Could we, instead just check if suffix is `-link` or just `link` and trip appropriately? E.g.:

```go
if strings.HasSuffix(k, "Link") {
  n := dashName(k)
  app := n[:len(n)-4]
  if strings.HasSuffix(n, "-link") {
    app = n[:len(n)-5]
  }
...
}
```

MiguelMoll:
Breaking current apps is exactly what we want to avoid. Splitting on `Link` is only part of the problem though. Given a link value of `Httpd80Link`, `dashName()` returns `httpd80link`. This could still run into "app not found" error if the original app is named `httpd-80` (not `httpd80`).

bladealslayer:
So currently both `httpd80` and `httpd-80` as app names would result in `Httpd80Link`, making it impossible to guess which is the original?

mattmanning:
Built a branch release from this and did some testing. Working for me 👍

PR #1184, opened 2016-09-02
attempted fix for wait: no child processes

ddollar:
Uses a more fleshed-out PID 1 that does not `log.Fatal()` on `ECHILD`


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;E3F2fI9V&#x2F;623-wait-no-child-processes"> wait: no child processes</a>

PR #1189, opened 2016-09-04
Build import and export

ddollar:
Carrying #1145 here because Hound is messing up hard on the old one.


mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;sghqRlOv&#x2F;493-convox-builds-export-import"> convox builds export &#x2F; import</a>

api/controllers/builds.go
```diff
@@ -266,6 +273,32 @@ func BuildCopy(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	return RenderJson(rw, b)
 }
 
+// BuildExport creats an artifact, representing a build, to be used with another Rack
```
frewsxcv: typo: "creats"

cmd/convox/builds.go
```diff
@@ -342,40 +367,35 @@ func cmdBuildsImport(c *cli.Context) error {
...
-	if c.Bool("id") {
-		fmt.Print(build.Release)
+	fmt.Fprintf(out, "\nRelease: %s\n", build.Release)
 
-	} else {
-		fmt.Println()
-		fmt.Printf("Imported %s\n", build.Id)
-		fmt.Printf("To deploy this copy run `convox releases promote %s --app %s`\n", build.Release, app)
+	if c.Bool("id") {
+		fmt.Println(build.Release)
```
MiguelMoll: Nitpicking as the end result is the same but for consistency we could use `fmt.Println(build.Release)` as well in `cmdBuildsCreate()` instead of `os.Stdout.Write([]byte(fmt.Sprintf("%s\n", build)))`. Same naming scheme for `out := os.Stdout` too. This isn't a big deal.

PR #1192, opened 2016-09-06
Fix Created Unknown for AWS::ECS::Service

ddollar:


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;IrLS4XOW&#x2F;671-get-rid-of-created-unknown-during-install"> get rid of “Created Unknown” during install</a>

PR #1196, opened 2016-09-07
Fix memcached security group port

ddollar:
Was fixed in the template but not in the compiled assets.


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;JHj5WkpR&#x2F;666-hammer-corp-pro-memcached-service-created-with-incorrect-port-in-security-group"> hammer-corp (pro): Memcached service created with incorrect port in security group</a>

ianmalott:
Confirmed `SecurityGroupIngress` is now using 11211 instead of 6379 for new memcached services in my development rack.

PR #1203, opened 2016-09-08
Handle a v2 or v1 manifest

MiguelMoll:
Let the Load() handle the different versions. Otherwise it would be:

v1
`yaml.Unmarshal([]byte(b.Manifest), &m.Services)`

v2
`yaml.Unmarshal([]byte(b.Manifest), &m)`

Deleteing images for a v1 manifest would return nil since Services would be empty


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;4NB2LiBF&#x2F;676-delete-image-fix"> Delete image fix</a>

PR #1205, opened 2016-09-08
add skew and logging to cron

### LINKED ISSUE #1164: scheduled jobs fail to start due to AWS ThrottlingException
(splitting the scheduled jobs issue out from #1133)

I've seen scheduled jobs fail to start because the ecs-agent itself gets a throttling exception while pulling the container from ECR. For jobs that only run daily, this means another 24 hours until it retries. My hunch was this happens most often during timeframes where multiple scheduled jobs are firing, but he ecs-agent logs show a completely idle agent before attempting to pull the container.

I tried asking ECS for help resolving this as it seemed like a bug on their end (either ecs-agent or ECR), but they've now pushed back and said the lambda that starts the task should have a retry built into it to cover this case, so we're back in Convox-land.

Here's their reply:

> 2016-08-25T12:00:04Z [INFO] Adding event module="eventhandler" change="ContainerChange: arn:aws:ecs:us-east-1:XXX:task/2b5f1779-0765-4d87-8f11-5ee8e29f657a run -> STOPPED, Reason CannotPullECRContainerError: ThrottlingException: Rate exceeded
        status code: 400, request id: 75567eb3-6abb-11e6-9d4c-cb4ccd1cee97, Known Sent: NONE"
> 
> Is related to the GetAuthorizationToken call in the ECS agent. The current TPS limit is 1 TPS with 30 TPS burst for fetching a new token. This isn't a limit that can be changed at the moment, but we're looking in making improvements so we can increase it, as well as cache tokens in the ECS agent.
> 
> We would recommend adding backoff and retry in your run task from the lambda call for the moment.

/cc @nzoschke

ddollar:
This adds a bit of logging for cron jobs and causes their start time to be skewed by up to 10 seconds to avoid a thundering herd of RunTasks and the resulting API throttling.

fixes #1164


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;l8CXIATE&#x2F;680-goodeggs-basic-scheduled-jobs-failing-due-to-aws-throttling-exceptions"> goodeggs (basic): Scheduled Jobs failing due to AWS Throttling Exceptions</a>

PR #1207, opened 2016-09-09
Remove the check/set for AWS_REGION

MiguelMoll:
The --region flag already checks AWS_REGION for a value. Checking the env variable directly will override the --region value.


MiguelMoll:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;NOS8pt7l&#x2F;685-add-new-env-var-region-for-convox-install"> Add new env var region for convox install</a>

PR #1209, opened 2016-09-09
Log output streaming

ddollar:
Get logs out as fast as possible without trying to ensure that they are in the right order. We could potentially add client-side optional sorting or something like that.

Currently only apps with lots of log streams (frequently crashing processes or frequent cron) will have potentially weirdly out-of-order logs.


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;lB7QCijZ&#x2F;687-log-streaming"> log streaming</a>

PR #1226, opened 2016-09-14
store un-interpolated manifest on the release

### LINKED ISSUE #1225: regression: docker-compose environment variables being altered
#1044 introduced environment variable interpolation for docker-compose.yml, but it has caused a regression in my production apps that contain `$` characters in their `environment` variables.

before, this docker-compose.yml worked:

```yaml
web:
  image: foo
  command: bar
  environment:
    - SEKRET=foo$bar
```

but after #1044 this is sent to CloudFormation as `SEKRET=foo`.

I'd argue applying interpolation to the raw YAML string as a whole is too naive a solution, and it should instead be applied to specific properties post-parsing.  Else, it'd be awesome to make a bunch of noise about such upgrade issues -- I'm now going through and double-quoting all of our environment variables...

bobzoller:
fixes #1225

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -57,10 +57,10 @@ func main() {
 
 	writeDockerAuth()
 
-	m, err := manifest.LoadFile(fmt.Sprintf("src/%s", manifestPath))
+	data, err := ioutil.ReadFile(fmt.Sprintf("src/%s", manifestPath))
 	handleError(err)
 
-	data, err := m.Raw()
+	m, err := manifest.Load(data)
 	handleError(err)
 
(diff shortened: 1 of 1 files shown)


mattmanning:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;bzbGUjN2&#x2F;713-docker-compose-yml-env-interpolation-issues"> docker-compose.yml env interpolation issues</a>

ddollar:
Fixed in #1231

bobzoller:
awesome @ddollar, #1231 looks amazing, can't wait.

PR #1230, opened 2016-09-15
autoscale down slowly

ddollar:
This prevents the autoscaler from stopping more than one instance at a time.


contra:
This has a pretty negative impact on the speed of deploys, doesn't it?

PR #1237, opened 2016-09-16
remove version check on rack api

ddollar:
This removes the required `Version` header requirement on the Rack API.

### DIFF
--- a/api/controllers/api.go
+++ b/api/controllers/api.go
@@ -30,13 +30,6 @@ func api(at string, handler ApiHandlerFunc) http.HandlerFunc {
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


MiguelMoll:
Is there a reason to not want to check a min version?

mwarkentin:
@MiguelMoll Was there an answer to your question above? Do we need to worry about older-than-expected clients causing issues after this change?

ddollar:
We are addressing deprecation at the API level (403 errors that tell you to upgrade your client when you try to use something that no longer exists)

mwarkentin:
@ddollar Perfect, thanks!

PR #1239, opened 2016-09-16
Make instance lifecycle handler smarter

ddollar:
Rebuilds the instance lifecycle handler in Go and makes it much smarter. When an instance is scheduled for termination in an AutoScalingGroup we now:

* Deregister the instance from the ECS cluster (stops all tasks and schedules them elsewhere)
* Deregister the instance from all Rack load balancers
* Wait for the instances to drain from the load balancers
* Continue the instance termination


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;3uBa8VDk&#x2F;724-app-and-rack-api-downtime-during-instance-rolls"> App and Rack API downtime during instance rolls</a>

ddollar:
We probably need to change the order here a bit and remove the instances from the ELBs before we deregister them from the ECS cluster.

PR #1242, opened 2016-09-20
Correctly use existing values during stack updates

ddollar:
This fixes a case where stack updating would break when faces with a new parameter as it tries to `UsePreviousValue` for it.


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;peupWdGC&#x2F;746-correctly-use-existing-stack-vars"> correctly use existing stack vars</a>

MiguelMoll:
Changes lgtm. I want to run it through the different upgrade paths first.

PR #1243, opened 2016-09-20
Use filesystem events wrapper to find changes rather than polling

ddollar:
Uses https://github.com/rjeczalik/notify to get filesystem watchers that work recursively via native filesystem event handlers on each OS.

Drops average CPU usage of `convox start` from 25% to 0% on my iMac.


ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;1XDRvWbW&#x2F;744-code-sync-using-large-amounts-of-cpu"> Code sync using large amounts of CPU</a>

MiguelMoll:
Seems a vim plugin (still determining which) was causing the errors I saw above. Not worth blocking on this release. :+1:

kaaloo:
system monitor_002

I'm still getting some wild cpu sheninigans going with basically nothing going on in the app except waiting for a request.  The change in the cpu profile is when shutting down the app.

kaaloo:
system monitor_003

This time with --no-sync.  All is well!

PR #1248, opened 2016-09-22
Remove monitor

beedub:
The monitor is killing instances in the cluster on the first signal of an ecs agent disconnection. According to samuelkarp, the ecs agent disconnections are normal.


ddollar:
Trying to grab this for a current release but it didn't rebase cleanly. Will try on the next pass.

ddollar:
Rebased in #1255

PR #1252, opened 2016-09-23
Code sync updates

ddollar:
* Reverts file sync handler to a polling scanner.
* Factors out a handler that decides when to scan the directory next.
* Default handler scans the directory every 700ms trying to keep CPU usage to a minimum.
* Adds a MacOS-specific handler that waits for an FSEvent event to rescan the directory.
* Adds a Linux-specif handler that waits for an inotify event to rescan the directory.

Thanks to @nathany for doing most of the work!

Maybe @MiguelMoll (or other Linux user) will feel inspired to drop an inotify-based version on us :)


changes/watch_darwin.go
```diff
@@ -0,0 +1,84 @@
+// heavily inspired by https://github.com/fsnotify/go.fsevents
```
nathany: I hope it's not too heavily inspired by go.fsevents, because that code doesn't have an open source license. See https://github.com/fsnotify/go.fsevents/pull/7#issuecomment-63198985. 

Inspiration from https://github.com/fsnotify/fsevents would be better :-)

nathany: I should probably remove that repo from the fsnotify organization so it doesn't trip anyone up.

ddollar: I updated the comment to point to the other repo as this is largely influenced by the ideas in this wrapper but is not an outright copy. Thanks!

nathany: cool

nathany: Thanks David. If you have ideas on how the FSEvents wrapper could be improved, please do chime in. Pull requests are also welcome.

changes/watch_linux.go
```diff
@@ -0,0 +1,53 @@
+package changes
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"golang.org/x/exp/inotify"
```
nathany: Just FYI, fsnotify is based on x/exp/inotify, but more than likely has fixed a few bugs over the years. I have been thinking of splitting out a github.com/fsnotify/inotify or github.com/fsnotify/fsnotify/inotify package at some point. I'll be sure to let you know if that ever gets done. :-)

PR #1254, opened 2016-09-23
read default cpu and memory scale from manifest

### LINKED ISSUE #353: Support mem_limit in docker-compose.yml
```yaml
main:
  mem_limit: 1024m
  build: .
  ports:
    - 2000:2000
```

Something like that should be the same as a `convox deploy` followed by `convox scale main --memory=1024`

ddollar:
Reads the default CPU and Memory settings for a process from the `docker-compose.yml`.

```
web:
  cpu_shares: 100
  mem_limit: 20MB
```

Closes #353

### DIFF
--- a/api/models/release.go
+++ b/api/models/release.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


ddollar:
[Default memory limit](https://trello.com/c/10iOELfe/695-default-memory-limit)

ddollar:
@MiguelMoll I can't reproduce the problem you found. The `*Formation` parameters are being updated fine for me as long as they have never been set before.

PR #1255, opened 2016-09-23
[WIP] remove instance monitor from agent

ddollar:
Rebase of #1248

### TODO

- [ ] instead of removing completely let's find a way to kill an instance that has been disconnected for more than N minutes


ddollar:
blocked on @nzoschke 's input

nzoschke:
I definitely want to get rid of code, and get stable instances. The observations from disabling the monitor it on a production cluster are good, and probably enough to turn it off.

My main concern is that a known failure scenario -- an instance is running but not participating in the cluster -- has some automatic replacement story. Perhaps the EC2 / ASG health checks offer it?

beedub:
I just checked on our dev rack and it's been running without the monitor without any known issues. The last termination event for the related ASG was Sept 19th.

Our dev cluster has 10 instances and many deployments per day of multiple apps (much more so than our prod cluster).

This behavior is in contrast to 1-2 daily terminations with the monitor turned on.

I'd feel comfortable removing the agent-check on our rack.

That being said, new AMIs and new ecs-agent versions may introduce instability.

ddollar:
Closing for now. Will re-evaluate.

mwarkentin:
@ddollar After doing some further investigation, I think this might be related to the issues we've been seeing as well (have a support ticket opened for it).

I'm seeing the following in our rack logs:

```
2016-10-18T16:08:32Z web:20161006183008/4e93a1642a61 who="convox/monitor" what="marked instance i-072306e7c9c8abf58 unhealthy" why="ECS reported agent disconnected"
2016-10-18T16:13:32Z web:20161006183008/4e93a1642a61 who="convox/monitor" what="marked instance i-0ad5299eee3307320 unhealthy" why="ECS reported agent disconnected"
2016-10-18T16:18:32Z web:20161006183008/4e93a1642a61 who="convox/monitor" what="marked instance i-033e422c85951c53f unhealthy" why="ECS reported agent disconnected"
2016-10-18T16:18:32Z web:20161006183008/4e93a1642a61 who="convox/monitor" what="marked instance i-08bc2db4db91e1eaa unhealthy" why="ECS reported agent disconnected"
```

beedub:
Just curious, why is this closed for re-evaluation? The monitor is only causing turbulence with no benefit according to behavior we see in our racks. While it'd be ideal for there to be a more sophisticated monitor (an idea I saw thrown around), removing it completely for now seems like a better option than leaving it.

ddollar:
In my obeservation over the last two weeks I have seen 2 instances disconnect their agent and never reconnect. This leaves the cluster in a bad state.

I definitely agree that I'd rather not churn the instances if possible but it's better than making the cluster "lock up" some of its capacity. I'm definitely open to looking at a more sophisticated agent that waits a little longer before terminating an instance.

mwarkentin:
I think some sort of increased amount of time is important - this has been causing us all sorts of problems with any sort of rack update that requires a new set of instances (rack version updates, AMI changes, changing instance type). It seems (though I don't have hard numbers) that these updates would fail 40-50% of the time after hitting the 15m timeout because 3-4 instances in a row were started and then marked as unhealthy by the `convox/monitor` process due to the ECS agent being disconnected. Sometimes the rollback would fail for the same reason, and we would have to manually trigger a re-rollback.

Our workaround for now is going to be disabling `convox/monitor` before running any sort of rack update, but that's pretty hacky.

mwarkentin:
It looks like it also terminated 10 prod instances in an hour last night (we weren't doing anything with the rack - unsure how many of these actually made it into the rack vs. just being terminated after launch):

```
$ convox rack logs --filter why --since 12h --rack prod | grep disconnected
2016-10-19T08:57:27Z web:20161006183008/3bd775cade6a who="convox/monitor" what="marked instance i-0923c54d237f89abe unhealthy" why="ECS reported agent disconnected"
2016-10-19T09:02:27Z web:20161006183008/3bd775cade6a who="convox/monitor" what="marked instance i-0abf44a0a99084b58 unhealthy" why="ECS reported agent disconnected"
2016-10-19T09:07:27Z web:20161006183008/3bd775cade6a who="convox/monitor" what="marked instance i-09b056533d0008827 unhealthy" why="ECS reported agent disconnected"
2016-10-19T09:12:27Z web:20161006183008/3bd775cade6a who="convox/monitor" what="marked instance i-00cfd8196fcb03474 unhealthy" why="ECS reported agent disconnected"
2016-10-19T09:17:27Z web:20161006183008/3bd775cade6a who="convox/monitor" what="marked instance i-05a365c86303b8eb0 unhealthy" why="ECS reported agent disconnected"
2016-10-19T09:22:27Z web:20161006183008/3bd775cade6a who="convox/monitor" what="marked instance i-03331cfd11a501746 unhealthy" why="ECS reported agent disconnected"
2016-10-19T09:27:27Z web:20161006183008/3bd775cade6a who="convox/monitor" what="marked instance i-0e94d698ee4cbbd02 unhealthy" why="ECS reported agent disconnected"
2016-10-19T09:32:27Z web:20161006183008/3bd775cade6a who="convox/monitor" what="marked instance i-08b0a28343ce0eead unhealthy" why="ECS reported agent disconnected"
2016-10-19T09:37:27Z web:20161006183008/3bd775cade6a who="convox/monitor" what="marked instance i-06a4e1f43137ff496 unhealthy" why="ECS reported agent disconnected"
2016-10-19T09:42:27Z web:20161006183008/3bd775cade6a who="convox/monitor" what="marked instance i-02516120c813898d4 unhealthy" why="ECS reported agent disconnected"
```

It's also weird that these are all exactly 5 minutes apart.. not sure what might be the cause of that.

PR #1261, opened 2016-09-27
parameterize sqs

ddollar:


nzoschke:
[Setting properties for SQS service](https://trello.com/c/mD6BZ8d9/723-setting-properties-for-sqs-service)

provider/aws/templates/service/sqs.tmpl
```diff
@@ -2,6 +2,27 @@
   {
     "AWSTemplateFormatVersion" : "2010-09-09",
     "Parameters": {
+      "MessageRetentionPeriod": {
```
nzoschke: I don't see where these parameters are passed through to the resource creation.

ddollar: doh! good catch, fixed

PR #1271, opened 2016-09-28
Add generic Docker skeleton for detected Python apps (`convox init`)

soulshake:
Test suite passes, but still needs manual verification.


cmd/convox/templates/init/python/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
+version: "2"
+services:
+  web:
+    build: .
+    labels:
+      - convox.port.443.protocol=tls
+      - convox.port.443.proxy=true
```
mattmanning: This line enables PROXY protocol. Will the gunicorn backend be able to handle that? We might need to run something in front. In our rails image, for example we use nginx to read the PROXY protocol headers and set HTTP headers for the app backend:

https://github.com/convox/rails/blob/master/conf/nginx.conf#L35-L38
https://github.com/convox/rails/blob/master/bin/web#L6-L7

cmd/convox/templates/init/python/docker-compose.yml
```diff
@@ -0,0 +1,16 @@
...
+    links:
+      - database
+    ports:
+      - "80:5000"
```
mattmanning: Is it typical for flask backends to listen on ports 5000 and 8000?

mattmanning:
[Add generic Docker skeleton for detected Python apps (`convox init`)](https://trello.com/c/pHMAviur/781-add-generic-docker-skeleton-for-detected-python-apps-convox-init)

cmd/convox/templates/init/python/docker-compose.yml
soulshake: Port 5000 is the default, according to the [flask.Flask.run() doc](http://flask.pocoo.org/docs/0.11/api/#flask.Flask.run).

Port 8000, no.

However, this isn't meant to be Flask-specific, only Python-specific, and I believe 5000/8000 for http/https is common for Python apps.

ddollar:
This needs rebase, etc

soulshake:
[Deal with `convox init` Python PR](https://trello.com/c/MpYbboDo/1437-deal-with-convox-init-python-pr)

PR #1280, opened 2016-09-30
wait for services to converge before continuing instance termination

ddollar:


ddollar:
[App and Rack API downtime during instance rolls](https://trello.com/c/3uBa8VDk/724-app-and-rack-api-downtime-during-instance-rolls)

PR #1297, opened 2016-10-04
move registries to provider interface

ddollar:
Moves registries into the provider interface.

* `docker login` is now only ever called during `convox build`
* App servers no longer try to login to private registries on boot
* Solves a token expiration issue that was showing up after API instances had been up for a few hours


provider/mock_Provider.go
```diff
@@ -0,0 +1,1130 @@
+package provider
+
+import io "io"
+import mock "github.com/stretchr/testify/mock"
+import structs "github.com/convox/rack/api/structs"
+
+// MockProvider is an autogenerated mock type for the Provider type
+type MockProvider struct {
```
MiguelMoll: +1 for MockProvider

client/registries.go
```diff
@@ -23,34 +18,41 @@ func (c *Client) AddRegistry(server, username, password, email string) (*Registr
...
+	system, err := c.GetSystem()
 	if err != nil {
 		return nil, err
 	}
 
-	return &registry, nil
-}
+	// backwards compatible
+	// FIXME: pick the right version here
+	if system.Version < "30000000000000" {
```
MiguelMoll: Acceptable to go out like this?

api/main.go
```diff
@@ -17,12 +13,5 @@ func recoverWith(f func(err error)) {
 }
 
 func main() {
-	// prime the API and instance for builds by logging into private registries
-	// and pulling down latest app images
-	go func() {
-		models.LoginPrivateRegistries()
-		models.PullAppImages()
```
MiguelMoll: Build cache will be handled elsewhere?

client/registries.go
ddollar: Yes, I'll fix it in a quick CLI release once I have a release number for this release.

api/main.go
ddollar: It's handled by Docker. Pulling the images on the API instance doesn't necessarily help anymore anyway.

Issue #856, opened 2016-07-09
convox doctor


ianmalott:
Homebrew has one of my favorite commands: `brew doctor`.

```
❯ brew doctor
Please note that these warnings are just used to help the Homebrew maintainers
with debugging if you file an issue. If everything you use Homebrew for is
working fine: please don't worry and just ignore them. Thanks!

Warning: python is symlinked to python3
This will confuse build scripts and in general lead to subtle breakage.

Warning: Your Homebrew is outdated.
You haven't updated for at least 24 hours. This is a long time in brewland!
To update Homebrew, run `brew update`.
```

If the Convox CLI had a similar diagnostic tool, it could warn users about common issues. For example...

- Is docker installed?
- Are basic convox config files (Dockerfile, docker-compose.yml, .dockerignore) present?
- Is your rack up to date?
- Is your CLI up to date?
- Is the app directory large enough to potentially slow uploads?
- Do your rack versions (e.g. staging vs. production) match?

Would love to hear if others would find this feature useful. Are these kinds of checks too basic to add much value? Are docs good enough? Conversely, would this have made getting started with Convox easier? Any other checks you'd like to see included?

For the Convox team, what are some of our most common issues, and could we help users avoid them with this type of diagnostic command?

nzoschke:
I **love** this idea.

For a while there was `convox debug` which also helped pull a bunch of internal AWS logs / events for support purposes. It would be rad if that came back as `convox doctor myapp`.

mwarkentin:
That would be great if it could check the Cloudformation status and dump out the events if it's in a failed / rolled back state, rather than having to log in and find it in the console.

ianmalott:
`convox doctor` is in the CLI now.

PR #1300, opened 2016-10-07
Doctor

awsmsrc:
This pr implements a "doctor" cli command that should help diagnose issues with applications new to the convox platform and identify ways in which applications don't conform to the convox philosophy


nzoschke:
I pushed some more checks. @awsmsrc I could use help with a testing strategy, as merging this is blocked on codecov.

ianmalott:
Addresses #856

PR #1301, opened 2016-10-10
Optionally specify a service and or command with convox start 

awsmsrc:
This patch provides support for specifying a specific service to start within an application and additionally/optionally a particular command to run in the following format

```
convox start
convox start api
convox start api rake db:migrate
```

The main motivation is for one off tasks such as migrations and data seeding tasks.
It was decide to add these to start as opposed to adding something like a --local flag to run as in the future we may want a "local" rack "provider" and I didn't want to pollute the run namespace/command/api

### DIFF
--- a/cmd/convox/start.go
+++ b/cmd/convox/start.go
@@ -3,6 +3,7 @@ package main
 import (
 	"errors"
 	"fmt"
+	"log"
 	"os"
 	"os/exec"
 	"os/signal"
@@ -45,6 +46,10 @@ func init() {
 
 func cmdStart(c *cli.Context) error {
 	// go handleResize()
+	log.Printf("%#v", c.Args())
+	if len(c.Args()) == 1 {
+		//startProc
+	}
 
 	id, err := currentId()
 	if err != nil {
--- a/cmd/convox/start.go
+++ b/cmd/convox/start.go
@@ -19,7 +19,7 @@ func init() {
 	stdcli.RegisterCommand(cli.Command{
 		Name:        "start",
 		Description: "start an app for local development",
-		Usage:       "[directory]",
+		Usage:       "[service] [command]",
 		Action:      cmdStart,
 		Flags: []cli.Flag{
 			cli.StringFlag{
--- a/cmd/convox/start.go
+++ b/cmd/convox/start.go
@@ -77,6 +77,13 @@ func cmdStart(c *cli.Context) error {
 		return stdcli.ExitError(err)
 	}
 
+	if service != "" {
+		_, ok := m.Services[service]
+		if !ok {
+			return stdcli.ExitError(fmt.Errorf("Service %s not found in manifest", service))
+		}
+	}
+
 	if err := m.Shift(c.Int("shift")); err != nil {
 		return stdcli.ExitError(err)
 	}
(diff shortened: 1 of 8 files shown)


mwarkentin:
@awsmsrc Any thoughts on if this could potentially handle something like https://github.com/convox/rack/issues/827 as well?

awsmsrc:
@mwarkentin i do believe this would solve that problem, I don't think we want to pass any config other then the command to run though

manifest/build.go
```diff
@@ -6,11 +6,16 @@ import (
 	"strings"
 )
 
-func (m *Manifest) Build(dir, appName string, s Stream, cache bool) error {
+func (m *Manifest) Build(targetService, dir, appName string, s Stream, cache bool) error {
```
ddollar: Can we factor out a BuildService instead? This params list is getting rather unwieldy.

manifest/run.go
```diff
@@ -14,6 +14,9 @@ import (
 )
 
 type Run struct {
+	TargetService string
```
ddollar: Feels like there should be a better way to proliferate this.

cmd/convox/start.go
```diff
@@ -86,7 +96,7 @@ func cmdStart(c *cli.Context) error {
 	cache := !c.Bool("no-cache")
 	sync := !c.Bool("no-sync")
 
-	r := m.Run(dir, app, cache, sync)
+	r := m.Run(targetService, targetCommand, dir, app, cache, sync)
```
ddollar: could factor out a RunService here:
```
r := m.RunService(service, dir, app, RunOptions{
  Cache: true,
  Sync: true,
})
```

cmd/convox/start.go
```diff
@@ -45,6 +45,16 @@ func init() {
 
 func cmdStart(c *cli.Context) error {
 	// go handleResize()
+	var targetService string
+	var targetCommand []string
```
ddollar: These can just be `service` and `command`. The extra modifier `target` isn't really adding anything here.

cmd/convox/start.go
awsmsrc: we could simply do run and add service to the run options, that way it work exactly as it does now and we wouldn't need to broaden the manifest API

awsmsrc:
@ddollar i didn't factor out a BuildService RunService func but I added service and command to an options struct.

My thinking was that if we make a change to how we run a service we would have to make that change in two places.

awsmsrc:
[Convox start proc command](https://trello.com/c/bCjIeiDS/816-convox-start-proc-command)

ddollar:
If you factor this out you would implement Run like this:

```
func Run() {
  for _, s := range Services {
    RunService(s)
  }
}
```

No code duplication, factored out.

awsmsrc:
@ddollar just looked at this (with @MiguelMoll) in order to implement your suggestions, it's still not 100% clear to me where to make these improvements when I consider all the context. I'm sure you're on to something, would love to have a ten minute pairing session on this. Functionally this does work right now though which is good enough for workshop.

Note to self: One thing I just noticed in testing is we still need to add an error check for a missing service

awsmsrc:
Cache dependency inverted cc @ddollar

PR #1306, opened 2016-10-11
Parameterizing Postgres Version

sandeep:


ddollar:
Carrying forward in #1308

PR #1314, opened 2016-10-17
Proper error message when no racks are present

oss92:
If no racks were found the message used to say "Try one of: " and then an empty line.

screen shot 2016-10-17 at 18 58 37


oss92:
Rebased!

PR #1325, opened 2016-10-19
Client improvements for export and import

MiguelMoll:
The goal here is to to stream data with Reader/Writer interfaces instead of loading everything into memory.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -2,7 +2,6 @@ package client
 
 import (
 	"bufio"
-	"bytes"
 	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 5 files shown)


MiguelMoll:
[Client improvements for export and import](https://trello.com/c/xFAnqusH/864-client-improvements-for-export-and-import)

client/client.go
```diff
@@ -143,75 +141,102 @@ type PostMultipartOptions struct {
...
-
-		br = NewProgressReader(br, opts.Progress.Progress)
+			pr = NewProgressReader(r, opts.Progress.Progress)
+		}
 	}
 
-	req, err := c.request("POST", path, br)
+	writer := multipart.NewWriter(w)
+	streamErr := make(chan error)
+	go func() {
```
MiguelMoll: Since we are using an `io.Pipe` (so that files aren't loaded completely into memory) `CreateFormFile` blocks using the Pipe writer. By having this logic in a goroutine, the request can start reading from the pipe and the goroutine can write to it.

ddollar:
Looks good, please rebase

PR #1337, opened 2016-10-22
Ensure no downtime during deplyments and instance roll with autoscale enabled

MiguelMoll:
Fix includes changes to the lifecycle handler to deregister the instance from the cluster and draining the elb connections. Also adding a "converging" status for SystemGet() so that the monitor knows not to terminate instances unless status == running.

### DIFF
--- a/provider/aws/system_test.go
+++ b/provider/aws/system_test.go
@@ -32,6 +32,27 @@ func TestSystemGet(t *testing.T) {
 	}, s)
 }
 
+func TestSystemGetConverging(t *testing.T) {
+	provider := StubAwsProvider(
+		cycleSystemDescribeStacks,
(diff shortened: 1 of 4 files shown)


MiguelMoll:
[App and Rack API downtime during instance rolls](https://trello.com/c/3uBa8VDk/724-app-and-rack-api-downtime-during-instance-rolls)

provider/aws/system_test.go
```diff
@@ -14,6 +14,8 @@ import (
 func TestSystemGet(t *testing.T) {
 	provider := StubAwsProvider(
 		cycleSystemDescribeStacks,
+		cycleDescribeRackStackResources,
```
ddollar: Seems like it might be easy to add a test for the "converging" version.

PR #1339, opened 2016-10-23
Add support for the new us-east-2 (Ohio) region.

jtvhk:
Updated REGIONS and app.tmpl to include us-east-2.  👍


ddollar:
Could you try running `make -C api templates` in the root directory and including that in the PR please? If this makes any test fixtures fail you can delete them and run `make fixtures` to regenerate them.

ddollar:
Carrying forward in #1341

PR #1341, opened 2016-10-24
Add support for us-east-2

ddollar:
Replaced #1339


awsmsrc:
Note to self: This will require a console change too

PR #1356, opened 2016-10-27
add a lock around channel map writes

ddollar:


MiguelMoll:
Dupe of https://github.com/convox/rack/pull/1382

PR #1357, opened 2016-10-27
Fix rack status precedence

MiguelMoll:
Might be hard to tell from the diff, but this change just wraps the ASG check around an `if` to see if the stack is running. Otherwise all other stack status take precedence.


MiguelMoll:
[Rack status precedence](https://trello.com/c/KWvTgDEe/922-rack-status-precedence)

mwarkentin:
@MiguelMoll Ignoring whitespace makes this diff a lot easier to read: https://github.com/convox/rack/pull/1357/files?w=1

:)

PR #1366, opened 2016-11-02
Make 'convox start' respect mem_limit and cpu_shares

soulshake:
Currently, convox only reads `cpu_shares` and `mem_limit` from `docker-compose.yml` when a user runs `convox deploy`. It doesn't read those keys when running locally via `convox start`.

This PR is intended to make `convox start` read and respect those values.

**Point of confusion regarding `mem_limit`:** Earlier this week, I was under the impression that convox needed this to be an integer (i.e. without a text label), and treated it as MB. But while testing, it seems like convox can in fact convert unit labels? If someone could help me understand the current behavior, that would be much appreciated. (Edit: Maybe it's just the fact that I'm [appending `MB` here](https://github.com/convox/rack/blob/25ec6602eb5ffeef33c57237af751cadf0234255/manifest/process.go#L96)).

**Feedback needed:** I attempted to add a test for `mem_limit`, and would like to get some feedback before adding a test for `cpu_shares`.

FYI: These keys `cpu_shares` and `mem_limit` are read only once, during the first run of `convox deploy`; subsequent deploys do not re-read them if they are changed in `docker-compose.yml`. That's not addressed in this PR, though.)

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -128,14 +128,10 @@ func (m Manifest) Validate() []error {
 			}
 		}
 
-		// TODO test cpu_shares
-
-		// test mem_limit
-		// FIXME: Do we accept text labels or not?
-		mem_as_str := fmt.Sprintf("%#v", entry.Memory)
-		mem, err := strconv.Atoi(mem_as_str)
-		if err != nil || (mem < 4 && mem != 0) || mem > 10000 {
-			errors = append(errors, fmt.Errorf("%s has invalid mem_limit %s: should be a number in MB without a text unit label between 4 and 10000", entry.Name, mem_as_str))
+		// test mem_limit: Docker requires a mem_limit of at least 4mb (or 0)
+		mem_min := Memory(4194304)
+		if entry.Memory < mem_min && entry.Memory != 0 {
+			errors = append(errors, fmt.Errorf("%s has invalid mem_limit %#v: should be either 0, or at least %#v", entry.Name, entry.Memory, mem_min))
 		}
 	}
 
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -13,6 +13,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/docker/go-units"
 	"gopkg.in/yaml.v2"
 )
 
@@ -129,9 +130,15 @@ func (m Manifest) Validate() []error {
 		}
 
 		// test mem_limit: Docker requires a mem_limit of at least 4mb (or 0)
-		mem_min := Memory(4194304)
-		if entry.Memory < mem_min && entry.Memory != 0 {
-			errors = append(errors, fmt.Errorf("%s has invalid mem_limit %#v: should be either 0, or at least %#v", entry.Name, entry.Memory, mem_min))
+		mem_min := Memory(units.MB * 4)
+		mem := entry.Memory
+
+		if mem < mem_min && mem != 0 { //Memory(0) {
+			e := fmt.Errorf("%s has invalid mem_limit %#v: should be either 0, or at least %#vMB",
+								entry.Name,
+								mem / units.MB,
+								mem_min / units.MB)
+			errors = append(errors, e)
 		}
 	}
 
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -127,6 +127,16 @@ func (m Manifest) Validate() []error {
 				errors = append(errors, fmt.Errorf("%s links to service: %s which does not expose any ports", entry.Name, l))
 			}
 		}
+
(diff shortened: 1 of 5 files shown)


soulshake:
[Respect `mem_limit` and `cpu_shares` when running `convox start`](https://trello.com/c/DCIFERbD/943-respect-mem-limit-and-cpu-shares-when-running-convox-start)

MiguelMoll:
Seems the confusion come from the [docs](https://convox.com/docs/docker-compose-file#memory-limit) not being inline with the [implementation](https://github.com/convox/rack/blob/cpu-and-mem-in-convox-start/manifest/yaml.go#L247-L258).

`mem_limit` can take an int or a string (`10MB`) and UnmarshalYAML will handle it accordingly with the `Memory` field ultimately being in bytes.

Suggested doc changes:

- unit size is in bytes unless specified
- text labels are supported
- perhaps mention that the minimum is 4*MB* which is a docker limit

manifest/process.go
```diff
@@ -87,6 +88,14 @@ func NewProcess(app string, s Service, m Manifest) Process {
 		args = append(args, "-v", volume)
 	}
 
+	if s.Cpu != 0 {
+		args = append(args, "--cpu-shares",  strconv.FormatInt(s.Cpu, 10))
+	}
+
+	if s.Memory != 0 {
+		args = append(args, "--memory", fmt.Sprintf("%#vMB", s.Memory))
```
MiguelMoll: The MB isn't necessary here as we can pass the s.Memory as is.

manifest/manifest.go
```diff
@@ -127,6 +127,16 @@ func (m Manifest) Validate() []error {
 				errors = append(errors, fmt.Errorf("%s links to service: %s which does not expose any ports", entry.Name, l))
 			}
 		}
+
+		// TODO test cpu_shares
+
+		// test mem_limit
+		// FIXME: Do we accept text labels or not?
```
MiguelMoll: We do accept labels 👍

manifest/manifest.go
```diff
@@ -127,6 +127,16 @@ func (m Manifest) Validate() []error {
...
+		mem_as_str := fmt.Sprintf("%#v", entry.Memory)
+		mem, err := strconv.Atoi(mem_as_str)
+		if err != nil || (mem < 4 && mem != 0) || mem > 10000 {
+			errors = append(errors, fmt.Errorf("%s has invalid mem_limit %s: should be a number in MB without a text unit label between 4 and 10000", entry.Name, mem_as_str))
```
MiguelMoll: error message can be rewritten as we do accept text labels.

manifest/manifest.go
```diff
@@ -127,6 +127,16 @@ func (m Manifest) Validate() []error {
 				errors = append(errors, fmt.Errorf("%s links to service: %s which does not expose any ports", entry.Name, l))
 			}
 		}
+
+		// TODO test cpu_shares
+
+		// test mem_limit
+		// FIXME: Do we accept text labels or not?
+		mem_as_str := fmt.Sprintf("%#v", entry.Memory)
+		mem, err := strconv.Atoi(mem_as_str)
```
MiguelMoll: I believe this conversion is to check the value is between with 4 and 10,000.  Since Memory is really an int64, it can be used directly `(entry.Memory < 4 && entry.Memory != 0)`

manifest/manifest.go
```diff
@@ -127,6 +127,12 @@ func (m Manifest) Validate() []error {
 				errors = append(errors, fmt.Errorf("%s links to service: %s which does not expose any ports", entry.Name, l))
 			}
 		}
+
+		// test mem_limit: Docker requires a mem_limit of at least 4mb (or 0)
+		mem_min := Memory(4194304)
+		if entry.Memory < mem_min && entry.Memory != 0 {
+			errors = append(errors, fmt.Errorf("%s has invalid mem_limit %#v: should be either 0, or at least %#v", entry.Name, entry.Memory, mem_min))
```
MiguelMoll: I forgot to mention in my initial pass, this error should state explicitly what's going on.

For example entry.Memory while in bytes, should be displayed in "human readable" format in the error.
`mem := units.HumanSize(float64(entry.Memory))`
And while technically true, do we need to mention 0 as a valid value?

`At least 4MB is required for web.  Found:  "2.097 MB"`
Something along those lines.

MiguelMoll:
Small update to simplify/clarify the error. Otherwise, this change is good to go. 👍

soulshake:
[Document mem_limit & cpu_shares](https://trello.com/c/OLWaMMSn/1154-document-mem-limit-cpu-shares)

PR #1370, opened 2016-11-03
Use `LINK_PORT` variable if it's set

mwarkentin:
This brings PORT up to parity with SCHEME, PATH, USERNAME, and PASSWORD.

### Use Case

We are trying to use the official [`rabbitmq`](https://hub.docker.com/_/rabbitmq/) image for local development.

* That image [`EXPOSE`s multiple ports](https://github.com/docker-library/rabbitmq/blob/master/Dockerfile#L81)
* Convox uses the [first](https://github.com/convox/rack/blob/9ea46f4a3dfe20a9186f964640e7675b925e97d6/manifest/service.go#L289) [one](https://github.com/convox/rack/blob/9ea46f4a3dfe20a9186f964640e7675b925e97d6/manifest/service.go#L275-L279) exposed in `docker inspect`
* It doesn't provide a way to set it explicitly like the other `LINK_*` environment variables can be used

### TODO

- [x] Update docs: https://github.com/convox/site/pull/340


mwarkentin:
Came across these `LINK_*` references as well - not sure how they fit into things yet:

* https://github.com/convox/rack/blob/master/api/models/release.go#L392
* https://github.com/convox/rack/blob/master/api/models/fixtures/web_redis.yml
* https://github.com/convox/rack/blob/master/api/models/fixtures/web_redis.json

mwarkentin:
My main question is around what needs to be done in [release.go](https://github.com/convox/rack/blob/master/api/models/release.go#L483-L489).

* Do we need to support this both locally and on the rack? My use case is currently local only.
* Do we need to support both public ('`8000:8000`) and private (`8000`) ports?
* If so, would it make sense to write a function in [`ports.go`](https://github.com/convox/rack/blob/master/manifest/ports.go) which parses a port string and returns a `Port` instance?

mwarkentin:
Confirmed that this change works locally:

docker-compose.yml:

```yaml
worker:
  build: .
  command: /app/bin/worker
  environment:
    - ...
  links:
    - broker

broker:
  image: rabbitmq:3.6
  environment:
    - LINK_SCHEME=amqp
    - LINK_PORT=5672
  ports:
    - 5672
```

```
/Users/michaelwarkentin/go/src/github.com/convox/rack/cmd/convox/convox start

...

worker   │ running: docker run -i --rm --name notifications-worker ... --add-host broker:172.17.0.2 -e BROKER_SCHEME=amqp -e BROKER_HOST=172.17.0.2 -e BROKER_PORT=5672 -e BROKER_PATH= -e BROKER_USERNAME= -e BROKER_PASSWORD= -e BROKER_URL=amqp://172.17.0.2:5672 --add-host redis:172.17.0.4 ... notifications/worker sh -c /app/bin/worker

**Comment by ddollar** [comment]
Seems good. Yes, we should do it for production also, that would be here:

https://github.com/convox/rack/blob/master/api/models/release.go#L489

**Comment by mwarkentin** [comment]
@ddollar Thanks. I took a look at adding it there, but wasn't sure what to do for converting an environment variable (string) into a `Port` object: https://github.com/convox/rack/blob/master/manifest/ports.go#L5-L10

It's a little less clear to me how this would be used on a rack (as I'd expect the only option is to go through the service's ELB, which already has the exposed ports defined.

**Comment by ddollar** [comment]
I'm assuming this LINK_PORT defines which port shows up by default in the full link _URL ?

**Comment by ddollar** [comment]
Perhaps loop through `other.Ports` looking for the one where `LINK_PORT` matches `.Container`

**Comment by mwarkentin** [comment]
@ddollar I'll give that a shot, thanks!

File: api/models/release.go [hunk 1]
```diff
@@ -488,6 +489,20 @@ func (r *Release) resolveLinks(app App, manifest *manifest.Manifest) (*manifest.
 
 			port := other.Ports[0]
 
+			if other.Exports["LINK_PORT"] != "" {
+				i, _ := strconv.Atoi(other.Exports["LINK_PORT"])
+				if err != nil {
+					// handle error
+					fmt.Println(err)
```
mwarkentin: Golang noobie.. not sure what to do here in case of an error converting `LINK_PORT` to an `int`.

api/models/release.go
```diff
@@ -488,6 +489,20 @@ func (r *Release) resolveLinks(app App, manifest *manifest.Manifest) (*manifest.
...
+				}
+
+				for _, p := range other.Ports {
+					if i == p.Container {
+						port = p
+					}
+				}
```
mwarkentin: Anything we should do if they've set `LINK_PORT` but it doesn't match any of the exposed ports?

ddollar: That should be an error.

api/models/release.go
ddollar: `return nil, err`

api/models/release.go
```diff
@@ -18,7 +19,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/convox/rack/api/crypt"
 	"github.com/convox/rack/api/structs"
-	"github.com/convox/rack/manifest"
+	mp "github.com/convox/rack/manifest"
```
mwarkentin: Had to do this as I wasn't able to refer to `manifest.Port` in my code otherwise. Open on suggestions for a better way to handle this (or better name for it - `mp` stands for manifest package).

I updated other the other references that I found as well.

ddollar: So let's go the other way with this to avoid surprises around the namespace. Perhaps name the variable `m` down in the function and let this be `manifest`

mwarkentin: Can do, thanks!

api/models/release.go
```diff
@@ -389,8 +390,8 @@ func (r *Release) Formation() (string, error) {
 	return app.Formation(*manifest)
 }
 
-func (r *Release) resolveLinks(app App, manifest *manifest.Manifest) (*manifest.Manifest, error) {
-	m := *manifest
+func (r *Release) resolveLinks(app App, m *manifest.Manifest) (*manifest.Manifest, error) {
+	mp := *m
```
mwarkentin: Ran into one more existing variable name conflict here - renamed the pointer to the manifest `mp` and updated other references.

mwarkentin:
Updated docs here: https://github.com/convox/site/pull/340

mwarkentin:
@MiguelMoll Not sure if you saw this one, @ddollar mentioned that it could make it into the next release potentially?

MiguelMoll:
@mwarkentin Getting the release ready 👍 Could you update this branch with master? Then we'll be ready to go.

mwarkentin:
@MiguelMoll Updated, thanks!

mwarkentin:
@MiguelMoll Thanks!

PR #1381, opened 2016-11-09
Enable lambda VpcConfig for a syslog service

MiguelMoll:
If running in a private Rack.

- [ ] Document new syslog behavior. (https://github.com/convox/site/pull/337)


provider/aws/service.go
```diff
@@ -47,6 +47,7 @@ func (p *AWSProvider) ServiceCreate(name, kind string, params map[string]string)
 		}
 		req, err = p.createService(s)
 	case "syslog":
+		s.Parameters["IsPrivate"] = fmt.Sprintf("%t", p.SubnetsPrivate != "")
```
ddollar: should just be `Private` for consistency

PR #1385, opened 2016-11-10
paging support for describing stacks

MiguelMoll:
If the output from `DescribeStacks` exceeds 1 MB in size, we wouldn't have all the stacks in Cloudformation. This change adds next token ("paging") support.


MiguelMoll:
[Convox app no longer visible in any list views](https://trello.com/c/XiaLe4mD/1085-convox-app-no-longer-visible-in-any-list-views)

Issue #1349, opened 2016-10-25
env variables set before a deploy but not promoted are not picked up by the deploy


bisrael:
Scenario:

set an env variable needed in your next deploy:
`convox env set MYVAR=1`

deploy your app:
`convox deploy --description="Uses MYVAR"`

then you'll notice in the task definition in EC2 that `MYVAR` is not in the task's environment

mwarkentin:
I think we ran across this recently as well - I believe it used to work the other way around? I assume it was during the builds refactor that this behaviour changed?

bisrael:
it's certainly not the behavior i would expect

bladealslayer:
This certainly causes some trouble when new app is created, if the app needs some env variables to be present in order to start properly. Without the variables, the app is stuck in updating state, because containers keep exiting.

mattmanning:
![](https://github.trello.services/images/mini-trello-icon.png) [Deploying doesn't apply new env vars](https://trello.com/c/crP1xvK0/1114-deploying-doesn-t-apply-new-env-vars)

ddollar:
This is no longer the case

PR #1386, opened 2016-11-10
create release using provider API

MiguelMoll:
This fix would address the #1349 where new environment variables aren't picked up in the next deploy.


MiguelMoll:
[Deploying doesn't apply new env vars](https://trello.com/c/crP1xvK0/1114-deploying-doesn-t-apply-new-env-vars)

PR #1389, opened 2016-11-11
Configure ASG lifecycle for instance launch

mwarkentin:
* Define a new `autoscaling:EC2_INSTANCE_LAUNCHING` transition
* Mark lifecycle as complete at end of UserData right before sending CFN success signal

This should help resolve the issue with instances being marked unhealthy immediately after launching if you have a bootstrap script running which takes a couple of minutes.

The problem was that instances are being moved into `InService` state in the ASG before their UserData has completed - and the `convox/monitor` gets a list of `InService` instances from the ASG and compares them to the registered instances in ECS to determine if the ECS agent is connected.

If you have a multiple minute bootstrap script, the ECS agent is often not started before the monitor runs (currently every 5 minutes).

If we configure an ASG lifecycle for instance launch, it shouldn't move into `InService` until the UserData has completed and the ECS agent can launch.

### DIFF
--- a/provider/aws/dist/rack.json
+++ b/provider/aws/dist/rack.json
@@ -1031,6 +1031,7 @@
               "  - ", { "Ref": "InstanceRunCommand" }, "\n"
               ] ] }
             ] },
+            "  - /usr/bin/aws autoscaling complete-lifecycle-action --lifecycle-hook-name InstancesLifecycleStarting --auto-scaling-group-name ", {"Ref":"Instances"}," --lifecycle-action-result CONTINUE\n"
             "  - /opt/aws/bin/cfn-signal --stack ", { "Ref": "AWS::StackName" }, " --region ", {"Ref":"AWS::Region"}, " --resource Instances\n"
           ] ] }
         }
@@ -1092,6 +1093,14 @@
           ],
           "WaitOnResourceSignals": "true"
         }
+      }
+    },
+    "InstancesLifecycleStarting": {
+      "Type": "AWS::AutoScaling::LifecycleHook",
+      "Properties": {
+        "AutoScalingGroupName": { "Ref": "Instances" },
+        "HeartbeatTimeout": "900",
+        "LifecycleTransition": "autoscaling:EC2_INSTANCE_LAUNCHING"
       }
     },
     "InstancesLifecycle": {
--- a/provider/aws/dist/rack.json
+++ b/provider/aws/dist/rack.json
@@ -1032,7 +1032,7 @@
               ] ] }
             ] },
             "  - export INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)\n",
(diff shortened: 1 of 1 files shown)


provider/aws/dist/rack.json
```diff
@@ -1094,6 +1095,14 @@
         }
       }
     },
+    "InstancesLifecycleStarting": {
+      "Type": "AWS::AutoScaling::LifecycleHook",
+      "Properties": {
+        "AutoScalingGroupName": { "Ref": "Instances" },
+        "HeartbeatTimeout": "900",
```
mwarkentin: 15 minutes, same as our cloudformation timeout for an instance launch.

provider/aws/dist/rack.json
```diff
@@ -1031,6 +1031,7 @@
               "  - ", { "Ref": "InstanceRunCommand" }, "\n"
               ] ] }
             ] },
+            "  - /usr/bin/aws autoscaling complete-lifecycle-action --lifecycle-hook-name InstancesLifecycleStarting --auto-scaling-group-name ", {"Ref":"Instances"}," --lifecycle-action-result CONTINUE\n"
```
mwarkentin: Not sure if this name is correct, I just took it from the JSON key below.

provider/aws/dist/rack.json
```diff
@@ -1031,6 +1031,7 @@
               "  - ", { "Ref": "InstanceRunCommand" }, "\n"
               ] ] }
             ] },
+            "  - /usr/bin/aws autoscaling complete-lifecycle-action --lifecycle-hook-name InstancesLifecycleStarting --auto-scaling-group-name ", {"Ref":"Instances"}," --lifecycle-action-result CONTINUE\n",
```
mwarkentin: Not sure if this name is correct, I just took it from the JSON key below.

mwarkentin:
@MiguelMoll @ddollar What's the best way to test this? Can I just manually drop `rack.json` into the cloudformation template for a rack in the AWS console?

mwarkentin:
@ddollar @MiguelMoll So last night one of our prod instances was terminated.. and then 182 more were terminated until I disabled the monitor:



It would be nice if we could get this out quickly if we can verify it fixes the issue.

MiguelMoll:
Ran into a circular dependency between launch config's UserData referencing the autoscaling group's name before it can be created.

mwarkentin:
@MiguelMoll Ah, that makes sense. Wonder if it's easiest to grab that from the instance metadata store then instead if that's available..

mwarkentin:
http://serverfault.com/a/654856

MiguelMoll:
[Lifecycle on instance creation](https://trello.com/c/PrRvjzTt/1151-lifecycle-on-instance-creation)

mwarkentin:
I was able to get the ASG name like so - will look at updating the PR shortly: 
```
[ec2-user@convox-kraken--i-00892f788edd17145 ~]$ IID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id); aws autoscaling describe-auto-scaling-instances --instance-ids=$IID --query 'AutoScalingInstances[0].AutoScalingGroupName' --output text --region us-east-1
convox-kraken-Instances-1ST9C55WDGV44
```

MiguelMoll:
Finishing it via https://github.com/convox/rack/pull/1392 Thanks for the PR 👍

PR #1391, opened 2016-11-14
WIP: Fix missing `convox start` errors

ianmalott:
Given an env var `FOO` in docker-compose.yml that is not set in .env, `convox start` will fail silently rather than printing an error hinting at the cause of the problem. It used to print `env expected: FOO`.


soulshake:
Fixed in #1422

PR #1394, opened 2016-11-15
Clean up old builds

MiguelMoll:
To void hitting ECR [limits](http://docs.aws.amazon.com/AmazonECR/latest/userguide/service_limits.html), keep the 50 newest builds while deleting anything older.

### DIFF
--- a/provider/aws/builds.go
+++ b/provider/aws/builds.go
@@ -71,6 +71,26 @@ func (p *AWSProvider) BuildCreate(app, method, url string, opts structs.BuildOpt
 		},
 	}, nil)
 
+	// AWS currently has a limit of 1000 images in ECR
+	// This is a "hopefully temporary" and brute force means
+	// to prevent hitting limits during deployment
+	bs, err := p.BuildList(app, 150)
+	if err != nil {
+		fmt.Printf("Error listing builds for cleanup: %s\n", err.Error())
+	} else {
+		if len(bs) >= 50 {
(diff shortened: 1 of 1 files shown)


mwarkentin:
Is this different than some of the PRs from earlier this year? Eg. https://github.com/convox/rack/pull/978

MiguelMoll:
This change and #978 is kind-of related in that they both introduce clean up "work". 978 cleans up resource when an app is deleted, etc. This PR cleans up old builds, keeping the latest 50 only. This is to avoid hitting ECR image limits which a few users have hit recently.

MiguelMoll:
[Users hitting 1000-image limit](https://trello.com/c/AEZJGXYO/1079-users-hitting-1000-image-limit)

provider/aws/builds.go
```diff
@@ -71,6 +71,26 @@ func (p *AWSProvider) BuildCreate(app, method, url string, opts structs.BuildOpt
 		},
 	}, nil)
 
+	// AWS currently has a limit of 1000 images in ECR
+	// This is a "hopefully temporary" and brute force means
+	// to prevent hitting limits during deployment
+	bs, err := p.BuildList(app, 150)
+	if err != nil {
+		fmt.Printf("Error listing builds for cleanup: %s\n", err.Error())
```
ddollar: Let's not error the whole build if this fails.

PR #1395, opened 2016-11-15
Rack instance count must be >= 3

MiguelMoll:
Enforce minimum of 3 Rack instances instead of 2.

A Rack with an instance count <= 2 will have to bump the instance count to at least 3 before updating.

### DIFF
--- a/api/controllers/system.go
+++ b/api/controllers/system.go
@@ -56,8 +56,8 @@ func SystemUpdate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 			return httperr.Errorf(403, "scaling count prohibited when autoscale enabled")
 		case c == -1:
 			// -1 indicates no change
-		case c <= 1:
-			return httperr.Errorf(403, "count must be greater than 1")
+		case c <= 2:
+			return httperr.Errorf(403, "count must be greater than 2")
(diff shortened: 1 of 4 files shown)


MiguelMoll:
[Enforce minimum of 3 rack instances (instead of 2)](https://trello.com/c/5Jvg5DQn/828-enforce-minimum-of-3-rack-instances-instead-of-2)

matthewford:
@MiguelMoll the trello board is private - so not sure why this was enforced, we have one client where three instances is a bit overkill for their application. We'll have to increase it, but wondering why this was needed?

MiguelMoll:
Sure. Since Rack runs two copies of the Rack API process for optimal uptime, if a Rack doesn't have at least three instances updates can't roll out and your CloudFormation stack will get wedged. 

When Cloudformation tries to rollback the update, the limit of two instances prevents the rollback from succeeding. We've seen a few users getting stuck because of this scenario recently.

matthewford:
@MiguelMoll fair enough, could the API not set with a minimum of 1 process running? and an optimal of 2

MiguelMoll:
That is an option we considered but didn't go down that route just yet. Could be an option in the future.

PR #1397, opened 2016-11-16
Add last line to docker error

MiguelMoll:
The previous error returned could be a vague "exit status 1". This way we get a more detailed error message.


MiguelMoll:
Tagging along this PR to get the related ECR fix in.

Login into ECR without protocol scheme:
Issue with newer docker versions if https is used via login causes
push/pull commands to fail. Fixed slated for docker 1.13.0

- docker/docker#22910
- docker/docker#23100 (comment)

MiguelMoll:
@awsmsrc Thoughts? ^

Issue #1169, opened 2016-08-30
add/fix ability to update service versions


contra:
Example: Updating a Postgres on RDS from 9.4.7 -> 9.5.2

contra:
For an added level of severity, you can't manually update versions on convox created instances.

Example of trying to update a postgres instance on RDS:

<img width="845" alt="screen shot 2016-08-30 at 3 17 10 pm" src="https://cloud.githubusercontent.com/assets/425716/18103497/df6fb364-6ec4-11e6-80fb-5122d1f30061.png">

yields

`Current Parameter Group (stae-production-municipal-data-parametergroup-xs8coft2mf7p) is non-default. You need to explicitly specify a new Parameter Group in this case (default or custom) (Service: AmazonRDS; Status Code: 400; Error Code: InvalidParameterCombination; Request ID: 2d530ebc-6ee6-11e6-9fc1-c16dbcd2ec7c)`

lisajluo:
You can actually manually upgrade if you change the Parameter Group, but subsequent attempts to update the upgraded service via convox cli will fail.

contra:
Anything I can do to help out w/ this? If I had a pointer or two on where to start I can attempt a PR on this.

PG 9.6 just released, really eager to be able to use that.

contra:
Any chance this can be added to the `Services` milestone?

contra:
This looks like it might be possible now - can anyone on the team confirm?

jordansexton:
This is still not possible, I just ran into attempting to upgrade from 9.4.7 to 9.5.7

```
convox version
client: 20170819183623
server: 20170918191932 (console.convox.com)
```

PR #1398, opened 2016-11-16
update postgres version to 9.6.1

contra:
https://www.postgresql.org/about/news/1717/


contra:
To avoid having to do these in the future - https://github.com/convox/rack/issues/1169

MiguelMoll:
Thanks for the PR @contra 👍 . When you get a chance, update the branch with `master` so we can get this in the next release.

MiguelMoll:
Taking this PR forward via https://github.com/convox/rack/pull/1412 Thanks for the PR @contra 👍

PR #1406, opened 2016-11-17
Extract one-off process name

MiguelMoll:
Check for slice size and replace string with method. This way we don't assume slice sizes.


MiguelMoll:
[`convox ps` panic when running one-off](https://trello.com/c/uwOSnMmN/1175-convox-ps-panic-when-running-one-off)

PR #1407, opened 2016-11-17
syslog Private parameter defaults to false

MiguelMoll:


mattmanning:
[Set default for Private parameter on syslog stacks](https://trello.com/c/ILV70nc0/1208-set-default-for-private-parameter-on-syslog-stacks)

provider/aws/templates/service/syslog.tmpl
```diff
@@ -19,6 +19,7 @@
   "Parameters": {
     "Private": {
         "Description": "Flag stating if Rack is priavte",
```
mwarkentin: Fix this typo while you're in here? `priavte` -> `private`

PR #1415, opened 2016-11-21
Run cron jobs via convox start

MiguelMoll:
Services that have a `convox.cron` label should have the job ran via `convox start` (Try to take advantage of manifest package)

- [x] Service links should use services that are already running


MiguelMoll:
[Local cron functionality with `convox start`](https://trello.com/c/DQquBZDP/1247-local-cron-functionality-with-convox-start)

PR #1417, opened 2016-11-22
Stack outputs

mwarkentin:
Per discussion in support ticket (wave#88), I've enabled outputs for everything that is not sensitive (password and AWS credentials, basically).

### Use Case

Enable external tools to hook in and reference configuration settings of our convox stack.

Most useful for us at this time:

* Subnet IDs
* Security groups
* NAT Gateway IPs
* ASG ID
* VPC ID


MiguelMoll:
@mwarkentin When you get a chance, update this branch with master so we can release it. :shipit:

mwarkentin:
@MiguelMoll Updated w/ latest master, thanks!

mwarkentin:
Thanks!

PR #1425, opened 2016-11-29
Fix disconnected agent

gmelika:
waits for missed heartbeats before terminating an instance

### DIFF
--- a/api/workers/cluster.go
+++ b/api/workers/cluster.go
@@ -67,6 +67,7 @@ func StartCluster() {
 
 				if !seenBefore {
 					disconnectedInstances[i.Id] = struct{}{}
+					fmt.Printf("who=\"convox/monitor\" what=\"instance %s missed it's first heartbeat\" why=\"ECS reported agent disconnected\"\n", i.Id)
 					continue
 				}
 				// Not registered or not connected => set Unhealthy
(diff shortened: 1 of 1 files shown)


mwarkentin:
Awesome! Would it make any sense to make this a configurable parameter on the rack? If it can count the number of times it's seen an instance, then it could be set to `2` which would be equivalent to this, or a higher number if it turns out that we need or want something else?

mwarkentin:
Also, if a heartbeat is missed, but then another is successful, it should probably "reset" the counter. If I miss a heartbeat, and then miss one more 60m later, the instance shouldn't be terminated (not sure if this would happen here, but I think so).

gmelika:
I think a rack param makes sense.  I'm unfamiliar with how this is done in convox; if someone can provide some pointers, I can add it

mwarkentin:
Perfect, wasn't sure if that `delete()` call was what was terminating the instance, thanks!

mwarkentin:
Rack parameters are defined here: https://github.com/convox/rack/blob/master/provider/aws/dist/rack.json#L164-L341

Not sure what the mechanism for passing that through to get to your code would be.

nzoschke:
Thanks for the patch! I'll cut a testing release now.

I hope we don't have to parameterize this. I'd like to find the permanent fix over adding another behavior that only a couple people would ever know to tune.

gmelika:
I have tested this and it is working as expected as well.

PR #1432, opened 2016-12-02
optionally show all processes running in the rack cluster

awsmsrc:


mwarkentin:
@awsmsrc Awesome, I think this will be helpful.

Any thoughts on if it would make sense to indicate which instance a process is running on as well? 

Use case:

We're running our rack on T2s, and sometimes we have an instance that has a bunch of processes running on it which is using up its credits, and we have another instance with barely any processes. I was trying to figure out which process for an app was running on the busy service, so I could stop it and (maybe) reschedule it onto another instance, but I couldn't map the convox process IDs to the processes in the AWS console to figure out which is which.

ddollar:
@mwarkentin I'd prefer to keep instance ID out of the actual CLI command but if you need to get to it you can call the REST API (convox api get /apps/NAME/processes)

The convox process IDs are the first chunk of the ECS Task ID

api/controllers/system.go
```diff
@@ -25,7 +25,9 @@ func SystemShow(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 }
 
 func SystemProcesses(rw http.ResponseWriter, r *http.Request) *httperr.Error {
-	ps, err := models.Provider().SystemProcesses()
+	all := r.URL.Query().Get("all")
+
+	ps, err := models.Provider().SystemProcesses(all == "true")
```
ddollar: Let's use an options struct here. See other provider methods like `BuildCreate` for an example.

client/system.go
```diff
@@ -60,11 +63,10 @@ func (c *Client) GetSystemCapacity() (*SystemCapacity, error) {
 	return &capacity, nil
 }
 
-func (c *Client) GetSystemProcesses() (Processes, error) {
+func (c *Client) GetSystemProcesses(all bool) (Processes, error) {
```
ddollar: Same here, let's make an options struct.

cmd/convox/ps.go
```diff
@@ -90,10 +90,10 @@ func cmdPs(c *cli.Context) error {
 }
 
 func displayProcesses(ps []client.Process) {
-	t := stdcli.NewTable("ID", "NAME", "RELEASE", "STARTED", "COMMAND")
+	t := stdcli.NewTable("ID", "APP", "NAME", "RELEASE", "STARTED", "COMMAND")
```
ddollar: We should not display the APP column when listing processes for a single app.

provider/aws/system.go
```diff
@@ -102,21 +104,38 @@ func (p *AWSProvider) SystemLogs(w io.Writer, opts structs.LogStreamOptions) err
...
-	if err != nil {
-		return nil, err
-	}
+func (p *AWSProvider) SystemProcesses(all bool) (structs.Processes, error) {
+	var tasks []string
+	var err error
 
-	fmt.Printf("tasks = %+v\n", tasks)
+	if all {
+		tres, err := p.ecs().ListTasks(&ecs.ListTasksInput{
```
ddollar: probably need to paginate this

mwarkentin:
Interesting, thanks!

cmd/convox/rack.go
```diff
@@ -69,6 +70,10 @@ func init() {
 						Name:  "stats",
 						Usage: "display process cpu/memory stats",
 					},
+					cli.BoolFlag{
+						Name:  "a, all",
```
MiguelMoll: Might not be a big deal, but `-a` is used throughout the cli for `--app`. Perhaps to avoid confusion (even though this is used in the rack subcommand), just stick to `--all`?

ddollar: I think `-a` is OK here since it's under the `convox rack` namespace. `ps -a` feels like a natural command to me.

MiguelMoll:
[convox rack ps --all](https://trello.com/c/t3pE6XEJ/1309-convox-rack-ps-all)

PR #1433, opened 2016-12-05
fix output var formatting for env

ddollar:


ddollar:
[output var formatting for dev env](https://trello.com/c/B3sOE3Vf/1322-output-var-formatting-for-dev-env)

provider/aws/dist/rack.json
```diff
@@ -156,7 +156,7 @@
         { "Ref": "ExistingVpc" }
       ] }
     },
-    "VPCCIDR": {
+    "Vpccidr": {
```
MiguelMoll: All caps was causing the weird underscores?

MiguelMoll:
Anything to expect with removing the registry?

provider/aws/dist/rack.json
ddollar: Yep

PR #1435, opened 2016-12-05
Return stdcli.Error(err) instead of stdcli.QOSEventSend(...).

soulshake:


soulshake:
[Restore more missing CLI errors](https://trello.com/c/2aL2SUrM/1312-restore-more-missing-cli-errors)

mattmanning:
Looks like your changes will have the intended effect, but this makes me wonder, did `QOSEventSend` used to return the text of the error message? If so, was it intentionally changed or is that a regression? Having to add all of these explicit message returns where we used to have error messages has a bit of a funny smell to me.

MiguelMoll:
After looking at the code again, QOSEventSend in particular, I believe it’s correct. QOSEventSend didn’t change, it was the cli logging commands that did. Hence why we now return stdcli.Error everywhere theres an error

soulshake:
[Fix missing `convox start` errors](https://trello.com/c/5opE8u0I/973-fix-missing-convox-start-errors)

cmd/convox/install.go
```diff
@@ -631,6 +659,7 @@ func FriendlyName(t string) string {
 func waitForAvailability(url string) error {
 	tick := time.Tick(10 * time.Second)
 	timeout := time.After(20 * time.Minute)
+	distinctID, err := currentId()
```
MiguelMoll: Should check for error here along with QOS event.

cmd/convox/install.go
```diff
@@ -651,6 +680,7 @@ func waitForAvailability(url string) error {
 		}
 	}
 
+	stdcli.QOSEventSend("cli-waitForAvailability-unknown", distinctID, stdcli.QOSEventProperties{Error: err})
```
MiguelMoll: This call shouldn't be needed if moved to line 662

cmd/convox/install.go
```diff
@@ -229,22 +241,24 @@ func cmdInstall(c *cli.Context) error {
 
 	versions, err := version.All()
 	if err != nil {
-		return stdcli.QOSEventSend("cli-install", "", stdcli.QOSEventProperties{Error: fmt.Errorf("error getting versions")})
+		stdcli.QOSEventSend("cli-install", "", stdcli.QOSEventProperties{Error: fmt.Errorf("error getting versions: %s", err)})
```
MiguelMoll: Is `distinctID` left out on purpose?

cmd/convox/install.go
```diff
@@ -229,22 +241,24 @@ func cmdInstall(c *cli.Context) error {
...
+		return stdcli.Error(err)
 	}
 
 	version, err := versions.Resolve(c.String("version"))
 	if err != nil {
-		return stdcli.QOSEventSend("cli-install", "", stdcli.QOSEventProperties{Error: fmt.Errorf("error resolving version")})
+		stdcli.QOSEventSend("cli-install", "", stdcli.QOSEventProperties{Error: fmt.Errorf("error resolving version: %s", err)})
```
MiguelMoll: Same here, no `distinctID`.

PR #1436, opened 2016-12-05
Minsize count

MiguelMoll:
This change makes the desired and minimum instance count equal. The logic that was done in the lifecycle hook has been moved to system get information with the converging status.

Also removes unused code.


MiguelMoll:
[Prevent update from causing Rack downtime](https://trello.com/c/UybxhpBT/1308-prevent-update-from-causing-rack-downtime)

PR #1437, opened 2016-12-05
update deployment min to 50 for faster rollouts

ddollar:


ddollar:
[deployment min 50 for apps](https://trello.com/c/9mV3YF89/1320-deployment-min-50-for-apps)

PR #1439, opened 2016-12-06
Existing VPC updates

mattmanning:
Changes rack template to add Gateway Attachment, Routes, and Route Table Associations when installing into an existing VPC.

Updates the client to require passing an existing Internet Gateway gateway via a new `--internet-gateway` option.


mattmanning:
[Fix installing into an existing VPC](https://trello.com/c/VsGxzMJn/1318-fix-installing-into-an-existing-vpc)

PR #1430, opened 2016-12-02
Add additional cloudformation stack outputs

mwarkentin:
This builds on https://github.com/convox/rack/pull/1417 to add more
useful resources:

* AutoscalingGroup
* AvailabilityZones
* BalancerSecurityGroup
* Gateway
* GatewayAttachment
* IamRole
* InstanceProfile
* NatGateways
* NatIps

PR #1440, opened 2016-12-06
New stack outputs

MiguelMoll:
Taking PR #1430 forward.


mwarkentin:
Thanks for catching those issues! 😄

PR #1443, opened 2016-12-07
Enhancements for priveledged DataDog agent

nzoschke:
* Enables InstanceBootCommand to start an ECS tasks on every instance.
* Enables apps to mount /proc and /cgroup

This came up when working through the [DataDog ECS Integration doc](http://docs.datadoghq.com/integrations/ecs/):

> For Name enter docker_sock. For Source Path, enter /var/run/docker.sock.
> Add another volume with the name proc and source path of /proc/.
> Add another volume with the name cgroup and source path of /cgroup/.

> Create or Modify your IAM Policy
> If you are modifying the IAM Policy you created for your cluster, you may only need to add one Action: ecs:StartTask.

This does have security implications, so please review thoughtfully.


nzoschke:
[Priveledged Rack Apps Review](https://trello.com/c/KFUpWvO3/1344-priveledged-rack-apps-review)

provider/aws/dist/rack.json
```diff
@@ -971,6 +971,7 @@
                     "ecs:DiscoverPollEndpoint",
                     "ecs:Poll",
                     "ecs:RegisterContainerInstance",
+                    "ecs:StartTask",
```
ddollar: Is this still needed?

nzoschke: Not needed for the new approach of deploying the agent as a task. It feels right in general that an instance can manage its own tasks, but I'll remove it for sake of security.

PR #1444, opened 2016-12-07
Use Dockerfile ENV value if present

### LINKED ISSUE #1055: Code sync should resolve environment variables
A user reports:

```
ENV TARGET /usr/local/app 
COPY src $TARGET
```

Using the above, there will not be a /change agent registered for the target location. 
If you change the COPY to use the path directly, it will work.

MiguelMoll:
This change will allow code sync to use a Dockerfile `ENV` value if present in an `ADD` or `COPY` statement.

Fixes #1055

### DIFF
--- a/manifest/fixtures/Dockerfile-env-dir
+++ b/manifest/fixtures/Dockerfile-env-dir
@@ -1,0 +1,21 @@
+FROM convox/rails
+
+ENV DIR /app
+ENV FOO bar
+
+# copy only the files needed for bundle install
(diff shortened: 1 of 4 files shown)


MiguelMoll:
[Fix code sync env var support](https://trello.com/c/aSmtMZYM/1152-fix-code-sync-env-var-support)

MiguelMoll:
[Fix code sync env var support](https://trello.com/c/aSmtMZYM/1152-fix-code-sync-env-var-support)

PR #1447, opened 2016-12-07
Conditions added to the gateway stack outputs

MiguelMoll:
Also added new env var for private rack installs.

### DIFF
--- a/provider/aws/dist/rack.json
+++ b/provider/aws/dist/rack.json
@@ -104,9 +104,11 @@
       "Value": { "Ref": "EncryptionKey" }
     },
     "Gateway": {
+      "Condition": "BlankExistingVpc",
       "Value": { "Ref": "Gateway" }
     },
     "GatewayAttachment": {
+      "Condition": "BlankExistingVpc",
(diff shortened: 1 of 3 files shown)


provider/aws/dist/rack.json
```diff
@@ -104,9 +104,11 @@
       "Value": { "Ref": "EncryptionKey" }
     },
     "Gateway": {
+      "Condition": "BlankExistingVpc",
       "Value": { "Ref": "Gateway" }
     },
     "GatewayAttachment": {
+      "Condition": "BlankExistingVpc",
```
ddollar: Don't we make the Gateway attachment regardless of blank existing VPC?

MiguelMoll: I was using the same condition here: https://github.com/convox/rack/blob/fda3718f67981724f866149d1e5d293cf8fee130/provider/aws/dist/rack.json#L656

MiguelMoll:
[Fix `Gateway` output condition](https://trello.com/c/w8lOsISc/1348-fix-gateway-output-condition)

PR #1454, opened 2016-12-08
Added Canadian ca-central-1 region

idyll:
Not sure I got all the places, so please just tell me what's needed.

### DIFF
--- a/REGIONS
+++ b/REGIONS
@@ -1,6 +1,7 @@
 ap-northeast-1
 ap-southeast-1
 ap-southeast-2
+ca-central-1
 eu-central-1
 eu-west-1
 us-east-1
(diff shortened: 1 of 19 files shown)


MiguelMoll:
Thanks for the PR! Great stuff. I'll carry if forward. 👍

MiguelMoll:
Doh! Seems like `ca-central-1` doesn't support [Lambda](http://docs.aws.amazon.com/general/latest/gr/rande.html#lambda_region) just yet. Since Rack depends on it, we'll revisit adding `ca-central-1` when Lambda is ready to go.

Thanks again for the effort though. Greatly appreciated.

idyll:
It's missing:

 - [] EFS
 - [] Lambda

I will re-open as soon as they land.

PR #1456, opened 2016-12-09
Update Development.md and append a couple needed keys when writing to…

soulshake:
… .env.


bin/export-env
```diff
@@ -5,3 +5,7 @@ stack=$1
 [ -z $stack ] && echo "usage: export-env <stack>" && exit 1
 
 aws cloudformation describe-stacks --stack-name $stack | jq -r '.Stacks[0].Outputs[] | [(.OutputKey | gsub("(?<pre>.)(?<up>[A-Z])";"\(.pre)_\(.up)") | ascii_upcase), .OutputValue] | join("=")'
+
+echo "REGISTRY_HOST="
+echo "VPCCIDR="
```
MiguelMoll: This parameter was recently added (https://github.com/convox/rack/blob/master/provider/aws/dist/rack.json#L231) so export-env will grab it automatically.

PR #1457, opened 2016-12-09
Mem limit patch

soulshake:
Print mem_limit in both bytes and MB in error message.

If `docker-compose.yml` contains `mem_limit: 2048`, the current error is:

```
$ convox start
ERROR: counter has invalid mem_limit 0: should be either 0, or at least 4MB
```

New error will be:

```
$ convox start
ERROR: counter service has invalid mem_limit 2048 bytes (0 MB): should be either 0, or at least 4MB
```

edit: update format.

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -134,8 +134,9 @@ func (m Manifest) Validate() []error {
 		mem := entry.Memory
 
 		if mem < mem_min && mem != 0 { //Memory(0) {
-			e := fmt.Errorf("%s has invalid mem_limit %#v: should be either 0, or at least %#vMB",
+			e := fmt.Errorf("%s has invalid mem_limit %#v bytes / %#v MB: should be either 0, or at least %#vMB",
 				entry.Name,
+				mem,
 				mem/units.MB,
 				mem_min/units.MB)
 			errors = append(errors, e)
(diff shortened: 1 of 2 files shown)


soulshake:
[`mem_limit` unhelpful error](https://trello.com/c/S9xZlPRl/1222-mem-limit-unhelpful-error)

manifest/manifest.go
```diff
@@ -134,8 +134,9 @@ func (m Manifest) Validate() []error {
 		mem := entry.Memory
 
 		if mem < mem_min && mem != 0 { //Memory(0) {
-			e := fmt.Errorf("%s has invalid mem_limit %#v: should be either 0, or at least %#vMB",
+			e := fmt.Errorf("%s has invalid mem_limit %#v bytes / %#v MB: should be either 0, or at least %#vMB",
```
nzoschke: So the input is always in bytes? In that case I'd keep the validation message in bytes too. How about something like:

> invalid mem_limit 2048 bytes (0 MB). Should be either 0 bytes or at least 4000000 bytes (4 MB)"

nzoschke: yeah, print both. I was slightly confused by `200 bytes / 0 MB`, it felt like the `/` meant out of or something. So I recommend parenthesis.

nzoschke:
LGTM!

PR #1459, opened 2016-12-09
App name with dots

soulshake:
When inferring app name from dirname, replace dots with hyphens.
We should still fail when user specifies `--app foo.bar`, though.

I tried adding a test for the latter case, but it doesn't do what I expect:

```
--- FAIL: TestAppsCreateWithDotsInName (0.03s)
        Error Trace:    exec.go:35
	Error:		Not equal: 0 (expected)
			        != 1 (actual)
	Messages:	exit code should be equal
		
        Error Trace:    exec.go:37
	Error:		Not equal: "Creating app foo.bar... ERROR: app name can contain only alphanumeric characters, dashes and must be between 4 and 30 characters\n" (expected)
			        != "Creating app foo.bar... " (actual)
	Messages:	stdout should be equal
		
FAIL
```

Any hints on why the expected error message doesn't show up would be appreciated. 

Also, how would I go about adding a test when the app is inferred from the dirname, rather than specified via the `--app` flag?


ddollar:
The error message is probably in stderr and not stdout

soulshake:
[Replace dots with hyphens when app name inferred from current directory](https://trello.com/c/bDajHjcu/1405-replace-dots-with-hyphens-when-app-name-inferred-from-current-directory)

soulshake:
I'm guessing the failed Travis test is related to [this](https://github.com/convox/rack/blob/2a1b722b5f238cd4d4c616fe603e0c79f399352e/cmd/convox/apps_test.go#L61).

Not sure how to fix it though.

mattmanning:
[Document app name inference and app name restrictions](https://trello.com/c/MOCikIwa/1464-document-app-name-inference-and-app-name-restrictions)

PR #1465, opened 2016-12-12
Handle resources correctly for older existing VPC stacks

mattmanning:
Older Racks that were installed into an existing VPC won't have an InternetGateway parameter like the newer Racks will. Extra conditionals are needed to prevent creating AWS resources that depend on that parameter for Racks with ExistingVpc set.


mattmanning:
[Fix installing into an existing VPC](https://trello.com/c/VsGxzMJn/1318-fix-installing-into-an-existing-vpc)

PR #1470, opened 2016-12-13
Change sync Contains logic to check string paths

MiguelMoll:
Using `Rel()` in the sync's `Contains()` was causing issues with similar file/directory names. (e.g migration file vs migrations directory).

Originally `filepath.HasPrefix()` was used to compare paths based simply on strings, I thought using the same logic in `Contains()` would simplify it. (Side note: filepath.HasPrefix() just calls `strings.HasPrefix()` and only provided for backwards compatibility).

I'd really appreciate extra eyes here (especially @ddollar since he was the original author of this block) to see if I might be missing some case that isn't obvious to me.

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -57,23 +57,11 @@ func NewSync(container, local, remote string) (*Sync, error) {
 }
 
 func (s *Sync) Contains(t Sync) bool {
-	if !filepath.HasPrefix(t.Local, s.Local) {
+	if s.Local != t.Local {
 		return false
 	}
 
-	lr, err := filepath.Rel(s.Local, t.Local)
-
-	if err != nil {
-		return false
-	}
-
-	rr, err := filepath.Rel(s.Remote, t.Remote)
-
-	if err != nil {
-		return false
-	}
-
-	return lr == rr
+	return s.Remote == t.Remote
 }
 
 func (s *Sync) Start(st Stream) error {


MiguelMoll:
[code sync: directory name bug](https://trello.com/c/nEHLWliO/1301-code-sync-directory-name-bug)

soulshake:
I can confirm this fixes the issue for me, at least based on the [test I was using to reproduce it](https://github.com/soulshake/convox-code-sync-issue).

blockjon:
I am hoping this solves the problem I'm having - I would estimate that 50% of the time, files are not sync'ing as expected when I do a `convox start` which means its not currently reliable enough to use for development. If I can provide any additional insight for verification of a fix, for example, just let me know.

MiguelMoll:
Sorry to hear that @blockjon. I'd greatly appreciate any info you could share:

- How do the `ADD`/`COPY` statements look like in your Dockerfile? Could you share them?
- When sync *doesn't* work, would you happen to know how many files are supposed to be syncing? 5 files vs 100 files?
- How many files are syncing when it does work?
- How are the changes being made? For example, manual code changes or running a command (e.g `npm install foo`)

blockjon:
@MiguelMoll I have 9547 files in my project as it stands today. Unclear what makes it work... ill provide more details when I can find more. Haven't seen it work all day.

blockjon:
@miguelmoll Not sure if this has something to do with it but I'm trying to launch two containers that are mapped to the same directory on my mac. Using plain docker volume bind mounts has no problem with this.

blockjon:
@MiguelMoll Here's roughly the area where I move files into the container: ```WORKDIR /home/app

COPY requirements/prod.txt requirements/

ADD ./shortcuts /home/app/shortcuts

RUN ...

RUN ...

RUN ...
RUN ...
RUN ...

ADD package.json /tmp/package.json
RUN ...
RUN ...
RUN ...

COPY docker docker/
RUN ...

RUN ...

EXPOSE 80
EXPOSE 9090
EXPOSE 9010

RUN ...

COPY docker/static/web/.bashrc /root/.bashrc

ADD . /home/app/
```

MiguelMoll:
This is great intel @blockjon. Thanks! Unfortunately this PR won't do the trick. But this is still my top priority and hope to have a fix soon.

blockjon:
@MiguelMoll We literally cannot use convox in development with a working file syncer. :(

PR #1472, opened 2016-12-14
Bypass temp dir for incoming files

### LINKED ISSUE #1469: code sync crashes when /tmp is on a different partion(?)
I have `/tmp` on a ram disk and the code sync doesn't like that:

```
$ mount | fgrep /tmp
tmpfs on /tmp type tmpfs (rw,noatime)
$ uname -a
Linux easybox 4.3.0-1-amd64 #1 SMP Debian 4.3.3-7 (2016-01-19) x86_64 GNU/Linux
$ convox -v
client: 20161208160629
```

During `convox start` there's one log entry that looks like this:
```
convox       │ error: rename /tmp/foo.bar841595221 /home/richard/project/foo.bar: invalid cross-device link
```

and obviously nothing syncs

MiguelMoll:
There was an issue when renaming a temp file that was in a different partition/filesystem. This change writes directly to the target.

Fixes #1469

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -7,7 +7,6 @@ import (
 	"compress/gzip"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


MiguelMoll:
[code sync crashes when /tmp is on a different file system(?)](https://trello.com/c/FDuqBXVI/1365-code-sync-crashes-when-tmp-is-on-a-different-file-system)

PR #1473, opened 2016-12-14
Stop local containers on convox start error

### LINKED ISSUE #1462: Convox Start waits 5s before giving up on a container to start convox/proxy
I am not a go developer by any means; from what I see [here](https://github.com/convox/rack/blob/ba9ad1324b64c881521f09b173bc1a25a82ec42a/manifest/run.go#L160-L165) and [here](https://github.com/convox/rack/blob/0448b04268b9e7808a83037374ebc52cb77fd5b9/manifest/run.go#L234-L248) `convox start` will wait 5s before starting the convox/proxy for a container and in this case, fails to bring up convox/proxy without message.

I have a dev with an older laptop and a certain docker container we use takes 4-8 seconds to come up on his machine depending on the load.

Understandably this may be a poorly thought out docker container if it takes so long to come up but perhaps this could be a configuration option for older systems where 5s may be too optimistic?  Or at least provide a status message that reports convox/proxy was not started because dependent container took too long to start.

MiguelMoll:
Also increased the timeout on waiting for containers to start with an error message if needed: Fixes #1462

### DIFF
--- a/cmd/convox/start.go
+++ b/cmd/convox/start.go
@@ -127,6 +127,8 @@ func cmdStart(c *cli.Context) error {
 
 	err = r.Start()
 	if err != nil {
+		r.Stop()
+
 		stdcli.QOSEventSend("cli-start", id, stdcli.QOSEventProperties{
 			ValidationError: err,
 			AppType:         appType,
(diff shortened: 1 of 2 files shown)


MiguelMoll:
[convox/proxy doesn't stop when a container crashes and `convox start` exits](https://trello.com/c/JXEylIWR/1370-convox-proxy-doesn-t-stop-when-a-container-crashes-and-convox-start-exits)

PR #1483, opened 2016-12-16
make start scanner buffer 10mb

ddollar:
this fixes a crash when start sees a really long line


ddollar:
```version: 2
services:
  web:
    command: ruby -e 'puts "foo"*100000'
    image: convox/ruby
```

PR #1486, opened 2016-12-16
Return if no change in system count

MiguelMoll:


api/workers/autoscale.go
```diff
@@ -83,18 +83,20 @@ func autoscaleRack() {
...
+		} else {
+			return
 		}
 	} else {
 		system.Count = desired
 	}
 
-	log.Logf("change=%d", (desired - system.Count))
-
 	err = models.Provider().SystemSave(*system)
```
ddollar: Would be nice to remove the return above and just add a guard here to return if there are no changes.

We should also move the log back down here. I'd like to see the actual changes it's about to try to make after all of the adjustments made by the algorithm.

api/workers/autoscale.go
```diff
@@ -93,8 +95,12 @@ func autoscaleRack() {
 		system.Count = desired
 	}
 
-	log.Logf("change=%d", (desired - system.Count))
+	// nothing changed, return
+	if system.Count == oldCount {
+		return
+	}
 
+	log.Logf("change=%d", (system.Count - oldCount))
```
ddollar: Probably just switch the log to just above here. So we still see the change=0 log.

PR #1488, opened 2016-12-17
Set CF InternetGateway parameter based on CLI input

mattmanning:
The parameter was not being properly set, so the resulting stack was not being created correctly.

## Quick CLI Release Playbook
- [x] Code review
- [ ] Merge into master
- [ ] Release CLI


mattmanning:
[ExistingVPC CLI fix](https://trello.com/c/38DpBteY/1399-existingvpc-cli-fix)

PR #1489, opened 2016-12-18
[WIP] Prevent file sync from locking up when syncing large amount of files

MiguelMoll:
This change would address two issues when syncing 1000+ files. Docker
exec will error out if the argument list is too long so we batch files
to avoid that error.

Also prevent the go-docker client library from blocking when the exec
process has finished and retry on error.

Rough details on how sync works:
- sync is running in a loop keeping track of file changes (e.g new/edit/deleted file)
- snap shot of the tracked directories are taken to see what has changed
- after one second, any changes that have taken place acted on
   - upload/outgoing == changes from the host to the container
   - download/incoming == changes from container to the host (except changes to delete files are ignored)
- the `/changed` command keeps track of what changes and writes to stdout for sync to pick up on
- Steps when a file changes in a container:
    - sync batches the files that need to change and executes a tar command via docker exec process of the files that are read to stdout
    - sync reads the output and recreates the files on the host with said changes

### DIFF
--- a/sync/sync.go
+++ b/sync/sync.go
@@ -261,7 +261,7 @@ func (s *Sync) syncOutgoingAdds(adds []changes.Change, st Stream) {
 		remote := filepath.Join(s.Remote, a.Path)
 
 		s.lock.Lock()
-		s.incomingBlocks[a.Path] += 1
+		s.incomingBlocks[a.Path]++
 		s.lock.Unlock()
 
 		tgz.WriteHeader(&tar.Header{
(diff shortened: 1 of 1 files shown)


ddollar:
Looking really good. Would probably be good to add more comments describing how the sync process works now that you have your brain around it.

MiguelMoll:
[Files Not Syncing to Containers](https://trello.com/c/Wlo38pGT/1181-files-not-syncing-to-containers)

PR #1494, opened 2016-12-19
Disable OS File Sharing on Windows

nzoschke:
NFTS -> container is not working. Use normal docker volume.

### DIFF
--- a/manifest/process.go
+++ b/manifest/process.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"path/filepath"
 	"runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


manifest/process.go
```diff
@@ -107,14 +108,24 @@ func (p *Process) GenerateArgs(opts *ArgOptions) []string {
...
 		if !strings.Contains(volume, ":") {
-			home, err := homedir.Dir()
-			if err != nil {
-				log.Fatal(err)
+			home := ""
+			var err error
+
+			switch runtime.GOOS {
+			case "windows":
+				home = "/tmp" // prefix with /tmp to use Docker Volume
```
ddollar: Probably want something non-tmp so we at least still get persistence.

nzoschke: @ddollar changed to /home/convox.

But /tmp/.convox is persisted across `convox starts` too.

PR #1499, opened 2016-12-21
Fix host path for one-off volumes

### LINKED ISSUE #1375: EFS volumes not mounted for one-off processes
When running a one-off process with `convox run`, EFS volumes are not mounted.

For example, if I attach to a running 'web' process:
```
$ convox exec a91ba9f113a9 'mount | grep nfs | wc -l'
2
```

But if I run a one-off command:
```
$ convox run web 'mount | grep nfs | wc -l'
0
```

FWIW, I'm running current versions of Convox:
```
$ convox -version
client: 20161028232500
server: 20161102160040 (console.convox.com)

mattmanning:
The container path was not being prefixed with `/volumes/rack-app/service` as expected. This change adds the path prefix.

Fixes #1375

### DIFF
--- a/provider/aws/processes.go
+++ b/provider/aws/processes.go
@@ -701,11 +701,12 @@ func (p *AWSProvider) generateTaskDefinition(app, process, release string) (*ecs
 
 	for i, mv := range s.MountableVolumes() {
 		name := fmt.Sprintf("volume-%d", i)
+		host := fmt.Sprintf("/volumes/%s-%s/%s%s", p.Rack, app, process, mv.Host)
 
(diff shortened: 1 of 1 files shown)


mattmanning:
[EFS volumes not mounted for one-off processes](https://trello.com/c/BHsNvcEa/1265-efs-volumes-not-mounted-for-one-off-processes)

nzoschke:
Matches `{ "Fn::Join": [ "", [ "/volumes/", { "Ref": "AWS::StackName" }, "/{{$e.Name}}{{.Host}}:{{.Container}}" ] ] }`

ddollar:
Any way to get a quick test around this?

PR #1503, opened 2016-12-22
Daemons

nzoschke:
Specify a `convox.daemon` label, and the services will run once on every ECS agent. Uses [ECS CloudWatch Events](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs_cwet.html) and Lambda to start daemon on every instance.

```
version: '2'
services:
  agent:
    image: datadog/docker-dd-agent:latest
    environment:
     - API_KEY
    volumes:
      - /var/run/docker.sock:/var/run/docker.sock
      - /proc/:/host/proc
      - /cgroup/:/host/sys/fs/cgroup
    privileged: true
    labels:
      - convox.daemon=true
```


mwarkentin:
Interesting! Eventually maybe the implementation can be swapped out with the native ECS daemon functionality that's being added.

How does this look from a `convox scale` perspective? Will that show you how many processes are running? Is scaling the count of processes disabled for daemons?

ddollar:
Should this be using placement policies like #1504?

nzoschke:
Yes, this should be using placement policies and autoscaling stuff. I'll close this and add the daemon label design proposal to that issue.

ddollar:
Seems like placement isn't the right thing here and this strategy is fine. Seems like just a few Hound violations from being able to ship this?

ddollar:
[Daemon Scheduler](https://trello.com/c/3LNqIY35/1367-daemon-scheduler)

ddollar:
Can reopen with some rebase love.

spencerhakim:
@ddollar I picked this up last night actually, but got distracted when I discovered Docker 1.13 broke `make test`, as the Docker CLI now makes a couple of extra API calls on startup (I'll make a separate issue about this). Anyway, I just need to address the Hound issues and I'll open a PR, if that's cool with @nzoschke.

ddollar:
Please feel free to take this branch and rebase/push it up as a new PR. I think the Docker 1.13 stuff is solved now.

PR #1505, opened 2017-01-03
In docker-compose.yml files generated by 'convox init', version should be a string, not an integer.

soulshake:
Our `convox init` templates have an invalid format:
```
$ docker-compose config
ERROR: Version in "./docker-compose.yml" is invalid - it should be a string.
```

This is because `docker-compose` expects `version: "2"` and not `version: 2`.


soulshake:
[docker-compose.yml templates](https://trello.com/c/vZEGnMRg/1475-docker-compose-yml-templates)

PR #1513, opened 2017-01-05
freeze docker version and update ecs amis

ddollar:


mwarkentin:
docker version or golang version?

ddollar:
Docker version. They upgraded alpine or something in the Go 1.7.4 image.

mwarkentin:
Ah yes, python and ruby images just upgraded to alpine 3.5 as well.

PR #1527, opened 2017-01-08
Add CONVOX_WAIT envvar CLI option

soulshake:
I prefer the `--wait` behavior, but I always forget to type it when running CLI commands.

This PR introduces a new environment variable, `CONVOX_WAIT`, which, when set, assumes `--wait` by default.


mwarkentin:
Do you set `CONVOX_WAIT=true` or something else?

soulshake:
[urfave/cli](https://github.com/urfave/cli) has a [bool type](https://github.com/urfave/cli/blob/61f519fe5e57c2518c03627b194899a105838eba/flag-types.json#L3) which gets parsed using [strconv.ParseBool](https://golang.org/pkg/strconv/#ParseBool), which accepts 1, t, T, TRUE, true, True, 0, f, F, FALSE, false, False.
(The strconv.ParseBool doc says "Any other value returns an error" but the CLI just defaults to false if some other value is specified.)

till:
Is this to work against the problem where a container takes longer than x seconds to load and convox things it's defunct?

soulshake:
[CONVOX_WAIT](https://trello.com/c/WR2jNATu/1505-convox-wait)

PR #1530, opened 2017-01-09
don't give cron lambda a custom name

mattmanning:
If an app has existing cron jobs and its rack is updated >= 20170106053301, subsequent deployments of that app will hit the CloudFormation error:

> CloudFormation cannot update a stack when a custom-named resource requires replacing. Rename production-fixd-cron and update the stack again.

when CF tries to replace the function. This is because CloudFormation tries to replace the old cron Lambda function running in the node 0.10 runtime with a new one in the node 4.3 runtime, but it can't because of the custom name.

This PR removes the custom name (it's not needed) so that the function can be replaced.


mattmanning:
[Figure out how to replace cron Lambda functions.](https://trello.com/c/xwAHTHA9/1502-figure-out-lambda-how-to-replace-cron-lambda-functions)

mattmanning:
I don't have another rack with a cron lambda still in nodejs 0.10 to test this with, but I tested it with a custom-named Lambda function already running in nodejs 4.3 and it updated to the non-custom name with no problem.

PR #1531, opened 2017-01-09
prevent `convox exec` and friends from hanging on error

ddollar:


ddollar:
[convox exec with bad id hangs terminal](https://trello.com/c/QlMzrejC/1471-convox-exec-with-bad-id-hangs-terminal)

PR #1532, opened 2017-01-09
stop handling old-style services

ddollar:


ddollar:
[Service name conflicts with full stack name](https://trello.com/c/0KHDM4xv/1182-service-name-conflicts-with-full-stack-name)

MiguelMoll:
Would old services have to be updated if this update is applied? Anything to document now that the "old" way is no longer supported?

ddollar:
Yes, we'll need to add some documentation around the old services and/or Rack versions that apply.

mwarkentin:
It might be an issue if "delete and recreate" is the solution here (eg. for RDS).

skwp:
Is this going to affect any of our custom services like fluentd?

mattmanning:
[Send out email notification about breaking changes for old-style services and self-hosted Docker registry](https://trello.com/c/OlcYuXhZ/1545-send-out-email-notification-about-breaking-changes-for-old-style-services-and-self-hosted-docker-registry)

Issue #665, opened 2016-05-20
convox should support udp


beedub:
ECS now supports it. CF doesn't seem to natively, but [it seem like there is a workaround](https://forums.aws.amazon.com/thread.jspa?messageID=720619).

ddollar:
We create our services and task definitons with custom lambda functions anyway so we should be able to use this.

Proposed design:

```
web:
  labels:
    - convox.port.5184.protocol=udp
  ports:
    - 5184:9000
```

This would figure the task definiton to use UDP as the protocol and would not create a load balancer.

beedub:
What's the estimated difficulty of adding this? If the convox team doesn't have the bandwidth to tackle this, I may take a stab

ddollar:
I think this shouldn't be too difficult:

You'll first need a development rack:
https://github.com/convox/rack/blob/master/Development.md

You'll want to make sure the load balancer listener is not created when the protocol is UDP:
https://github.com/convox/rack/blob/master/api/models/templates/app.tmpl#L346

You'll want to add the protocol in here:
https://github.com/convox/rack/blob/master/api/models/templates/app.tmpl#L627

Then you'd want to grab the protocol here and add it to the TaskDefinition:
https://github.com/convox/rack/blob/master/api/cmd/formation/handler/ecs.go#L482

The one tricky part of this will be while you are working on the lambda function here in this last step the only way to test it will be rebuilding the .zip file and uploading it from the AWS console.

`cd api/cmd/formation && make` will net you a `formation.zip` to be uploaded to Lambda. You'll want to find the Lambda function that is the CustomTopic of the app you are deploying.

beedub:
It looks like `docker-compose` (because of docker?) accepts `/udp` at the end of each port configuration https://github.com/docker/compose/issues/1334. Should convox also accept this config? If so, should it _only_ accept this config, or accept this config & @ddollar's suggestion above?

ddollar:
If there's a Docker / docker-compose standard let's use that rather than non-standard label config

ddollar:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;BPAoaABo&#x2F;464-convox-udp"><img src="https:&#x2F;&#x2F;github.trello.services&#x2F;images&#x2F;trello-icon.png" width="12" height="12"> convox udp</a>

spencerhakim:
[...]

Issue #492, opened 2016-03-30
deploy apps without ELBs (but with ports exposed) - for external service discovery


soulware:
Thinking about service discovery in the context of internal (micro)-services.

We _can_ use the convox created load balancers (one per service) but at some point this might become unwieldy (_lots_ of ELBs).
There is also no real service discovery going on - you still need to know either the endpoints
of the load balancers or set DNS up to map cleanly to them.

I was wondering if there is a more lightweight/simpler approach where convox apps could (optionally) be deployed with ports exposed but without the load balancers - and leave service discovery up to an external setup like Consul+Registrator or something similar.

https://www.consul.io/
https://github.com/gliderlabs/registrator

It would be great if we could -
- deploy a service with port(s) exposed, but without an ELB being created
- let ECS choose a random port (so we could deploy more than one task per EC2 instance)

From the ECS docs -

https://aws.amazon.com/blogs/compute/service-discovery-via-consul-with-amazon-ecs/

```
"portMappings": [
  {
    "containerPort": 4567
  }
]
```

> Notice that you are not defining the host port mapping. Docker automatically assigns a port on the host so the port number is discovered via the Consul service discovery. This allows you to run multiple tasks or services of the same type on a single ECS instance.

There is some mention of host port "0" here -
https://forums.aws.amazon.com/message.jspa?messageID=596662

> If you leave out the host port, or specify it as being '0', in the task definition, then the port will be chosen automatically and reported back to the service.  After the task is running, describing it should show what port(s) it bound to.

So maybe one way to do this via `docker-compose.yml` would be to allow
ports to be specified something like -

```
# public ELB mapping port 80 -> 80
ports:
  - 80:80

# internal ELB mapping port 80 -> 80
ports:
  - 80

# *no* ELB and random host port in ECS
ports:
  - 0:80
```

Thoughts?

ShakataGaNai:
I'm a fan of NOT creating an ELB for every service. For one or two apps that's ok, but if running say 10 different "light" projects, you'd end up paying $200/mo just for ELB's.

Perhaps another option is one ELB pointing at an internal load-balancer that's then integrated into the various applications?

Crashthatch:
[...]

PR #1535, opened 2017-01-10
UDP support + Disable ELB creation

spencerhakim:
First attempt at #665 + #492, I'd love to get some feedback as I go. I haven't tested any of it yet, but hopefully there won't be too many problems.

EDIT: OK, some notes. Since ELBs only support TCP, I treat ports as having one of 3 possible states: Public/External TCP, Internal TCP, or UDP. I also did my best to update any code that used something like `len(s.Ports) > 0` to determine if an ELB should be created to instead use `s.HasBalancer()`, and in general to add protocol checks to code that assumed ports would always be TCP. The regex I used for parsing the ports from the yaml file is forgiving and backwards compatible, while the regex used for parsing ports from the ECSTaskDefinition is not; if that was a mistake, I can easily correct it.

As a side-effect of adding UDP support, I was able to easily add in a check for `convox.elb=false` to disable ELB creation entirely, even when TCP ports are defined. Feel free to suggest a better name for this label.

I still need to add my own tests, but I've fixed the existing tests to match.


manifest/yaml.go
```diff
@@ -271,41 +272,29 @@ func (pp *Ports) UnmarshalYAML(unmarshal func(interface{}) error) error {
...
-				return fmt.Errorf("error parsing port: %s", err)
-			}
+		container, _ := strconv.Atoi(parts[2])
+		p.Container = container
+		p.Balancer = container
 
-			p.Name = parts[0]
-			p.Container = n
+		// Only TCP ports can be "public" (in the ELB sense) or have an ELB at all
+		if parts[1] != "" && p.Protocol == TCP {
```
spencerhakim: I debated with myself heavily on how to handle this. I think a proper `State` enum to encapsulate `Protocol` + `Public` and ensure that UDP ports are never accidentally set to public would be cleanest, but I opted for the less invasive option of simply adding `Protocol` and leaving `Public` untouched. I'm still debating whether or not to go back and refactor it.

manifest/service.go
```diff
@@ -96,14 +96,24 @@ func (s *Service) Process(app string, m Manifest) Process {
 }
 
 func (s Service) HasBalancer() bool {
-	return len(s.Ports) > 0
+	if s.Labels["convox.elb"] == "false" {
```
ddollar: Would like to avoid AWS-specific naming here. Perhaps `convox.balancer`

manifest/yaml.go
ddollar: Makes sense to me

spencerhakim: Sorry, just to clarify, you think it makes sense to go back and change it, or to leave it as is?

ddollar: I think this change makes sense as is.

spencerhakim:
I added some new tests, and the diffs on the generated fixtures looked correct, though I'm hardly an expert on CloudFormation. My only concern is how I handled `PortMappings` in the `ECSTaskDefinition`; I'm not clear on when/how host ports are selected (random?), and while you'd normally have a parameter such as `MainPort443Host` generated and referenced, I'm not sure how dependent that is on actually having an ELB (which you wouldn't have for UDP or `convox.balancer=false`), so right now it generates

`"443:5000/tcp"`

instead of

```json
{
  "Fn::Join": [
    ":",
    [
      {
        "Ref": "MainPort443Host"
      },
      "5000/tcp"
    ]
  ]
}
```

in those scenarios. Could definitely use some guidance on this. As always, any additional feedback would be appreciated.

ddollar:
It needs to generate the Fn::Join with the Ref in the middle. This is used to select the intermediate instance port (that sits between the ELB and the container port)

spencerhakim:
Sorry if I wasn't clear, that's still the behavior for TCP ports with ELBs enabled (see [here](https://github.com/convox/rack/pull/1535/files#diff-b199ac8aa1c94cf65c1d0181753caf2dR268)), the new behavior I described only affects UDP ports, which can't use ELBs, or TCP ports when `convox.balancer=false`. I have a fixture with both behaviors [here](https://github.com/spencerhakim/rack/blob/048b8bc9387d0888d783ddaffb935261da8bcc15/api/models/fixtures/web_tcp_udp.json#L766).

spencerhakim:
Side note: `make fixtures` currently only generates missing .json files, but the [documentation](https://github.com/convox/rack/blob/master/Development.md#infrastructure-changes) seems to suggest that it should regenerate all of them.

spencerhakim:
So, is there anything else that needs to be done with this? I've removed WIP from the PR name as I think I've done everything I should.

ddollar:
Sorry, have been pushing through some changes lately and looks like this branch needs a rebase. Once you do that I'll review and get it in to our next release.

ddollar:
It also appears that hound has noted some style issues that need to be resolved, mostly around clarifying comments. Let me know if you're not sure how to resolve those.

spencerhakim:
Most of the hound issues are from pre-existing code that I had just moved around. I can add comments to that stuff if you'd like; is it an on-going process in general to add comments to existing code?

ddollar:
Yes, in general we're trying to satisfy the Hound going forward. Let me know if this is too burdensome and I can help.

spencerhakim:
Not a problem, I'm glad to hear that that's an ongoing goal for the codebase and am happy to contribute.

spencerhakim:
Not sure why, but the unit tests started failing after the rebase, both locally and on Travis. I didn't touch the affected tests, and the git logs for the relevant files also don't show any recent changes, so I'm pretty confused.

ddollar:
This may be an issue with master, I'm investigating.

spencerhakim:
@ddollar I figured it out, looks like #1536 [missed something](https://github.com/convox/rack/blob/master/api/models/templates/app.tmpl#L9), plus fixtures need to be regenerated. I can fix it here, or wait for you to fix it in master.

spencerhakim:
Rebased again now that those fixes are in, all the GitHub checks should pass now.

ddollar:
You should be able to safely rebase on master now, it should no longer in flux. Sorry for all the hassle!

spencerhakim:
OK, rebased again, should still be all good.

spencerhakim:
Fixed the issue with installing a rack, and rebased.

spencerhakim:
Rebased and fixed the merge conflict. Any reason we can't get this merged soon?

MiguelMoll:
@spencerhakim Thanks for taking care of this. Any chance you could knock out one more rebase because of the release going out today? That way I can starting to get this out next. Thanks!

spencerhakim:
@MiguelMoll Done.

.editorconfig
```diff
@@ -0,0 +1,19 @@
+root = true
+
```
MiguelMoll: Believe this file was committed by mistake.

spencerhakim:
No, but I can remove it if you'd like, though it's probably a good idea to leave it. http://editorconfig.org/

.editorconfig
spencerhakim: No, but I can remove it if you'd like, though it's probably a good idea to leave it. http://editorconfig.org/

MiguelMoll: Yes please that'd be appreciated. As I test this out today, if you like I can tackle any issues that run up. That way I don't keep bothering about little stuff like that.

spencerhakim: Removed. And that's fine with me, it's your project after all.

till:
@MiguelMoll is this merged?

MiguelMoll:
Not yet. It's making its way through our testing and CI process. Hopefully very soon.

MiguelMoll:
Merged via #1639. Thanks again @spencerhakim for this PR! 👍

mattmanning:
[Document UDP support and convox.balancer label](https://trello.com/c/T9XWxY6F/1774-document-udp-support-and-convox-balancer-label)

spencerhakim:
@mattmanning I started on that over [here](https://github.com/convox/site/pull/457). Feel free to use it as a starting point and close the PR.

advance512:
Did this PR break envvar based linking when `convox.balancer=false`?

spencerhakim:
@advance512 Linking depends on a load balancer, so if you explicitly disable the creation of a load balancer then, yes, it breaks linking.

advance512:
Would it not make sense to at least pass one of the created container's endpoint details?

spencerhakim:
@advance512 That doesn't really work due to the somewhat ephemeral nature of containers. Not sure if there's even a clean way Convox could support that if they wanted to. Certainly outside the scope of this PR. If avoiding ELBs is a priority for you, you should look into using a service discovery system like etcd or Consul.

till:
@advance512 we are in the same boat. you need to get creative with service discovery, i think. Did you have any thoughts on this yet?

advance512:
@till 
I am pretty sure many others have been in this same boat, and in fact this might be the "killer feature" that prevents the project from being more popular among small startups and enthusiasts. I have 8 microservices in my project (and this will expand, a microservice per external system I interface into), but only one public access point. That is $140 that I would pay for no reason at all. Too bad, because I really like Convox otherwise.

Anyways, this missing feature might sadly be the reason I will stop using Convox and move to some sort of K8s based solution (like Rancher, Mantl or OpenShift - or many others).

For now, after deployment, I've created a small script that runs `convox api get /apps/APP_NAME/processes`, and updates DNS records in Route53 in which I've created an Private Zone (microservice.app_name.internal). 

Not very fun, definitely not production worthy, but it works.

ddollar:
Stay tuned, we have some good stuff in the works for you folks.

Now that we have host-based routing in ALB we can easily do one-ALB-per-Rack and route the internal hostnames correctly. The team is hard at work getting a completely revamped approach centered around this ready. We're hoping to have something for you all to try in the next few weeks.

till:
@advance512 while @ddollar is working on something, we are playing with this method: https://github.com/convox-examples/dd-agent/pull/3

You basically scale the app to the number of instances and then use something like they did to "discover" the app locally.

advance512:
@till  - that sounds great. Will have a look - though I think it is quickly nearing the same complexity as having a Consul cluster which is what I was trying to avoid.

till:
@advance512 not at all using dd-agent, I am just using/testing their ruby hack for a local statsd that we run.

But yeah, I am also trying to avoid running consul or similar tools.

spencerhakim:
@advance512 For the `dd-agent` example, if you set the label `convox.agent=true` (which scales the process to exactly one instance per EC2 host) and expose `8125/udp`, you can do something like `export EC2_HOSTNAME=$(curl -s http://169.254.169.254/latest/meta-data/local-hostname)` and get a poor man's service discovery by just pointing your app at `${EC2_HOSTNAME}:8125`. `dd-agent` is actually what motivated me to work on this PR and the `convox.agent` PR.

For more info on that `curl` endpoint: http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html

advance512:
I tried using Consul and Linkerd. Let me just say it's very, very complex with Convox.

There are no guides or tutorials on the issue. To create consul cluster, you'd need the addresses of the other servers (TCP, not HTTP), and there is simply no easy way to do it that I could found. You could use `Networking mode: Host` in an ECS task definition to make things easier, and in fact that is what the best article I found [suggests](https://blog.unif.io/deploying-consul-with-ecs-part-2-38de98a56611), but that means you're exiting the Convox world into "direct to AWS" world. You'd probably need to defined DNS records to your Consul servers. Or, you could deploy the Consul servers on their own EC2 instances. However, you'd still need Consul clients and Consul-Registrator on each server. Adding Linkerd for a network mesh puts you in the same problem as with Consul, the need for hard-set peers that are accessible not via ELBs. Then there are issues with docker-compose vs local vs remote - e.g. you need `networks` in the docker-compose to make it work, and it indeed works with Docker Compose but not with `convox start`, you would need to create an exterior network. In production, you'd need `links`. `depends_on` also works in some places and not others.. I don't remember exactly.

I spent nearly 2 days trying to get this to work and eventually just gave up. For now I'm working, again, with multiple ELBs and am actively waiting for the new ALB feature.. if it doesn't arrive soon, I'll have to migrate to Kubernetes. I was trying to avoid this due to not wanting to spend a lot of time on DevOps work at this point in the life of the project.. but I also can't be spending hundreds of dollars per month for ELBs, either.

PR #1536, opened 2017-01-10
Disable internal registry, fix us-east-2, add us-west-1

ddollar:
Also adds us-east-2 to CI


ddollar:
[disable internal registry and fix us-east-2](https://trello.com/c/gLBuKNtV/1531-disable-internal-registry-and-fix-us-east-2)

PR #1537, opened 2017-01-11
Give AWS credential csv files provided as an argument priority over env vars

soulshake:
By accepting this PR, we will:

- add a new test `TestConvoxInstallFileCredentialsInsufficientPermissions`
- check the format of the credentials file more carefully
- be more verbose about what AWS account we're installing to
- prefer credentials in provided file over envvars
- validate user access permissions, _kind of_ (see below)

**IMPORTANT**

Previously, the `validateUserAccess` function in `cmd/convox/install.go` simply returned `nil`. If this PR is accepted, it will only return `nil` if the provided secret key is `test`. This is necessary for the permissions error to bubble up, but there are two problems:

- I don't know what unexpected behaviors this change might cause, due to lacking test coverage
- we're not actually validating user permissions correctly (`Iam.GetUser` permissions is not what matters, IIUC), which I suppose is why the function was disabled in the first place.

Feedback welcome

### DIFF
--- a/cmd/convox/install.go
+++ b/cmd/convox/install.go
@@ -370,6 +370,7 @@ func cmdInstall(c *cli.Context) error {
 		req.TemplateBody = aws.String(t.String())
 	}
 
+	//log.Fatal("Aborting before creating stack.")
 	res, err := CloudFormation.CreateStack(req)
 	if err != nil {
 		if awsErr, ok := err.(awserr.Error); ok {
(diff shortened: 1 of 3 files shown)


soulshake:
[CLI Installer CSV Broken](https://trello.com/c/OvHFfNmL/1514-cli-installer-csv-broken)

soulshake:
Ah, my test broke, the error is different now (`InvalidClientTokenId`). Don't merge

ddollar:
Can you include an example terminal transcript using this functionality in the PR description please.

soulshake:
@ddollar hold off on merging this please, @MiguelMoll has just pointed out that he gets a different CSV format when he downloads credentials.

ddollar:
Needs a rebase

soulshake:
@ddollar rebase done

PR #1540, opened 2017-01-11
Enhancement: configurable wait option for `convox start`

till:
This aims to address the arbitrary wait timeout in `convox start`. The default is
the current 10 second interval, but this option lets developers tune that if the
container takes a little longer to start.

### DIFF
--- a/manifest/run.go
+++ b/manifest/run.go
@@ -33,6 +33,7 @@ type RunOptions struct {
 	Cache   bool
 	Quiet   bool
 	Sync    bool
+	Wait    int
 }
 
 // NewRun Default constructor method for a Run object
@@ -180,7 +181,7 @@ func (r *Run) Start() error {
 
 		r.Processes[p.Name] = p
 
-		if err := waitForContainer(p.Name, s); err != nil {
+		if err := waitForContainer(p.Name, s, r.Opts.Wait); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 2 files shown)


manifest/run.go
```diff
@@ -238,23 +239,22 @@ func pruneSyncs(syncs []sync.Sync) []sync.Sync {
 	return pruned
 }
 
-func waitForContainer(container string, service Service) error {
-	i := 0
+func waitForContainer(container string, service Service, wait int) error {
+	waitTimeout := time.add(time.parseDuration("%ds", wait))
```
till: I rewrote this to use actual time. At least I am hoping I did. The idea was to say, "now+configured interval".

(Disclaimer: I think I've never written that much Go before.)

manifest/run.go
```diff
@@ -238,23 +239,22 @@ func pruneSyncs(syncs []sync.Sync) []sync.Sync {
...
 
 		if host != "" {
 			return nil
 		}
 
-		// wait 10s max
-		if i > 100 {
-			return fmt.Errorf("%s failed to start within 10 seconds", container)
+		// wait for the container to start, default: 10 seconds
+		if time.After(waitTimeout) {
```
till: Here's where I am using it again — hopefully using `time.After()` correctly.

manifest/run.go
```diff
@@ -238,23 +239,22 @@ func pruneSyncs(syncs []sync.Sync) []sync.Sync {
...
+			return fmt.Errorf("%s failed to start within %d seconds", container, wait)
 		}
 
-		time.Sleep(100 * time.Millisecond)
+		time.Sleep(time.parseDuration("1s"))
```
till: It seemed like your `time.Sleep()` incrementally gets longer and longer. I am not sure if that was the intention. I decided to just sleep for one second.

ddollar:
Going to try to solve this in #1545

PR #1541, opened 2017-01-11
update ecs amis

ddollar:


mwarkentin:
Might want to be a bit careful with this one if it's upgrading docker 1.11 -> 1.12.

I'm experimenting with the new AMIs on our dev rack as well.

mwarkentin:
Actually, according to [this](http://docs.aws.amazon.com/AmazonECS/latest/developerguide/container_agent_versions.html#ecs-optimized-ami-agent-versions), it looks like this is still on docker 1.11:

PR #1544, opened 2017-01-11
reset terminal after raw session ends

ddollar:


ddollar:
[reset terminal after exec exits](https://trello.com/c/8RXuUEp6/1533-reset-terminal-after-exec-exits)

mattmanning:
[`convox exec` leaves shell in weird state after exiting](https://trello.com/c/4rHrJkaT/381-convox-exec-leaves-shell-in-weird-state-after-exiting)

mwarkentin:
Not sure if this is related - but after running a `convox exec` command, input is no longer showing up in my terminal. I can still type and enter commands, just can't see what I'm doing (output shows up though).

```
client: 20170123152656
server: 20170111205227 (console.convox.com)
```

PR #1545, opened 2017-01-11
wait 60s for containers to come up on slower environments

ddollar:


ddollar:
[wait 60s for containers to come up on slower environments](https://trello.com/c/NejHvNOx/1534-wait-60s-for-containers-to-come-up-on-slower-environments)

manifest/run.go
```diff
@@ -249,8 +249,8 @@ func waitForContainer(container string, service Service) error {
 			return nil
 		}
 
-		// wait 10s max
-		if i > 100 {
+		// wait 60s max
+		if i > 600 {
 			return fmt.Errorf("%s failed to start within 10 seconds", container)
```
till: This line could be updated!

ddollar: Good catch

PR #1548, opened 2017-01-12
Paginate task listings

ddollar:
Allow `convox ps` to function when there are more than 100 containers on an app.

### DIFF
--- a/provider/aws/processes.go
+++ b/provider/aws/processes.go
@@ -223,59 +223,77 @@ func (p *AWSProvider) stackTasks(stack string) ([]string, error) {
 	return tasks, nil
 }
 
+const describeTasksPageSize = 100
+
 func (p *AWSProvider) taskProcesses(tasks []string) (structs.Processes, error) {
 	log := Logger.At("serviceProcesses").Namespace("tasks=%q", tasks).Start()
 
-	ptasks := []*string{}
-
-	for _, t := range tasks {
-		ptasks = append(ptasks, aws.String(t))
-	}
-
-	if len(ptasks) == 0 {
-		return structs.Processes{}, nil
-	}
-
-	tres, err := p.ecs().DescribeTasks(&ecs.DescribeTasksInput{
-		Cluster: aws.String(p.Cluster),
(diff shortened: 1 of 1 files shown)


ddollar:
[`convox ps` fails when there are more than 100 processes](https://trello.com/c/Owmz5SeP/1438-convox-ps-fails-when-there-are-more-than-100-processes)

api/models/models.go
```diff
@@ -51,7 +51,7 @@ func awsConfig() *aws.Config {
 		config.Region = aws.String(r)
 	}
 
-	if os.Getenv("DEBUG") != "" {
```
soulshake: What's the difference between `DEBUG` and `CONVOX_DEBUG`?

ddollar: CONVOX_DEBUG is for the CLI. DEBUG is used internally by rack

provider/aws/processes.go
```diff
@@ -205,18 +205,26 @@ func (p *AWSProvider) stackTasks(stack string) ([]string, error) {
...
 			Cluster:     aws.String(p.Cluster),
 			ServiceName: aws.String(s),
-		})
+		},
+			func(page *ecs.ListTasksOutput, lastPage bool) bool {
+				pageNum++
+				for _, arn := range page.TaskArns {
+					tasks = append(tasks, *arn)
+				}
+				fmt.Println(page)
```
MiguelMoll: Debug statement to remove?

provider/aws/processes.go
```diff
@@ -205,77 +205,103 @@ func (p *AWSProvider) stackTasks(stack string) ([]string, error) {
...
+				return pageNum <= 3
```
MiguelMoll: Only getting the first 3 pages of results?

provider/aws/processes.go
mattmanning: yes good eye

provider/aws/processes.go
mattmanning: Updated to iterate to the last page

PR #1553, opened 2017-01-13
remove the syslog lambda custom name so the function can be replaced

mattmanning:
When existing syslog resources are linked or unlinked Convox will try to replace them and create a new function in the nodejs 4.3 Lambda runtime because of the nodejs 0.10 Lambda runtime deprecation.

That replacement currently fails because the function has a custom name. This change removes the custom name.

This change is dependent on https://github.com/convox/syslog/pull/6 and should not be released until the new lambda.zip has been uploaded to our S3 buckets.


provider/aws/templates/service/syslog.tmpl
```diff
@@ -200,6 +197,7 @@
                         {
                           "Ref": "AWS::StackName"
                         },
+                        "*",
```
ddollar: Should probably be -* to avoid spillage from e.g.  `dev` and `dev-foo`

mattmanning: Good point, thanks

mattmanning:
[Handle custom-name replacement errors for syslog Lambda](https://trello.com/c/pfKyfj6s/1537-handle-custom-name-replacement-errors-for-syslog-lambda)

provider/aws/templates/service/syslog.tmpl
ddollar: Should actually be `-Function-*`

ddollar:
Seems like this might have gotten buried but the wildcard should be `-Function-*`

mattmanning:
My bad. I misread your original comment. Pattern should be correct now.

ddollar:
Needs rebase, removing my review until then.

mattmanning:
@ddollar Just noticed this one. Rebased and `make test` is passing.

mattmanning:
@ddollar Just rebased again. This one should be good to go.

PR #1556, opened 2017-01-14
Allow running a subset of tests

soulshake:
Since we [un]set some local envvars (`PROVIDER`, `CONVOX_WAIT`) in the Makefile, running subsets of tests via `go test` doesn't result in identical environments.

This PR allows running only some tests, e.g.:

`$ make test TEST=github.com/convox/rack/cmd/convox`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ templates:
 	make -C sync templates
 
 test:
-	env PROVIDER=test bin/test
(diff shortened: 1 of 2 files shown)


soulshake:
[Allow running subset of tests](https://trello.com/c/fx58VzJO/1565-allow-running-subset-of-tests)

PR #1063, opened 2016-08-15
Create KernelUser's managed policy

MiguelMoll:

PR #1561, opened 2017-01-17
Fix `convox ps stop`

spencerhakim:
As a result of #1063, `convox ps stop` was broken due to missing permissions for `ecs:StopTask`. This PR simply adds it.

I haven't checked for any other missing permissions, I just happened to stumble onto this one during some debugging.


ddollar:
Thanks!

PR #1574, opened 2017-01-18
Back out client calls to /resources for now to avoid 404s on older Racks

mattmanning:


mattmanning:
[Back out client calls to /resources for now to avoid 404s on older Racks](https://trello.com/c/uXGH1U87/1612-back-out-client-calls-to-resources-for-now-to-avoid-404s-on-older-racks)

PR #1579, opened 2017-01-19
Make Yes the default value for the Autoscale param

mattmanning:
Based on my testing this won't change the Autoscale setting for existing Racks, but should be helpful for new Racks.


MiguelMoll:
This change could make CI tricky as it didn't complete in one region. Given [example](https://circleci.com/gh/convox/rack/1847#tests/containers/2) is the Rack is autoscaling and putting the stack in an update state. This will prevent the uninstallation of a Rack.

mattmanning:
[Make Autoscale=Yes the default](https://trello.com/c/5UdEyIL9/1626-make-autoscale-yes-the-default)

PR #1584, opened 2017-01-19
Add ECR region to config

mattmanning:
This enables private console builds for Racks in regions other than us-east-1.


mattmanning:
[Fix private registries region issue](https://trello.com/c/pMZgudg6/1617-fix-private-registries-region-issue)

PR #1587, opened 2017-01-20
Restore HTTP SNS subscription confirmations

nzoschke:
This fixes creating new webhooks.

The SNS subscription stuff was depending on the /sns endpoint on port 80 to confirm subscriptions. It was also depending on the sns:ConfirmSubscription capability.

SNS does not deliver notification messages to a self-signed cert. https://forums.aws.amazon.com/thread.jspa?messageID=204271&#204271

### DIFF
--- a/provider/aws/dist/rack.json
+++ b/provider/aws/dist/rack.json
@@ -769,6 +769,7 @@
                       "sns:Publish",
                       "sns:SetTopicAttributes",
                       "sns:Subscribe",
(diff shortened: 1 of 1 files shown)


MiguelMoll:
[Slack Integration broken in Rack 20170118202051](https://trello.com/c/C4gKKVdc/1630-slack-integration-broken-in-rack-20170118202051)

PR #1589, opened 2017-01-20
Make authentication error message friendlier

ianmalott:


api/controllers/api.go
```diff
@@ -13,6 +13,10 @@ import (
 	"golang.org/x/net/websocket"
 )
 
+const authenticationErrorMessage = `Unable to authenticate.
+Please confirm your Rack API Key is correct.
+See https://convox.com/docs/api-keys/ for more information.`
```
ianmalott: 18: not sure if it's a good idea to put links in our error messages, since then we have to maintain those URLs

PR #1593, opened 2017-01-20
Add releases limit option

ianmalott:
Usage:

```shell
convox releases # defaults to 20
convox releases --limit=4
convox releases --limit=30
```

### DIFF
--- a/api/controllers/routes.go
+++ b/api/controllers/routes.go
@@ -35,7 +35,7 @@ func NewRouter() (router *mux.Router) {
 	router.HandleFunc("/apps/{app}/processes/{process}", api("process.stop", ProcessStop)).Methods("DELETE")
 	router.HandleFunc("/apps/{app}/processes/{process}/run", api("process.run.detach", ProcessRunDetached)).Methods("POST")
 	router.HandleFunc("/apps/{app}/releases", api("release.list", ReleaseList)).Methods("GET")
-	router.HandleFunc("/apps/{app}/releases", api("release.list", ReleaseList)).Methods("GET").Queries("limit", "")
+	router.HandleFunc("/apps/{app}/releases", api("release.list", ReleaseList)).Methods("GET").Queries("limit", "{limit:[0-9]+}")
 	router.HandleFunc("/apps/{app}/releases/{release}", api("release.get", ReleaseGet)).Methods("GET")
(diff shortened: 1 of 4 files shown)


api/controllers/routes.go
```diff
@@ -35,6 +35,7 @@ func NewRouter() (router *mux.Router) {
 	router.HandleFunc("/apps/{app}/processes/{process}", api("process.stop", ProcessStop)).Methods("DELETE")
 	router.HandleFunc("/apps/{app}/processes/{process}/run", api("process.run.detach", ProcessRunDetached)).Methods("POST")
 	router.HandleFunc("/apps/{app}/releases", api("release.list", ReleaseList)).Methods("GET")
+	router.HandleFunc("/apps/{app}/releases", api("release.list", ReleaseList)).Methods("GET").Queries("limit", "{limit:[0-9]+}")
```
ianmalott: I stuck this in a [separate commit](https://github.com/convox/rack/pull/1593/commits/7e9ae9a623668cb20adcc39981ea31164cff1d6d) because I wasn't sure if we'd want to restrict the param values with a regex, and I didn't find precedents to reference in convox/rack or convox/console.

```diff
-	router.HandleFunc("/apps/{app}/releases", api("release.list", ReleaseList)).Methods("GET").Queries("limit", "")
+	router.HandleFunc("/apps/{app}/releases", api("release.list", ReleaseList)).Methods("GET").Queries("limit", "{limit:[0-9]+}")
```

client/releases.go
```diff
@@ -29,6 +29,19 @@ func (c *Client) GetReleases(app string) (Releases, error) {
 	return releases, nil
 }
 
+// GetReleasesWithLimit requests releases of an app, overriding the default limit.
+func (c *Client) GetReleasesWithLimit(app string, limit int) (Releases, error) {
```
MiguelMoll: I'm wondering if we should condense this with `GetReleases(app string)` by adding the limit parameter to it. That way we keep one code path when it comes to getting releases.

api/controllers/routes.go
MiguelMoll: This makes sense and looks right. 👍

client/releases.go
ianmalott: Happy to go either way. I started going down the path you mention, but then I saw this pattern used for builds:

https://github.com/convox/rack/blob/master/client/builds.go#L40

MiguelMoll: All good. There is precedent so it's good as is. 👍

Issue #1560, opened 2017-01-16
`--id` support for convox env set


mwarkentin:
`convox build` and `convox deploy` both support a `--id` option to improve scripting support:

> --id				build logs on stderr, build id on stdout (useful for scripting)

It'd be nice if `convox env set` had this as well (which also creates a new release like the other two).

soulshake:
Ooh, I hadn't even noticed the existence of the `--id` flag. Will try to address this in a separate PR.

soulshake:
![](https://github.trello.services/images/mini-trello-icon.png) [--id support for 'convox env set'](https://trello.com/c/xACBkP84/1576-id-support-for-convox-env-set)

nzoschke:
Closing this in favor of an earlier issue #1424

PR #1594, opened 2017-01-20
Initial commit adding --id to 'convox env set'

soulshake:
Attempt to address #1560 
This needs to be refactored.

Also, I can't figure out how to test `--promote` behavior :(


soulshake:
[--id support for 'convox env set'](https://trello.com/c/HORtTg05/1598-id-support-for-convox-env-set)

cmd/convox/env.go
```diff
@@ -166,27 +175,46 @@ func cmdEnvSet(c *cli.Context) error {
...
 		return stdcli.Error(err)
 	}
 
-	fmt.Println("OK")
+	output.Write([]byte("OK\n"))
 
 	if releaseID != "" {
 		if c.Bool("promote") {
-			fmt.Printf("Promoting %s... ", releaseID)
+			output.Write([]byte(fmt.Sprintf("Promoting ")))
```
MiguelMoll: Any reason not to keep the same text as before? `"Promoting %s... "`

cmd/convox/env.go
```diff
@@ -166,27 +175,46 @@ func cmdEnvSet(c *cli.Context) error {
...
+			os.Stdout.Write([]byte(fmt.Sprintf("%s", releaseID)))
```
MiguelMoll: Don't think this is needed if the format above is followed.

cmd/convox/env.go
```diff
@@ -166,27 +175,46 @@ func cmdEnvSet(c *cli.Context) error {
...
 
 			_, err = rackClient(c).PromoteRelease(app, releaseID)
 			if err != nil {
 				return stdcli.Error(err)
 			}
 
-			fmt.Println("OK")
+			// If we're redirecting stdout but not stderr, print the release ID to stderr
+			if c.Bool("id") && terminal.IsTerminal(int(os.Stderr.Fd())) && !terminal.IsTerminal(int(os.Stdout.Fd())) {
```
MiguelMoll: Wouldn't a check for just `c.Bool("id")` be enough?

cmd/convox/env.go
```diff
@@ -166,27 +175,46 @@ func cmdEnvSet(c *cli.Context) error {
...
+				output.Write([]byte(fmt.Sprintf("%s", releaseID)))
+			}
+
+			output.Write([]byte("... OK\n"))
 		} else {
-			fmt.Printf("To deploy these changes run `convox releases promote %s`\n", releaseID)
+			output.Write([]byte("To deploy these changes run `convox releases promote "))
+			os.Stdout.Write([]byte(fmt.Sprintf("%s", releaseID)))
```
MiguelMoll: We could combine this two outputs as above.

cmd/convox/env.go
```diff
@@ -166,27 +175,46 @@ func cmdEnvSet(c *cli.Context) error {
...
+
+			// If we're redirecting stdout but not stderr, print the release ID to stderr
+			if c.Bool("id") && terminal.IsTerminal(int(os.Stderr.Fd())) && !terminal.IsTerminal(int(os.Stdout.Fd())) {
```
MiguelMoll: `c.Bool("id")` check?

cmd/convox/env.go
soulshake: As I recall from my manual testing, something like this is needed because otherwise if the user redirects the output to a file, stderr will print something like

```
To deploy these changes run `convox releases promote `
```
 without the release ID (and/or the release ID will get written to stdout twice.)

cmd/convox/env.go
```diff
@@ -11,6 +11,7 @@ import (
 	"strings"
 
 	"github.com/convox/rack/cmd/convox/stdcli"
+	//"golang.org/x/crypto/ssh/terminal"
```
MiguelMoll: We can remove this line if it's not used.

mwarkentin:
🎉

PR #1595, opened 2017-01-20
Verify app name and process ID when calling exec

MiguelMoll:


MiguelMoll:
[convox exec use correct app](https://trello.com/c/x8sPIPob/1648-convox-exec-use-correct-app)

MiguelMoll:
Good point. I can be of concern if the list reaches 100s of processes. Perhaps check the ARN name for the app name instead? Doesn't seem right.

PR #1602, opened 2017-01-24
WIP: Add Rack to Console from CLI `convox install`

nzoschke:
If logged into console, register the rack there instead of saving credentials locally.

There are a few design questions:

- [ ] How do you control what org and what rack name it is added to?
- [ ] Should switchHost not get called in other scenarios?


nzoschke:
[Add rack to console from `convox install`](https://trello.com/c/GTkDADim/1675-add-rack-to-console-from-convox-install)

nzoschke:
Closing this, not an important change for now.

PR #1611, opened 2017-01-25
Update syslog role to match stack name, not function name

mattmanning:


mattmanning:
[Update syslog role to match stack name, not function name](https://trello.com/c/V5T5FxP7/1693-update-syslog-role-to-match-stack-name-not-function-name)

PR #1615, opened 2017-01-25
New strategy for Rack service replacement

ddollar:
Avoids the stuck updates and scaling requirement

### DIFF
--- a/provider/aws/dist/rack.json
+++ b/provider/aws/dist/rack.json
@@ -1581,7 +1581,7 @@
         "HealthCheck": {
           "HealthyThreshold": "2",
           "Interval": 5,
-          "Target": "HTTP:4000/check",
+          "Target": "HTTP:3000/check",
           "Timeout": 3,
           "UnhealthyThreshold": "2"
         },
@@ -1595,13 +1595,13 @@
             "Protocol": "TCP",
             "LoadBalancerPort": "80",
             "InstanceProtocol": "TCP",
-            "InstancePort": "4000"
+            "InstancePort": "3000"
           },
           {
             "Protocol": "TCP",
             "LoadBalancerPort": "443",
             "InstanceProtocol": "TCP",
-            "InstancePort": "4001"
+            "InstancePort": "3001"
           }
         ],
         "LoadBalancerName": { "Fn::If": [ "PrivateApi",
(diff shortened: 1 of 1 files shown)


provider/aws/dist/rack.json
```diff
@@ -1680,16 +1680,16 @@
       },
       "Type": "AWS::EC2::SecurityGroup"
     },
-    "RackWeb": {
-      "DependsOn": [ "RackWebTasks", "Balancer", "Cluster", "CustomTopic", "Instances", "ServiceRole" ],
+    "WebService": {
```
mattmanning: Could we call this RackAPI just so it's a little easier to know what it is if looking at the Services list in ECS?

provider/aws/dist/rack.json
```diff
@@ -1698,11 +1698,11 @@
           }
         ],
         "Role": { "Fn::GetAtt": [ "ServiceRole", "Arn" ] },
-        "TaskDefinition": { "Ref": "RackWebTasks" }
+        "TaskDefinition": { "Ref": "WebTasks" }
       }
     },
-    "RackMonitor": {
```
mattmanning: I don't think RackMonitor needs this since it doesn't have ports, so doesn't have the same capacity problems as RackWeb. Were you just changing it to keep the naming scheme consistent?

Wondering if we could keep this as RackMonitor and call the other one RackAPI.

provider/aws/dist/rack.json
ddollar: In the docker-compose.yml for Rack these are `web` and `monitor`. Eventually I'd like Rack's own cloudformation template to come from the same code that generates our templates for apps.

provider/aws/dist/rack.json
ddollar: Yes consistency.

provider/aws/dist/rack.json
```diff
@@ -1680,16 +1680,16 @@
...
       "Type": "AWS::ECS::Service",
       "Properties": {
         "Cluster": { "Ref": "Cluster" },
         "DeploymentConfiguration": {
-          "MinimumHealthyPercent": "100",
+          "MinimumHealthyPercent": "33",
           "MaximumPercent": "200"
         },
-        "DesiredCount": "2",
+        "DesiredCount": "3",
```
MiguelMoll: Would we revert back to 2 processes once instance draining is working properly?

provider/aws/dist/rack.json
mattmanning: That makes sense.

PR #1621, opened 2017-01-26
option to disable autoscale during install

MiguelMoll:


cmd/convox/install.go
```diff
@@ -111,6 +111,10 @@ func init() {
 				Usage: "internet gateway id to use in existing vpc",
 			},
 			cli.BoolFlag{
+				Name:  "no-autoscale",
```
ddollar: This could be a `BoolTFlag` named `autoscale` and then both `--autoscale` and `--no-autoscale` should work with true being the default.

MiguelMoll: Nice, I wasn't aware of `BoolTFlag`. Good to know but for what it's worth, after looking it up, later version of`urfave/cli` remove that type and gave `BoolFlag` a default value.

PR #1625, opened 2017-01-27
'only available on console' -> 'only available when logged into console.convox.com'

soulshake:
Before:
```
$ convox racks
ERROR: only available in console
```

After:
```
$ convox racks
ERROR: only available when logged into console.convox.com (or your private console)
```

This change affects `convox switch` and `convox racks`.

### DIFF
--- a/api/controllers/racks.go
+++ b/api/controllers/racks.go
@@ -7,5 +7,5 @@ import (
 )
 
 func RackList(rw http.ResponseWriter, r *http.Request) *httperr.Error {
-	return httperr.Errorf(403, "only available on console")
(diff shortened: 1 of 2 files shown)


soulshake:
@ddollar How's this:
`ERROR: only available when logged into console.convox.com (or your private console)`

soulshake:
[Reword "ERROR: only available on console"](https://trello.com/c/82TziTBe/1607-reword-error-only-available-on-console)

ddollar:
A good error message tells them what is wrong and suggests how they can fix it. How about something like:

ERROR: Your CLI is pointing directly at a Rack. To log in to Console instead run `convox login`

Feel free to tweak the wording but the "whats wrong, how can a fix it" would be good to get across

PR #1628, opened 2017-01-27
fix race condition with --private installs

mattmanning:
There were several recent user reports of installations with --private failing.

These seem to be happening because instances are booting before the private network stack is fully set up. That means there is no route to the Internet, so `yum` commands are failing in the instance boot script.

This change uses a Tags hack to make private stacks wait for the NATs and subnet associations to be set up before creating the AutoScaling Group that boots the instances.

### DIFF
--- a/provider/aws/dist/rack.json
+++ b/provider/aws/dist/rack.json
@@ -637,15 +637,17 @@
         "ManagedPolicyArns": [ { "Ref": "KernelUserPolicy" }, { "Ref": "KernelUserIAMPolicy" } ]
       }
     },
-    "KernelUserPolicy": {
+    "KernelUserGeneralPolicy": {
       "Type" : "AWS::IAM::ManagedPolicy",
       "Properties" : {
         "Path": "/convox/",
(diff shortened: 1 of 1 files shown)


mattmanning:
I had a couple of good test installs with this but the last one failed with an error about policy documents not being allowed to be longer than 5120 characters, which is explained here: http://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-limits.html

@MiguelMoll Any thoughts on getting around that? Can we break out another document like you did for some of the IAM policy?

mattmanning:
[Install with --private fails](https://trello.com/c/ednHW9Fx/1703-install-with-private-fails)

MiguelMoll:
Unfortunately breaking it out is the only way. I suggest creating a new policy of the large permission set ( lines 650-776) that are pretty broad.

mattmanning:
Not able to get consistent successes, and looks like this change would trigger another ECS service replacement.

Closing this in favor of https://github.com/convox/rack/pull/1633

PR #1629, opened 2017-01-27
populate Type from Service tag if Resource tag is empty

mattmanning:
Fixes missing type from resource listing and errors like `ERROR: Resource type  does not have a link strategy`


mattmanning:
[resources created before the services->resources change don't show a type](https://trello.com/c/WmYjBuVO/1716-resources-created-before-the-services-resources-change-don-t-show-a-type)

PR #1642, opened 2017-01-30
Break out general KernelUser policy

mattmanning:
**Note**: this might cause another ECS Service replacement. We should test for that and take appropriate actions to mitigate update issues if so.

Break out general KernelUser policy to avoid AWS character limits.

Over time we will move more of these into "locked down" policy documents.

### DIFF
--- a/provider/aws/dist/rack.json
+++ b/provider/aws/dist/rack.json
@@ -634,18 +634,24 @@
       "Type": "AWS::IAM::User",
       "Properties": {
         "Path": "/convox/",
-        "ManagedPolicyArns": [ { "Ref": "KernelUserPolicy" }, { "Ref": "KernelUserIAMPolicy" } ]
+        "ManagedPolicyArns": [
+          { "Ref": "KernelUserPolicy" },
+          { "Ref": "KernelUserIAMPolicy" },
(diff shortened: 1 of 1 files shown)


mattmanning:
[KernelUserPolicy - Cannot exceed quota](https://trello.com/c/eBuvemHr/1735-kerneluserpolicy-cannot-exceed-quota)

provider/aws/dist/rack.json
```diff
@@ -634,18 +634,24 @@
       "Type": "AWS::IAM::User",
       "Properties": {
         "Path": "/convox/",
-        "ManagedPolicyArns": [ { "Ref": "KernelUserPolicy" }, { "Ref": "KernelUserIAMPolicy" } ]
+        "ManagedPolicyArns": [
+          { "Ref": "KernelUserPolicy" },
+          { "Ref": "KernelUserIAMPolicy" },
+          { "Ref": "KernelUserGeneralPolicy" }
```
MiguelMoll: `KernelUserGeneralPolicy` should be added to the `DependsOn` for the user resource.

mattmanning: @MiguelMoll Done, thanks.

PR #1646, opened 2017-01-31
Restore terminal before sending exit code to ch

mattmanning:
This fixes the problem where the terminal doesn't get restored when you exit a `convox exec` session.

### DIFF
--- a/client/processes.go
+++ b/client/processes.go
@@ -134,6 +134,7 @@ func (c *Client) StopProcess(app, id string) (*Process, error) {
 func copyWithExit(w io.Writer, r io.Reader, ch chan int) {
 	buf := make([]byte, 1024)
 	isTerminalRaw := false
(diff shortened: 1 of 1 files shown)


mattmanning:
[`convox exec` leaves shell in weird state after exiting](https://trello.com/c/dQED1tMt/1596-convox-exec-leaves-shell-in-weird-state-after-exiting)

MiguelMoll:
Going by a previous commit, seems the intent was to read data before making the terminal raw. Unfortunately the reason isn't clear.

client/processes.go
```diff
@@ -148,13 +149,14 @@ func copyWithExit(w io.Writer, r io.Reader, ch chan int) {
 		}
 
 		if !isTerminalRaw {
```
MiguelMoll: Could we remove this flag and check if we just set it to raw from the beginning?

mattmanning:
working for me 👍

PR #1649, opened 2017-01-31
Cherry-pick from convox-init-missing-error

soulshake:


MiguelMoll:
[Display visible error when `convox init` fails due to existing file(s)](https://trello.com/c/9E1rTHTF/1767-display-visible-error-when-convox-init-fails-due-to-existing-file-s)

PR #1651, opened 2017-02-01
NAT DependsOn KernelUser

mattmanning:
A CloudFormation stack update can fail if the NATs are updated before the KernelUserGeneralPolicy is created.

screen shot 2017-01-31 at 8 54 54 pm

This PR makes the NAT resources depend on the KernelUser (which depends on all its policies) to avoid this timing issue.

### DIFF
--- a/provider/aws/dist/rack.json
+++ b/provider/aws/dist/rack.json
@@ -961,7 +961,7 @@
     },
     "Nat0": {
       "Condition": "Private",
-      "DependsOn": "KernelUser",
+      "DependsOn": "KernelAccess",
       "Type": "AWS::EC2::NatGateway",
       "Properties": {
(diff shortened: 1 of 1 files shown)


mattmanning:
[Fix timing bug with NAT permissions](https://trello.com/c/x9kjKx3j/1780-fix-timing-bug-with-nat-permissions)

provider/aws/dist/rack.json
```diff
@@ -960,6 +961,7 @@
     },
     "Nat0": {
       "Condition": "Private",
+      "DependsOn": "KernelUser",
```
MiguelMoll: Should this be `KernelAccess` instead so the creds are generated? It has an implicit dependency on `KernelUser`

PR #1656, opened 2017-02-01
properly handle blank default vars in docker-compose.yml

ddollar:
This should allow users to specify environment in their docker-compose.yml like this:

```
- FOO=
```

and have a blank default value that they can override


ddollar:
[better handling of blank default env vars](https://trello.com/c/kbdYNuW0/1789-better-handling-of-blank-default-env-vars)

PR #1657, opened 2017-02-01
Build and Release Events

### LINKED ISSUE #1653: Build Event Status
Currently rack sends a "build:create" "success" event at the beginning of a build. It should instead send:

* "build:create" "start" at the beginning of the build
* "build:create" "success" at the end of the build
* "build:create" "error" if the build errors out for any reason

### LINKED ISSUE #1654: Release Event Status
Currently rack sends a "release:promoted" "success" event at the beginning of a release. It should instead send:

"release:promote" "start" at the beginning of the release
"release:promote" "success" at the end of the release
"release:promote" "error" if the release doesn't complete. This needs to capture CloudFormation rollbacks, etc.

MiguelMoll:
Usage of SendEvent increased to track build success and errors. Also adjusted Segment events accordingly.

Fixes #1653 
Fixes #1654

### DIFF
--- a/api/controllers/builds.go
+++ b/api/controllers/builds.go
@@ -34,9 +34,8 @@ func BuildCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 		Action: "build:create",
 		Status: "start",
 		Data: map[string]interface{}{
-			"app":  app,
-			"id":   "n/a",
-			"opts": opts,
+			"app": app,
+			"id":  "n/a",
 		},
 	}
 	models.Provider().EventSend(event, nil)
--- a/api/controllers/builds.go
+++ b/api/controllers/builds.go
@@ -33,7 +33,7 @@ func BuildCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 	event := &structs.Event{
 		Action: "build:create",
 		Status: "start",
-		Data: map[string]interface{}{
+		Data: map[string]string{
 			"app": app,
 			"id":  "n/a",
 		},
--- a/api/controllers/builds.go
+++ b/api/controllers/builds.go
@@ -53,6 +53,8 @@ func BuildCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 			return httperr.Server(err)
 		}
 
+		event.Status = "success"
+		models.Provider().EventSend(event, nil)
 		return RenderJson(rw, build)
 	}
 
--- a/api/controllers/builds.go
+++ b/api/controllers/builds.go
@@ -30,13 +30,26 @@ func BuildCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 		return httperr.Errorf(403, "endpoint deprecated, please update your client")
 	}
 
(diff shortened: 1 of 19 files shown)


MiguelMoll:
[Build Failure Notifications](https://trello.com/c/ytYZBtrd/1757-build-failure-notifications)

MiguelMoll:
[Build and Release Failure Segment Events](https://trello.com/c/RuYV8VVi/1758-build-and-release-failure-segment-events)

MiguelMoll:
[Release Failure Notifications and Logs](https://trello.com/c/biIZMOdH/1759-release-failure-notifications-and-logs)

MiguelMoll:
[Build and Release Notifcations and Segment Events](https://trello.com/c/RuYV8VVi/1758-build-and-release-notifcations-and-segment-events)

api/models/release.go
```diff
@@ -671,3 +668,93 @@ func waitForTemplate(bucket string, id string) error {
 	fmt.Printf("ns=kernel at=waitForTemplate.tick error=unknown bucket=%q release=%q\n", bucket, id)
 	return fmt.Errorf("unknown error")
 }
+
+// monitorReleasePromotion observes a release's rollout by verifying the app's CF stack status
+// Sends a notifcation on success or failure. This function blocks.
+// TODO: this should be in provider.ReleasePromote()
+func waitForPromotion(r *Release) {
```
ddollar: Can we do this with CloudFormation events?

We could create an SNS topic on the Rack that either calls back to a Rack HTTP endpoint (see webhooks) or have the Rack subscribe to an SQS queue that is subscribed to that SNS topic.

This should give us real "events" for CF updates (we could probably even use this to add things to the logs if we want to) without the polling which will fail if a rack container is restarted while waiting.

MiguelMoll: We definitely could use CF SNS events. We went with this method (knowing polling could fail) as the quickest way of getting promotion events.

Another option is lambda acting on CF events via SNS. One problem is finding a reliable way of knowing a stack update was a promotion or some other stack change (e.g scaling a process) without or little state management.

nzoschke: We should get this out with the simple polling strategy to start getting some visibility into releases. We will be able to use that data to drive another iteration of release notifications.

api/controllers/builds.go
```diff
@@ -30,31 +30,48 @@ func BuildCreate(rw http.ResponseWriter, r *http.Request) *httperr.Error {
 		return httperr.Errorf(403, "endpoint deprecated, please update your client")
 	}
 
+	event := &structs.Event{
+		Action: "build:create",
+		Status: "start",
+		Data: map[string]interface{}{
+			"app": app,
+			"id":  "n/a",
```
nzoschke: don't we have an ID generated at the start of the build?

MiguelMoll: Not in the model. We get an ID once the build is create by the provider.

MiguelMoll:
@ddollar @nzoschke @awsmsrc Please give this another look when you can.

PR #1601, opened 2017-01-23
Use Docker 1.13

soulshake:
TODO:
- [x] reorder docker API call stub-things
- [x] Update `.travis.yml` to run with docker 1.13
- [x] Install docker 1.13 in Dockerfile

PR #1663, opened 2017-02-02
Allow DOCKER_BIN to specify location of docker binary

soulshake:
This is a minimal implementation to demonstrate making the CLI use a `DOCKER_BIN` environment variable instead of just `docker`.

Currently, at various places in the code, the CLI grabs the output of `docker` commands.

If accepted (in a less hypothetical form), this PR would allow a user to set `CONVOX_ENV` to a fully-qualified path, e.g. `/usr/bin/docker12`, in order to easily switch between Docker versions. If not set, the output of `which docker` is used.

This is probably pretty pointless and unnecessary since the user can just alias `docker` to whatever binary they want. But I'm throwing it out there anyway, if only to raise the issue of the fact that **at some places in the code, a "different docker" is being used (namely, `github.com/fsouza/go-dockerclient`), which can cause unexpected contradictions, especially during tests.** Different versions of docker use different api endpoints (e.g. `/v1.24/auth` vs `/v1.25/auth`) and make extra calls (notably `/_ping` comes up in docker 1.13). But our docker stubs can only support one series of calls at a time, which vary from 1.12 to 1.13 (see https://github.com/convox/rack/pull/1601).

### DIFF
--- a/cmd/convox/stdcli/stdcli.go
+++ b/cmd/convox/stdcli/stdcli.go
@@ -25,6 +25,7 @@ var (
 	Querier    func(bin string, args ...string) ([]byte, error)
 	Spinner    *spinner.Spinner
 	Tagger     func() string
+	Docker     func() string
 )
 
 func init() {
@@ -35,6 +36,7 @@ func init() {
 	Runner = runExecCommand
 	Spinner = spinner.New(spinner.CharSets[9], 100*time.Millisecond)
 	Tagger = tagTimeUnix
+	Docker = getDockerBin
 
 	cli.AppHelpTemplate = `{{.Name}}: {{.Usage}}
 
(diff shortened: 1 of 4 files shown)


cmd/convox/stdcli/stdcli.go
```diff
@@ -264,6 +266,19 @@ func queryExecCommand(bin string, args ...string) ([]byte, error) {
 	return exec.Command(bin, args...).CombinedOutput()
 }
 
+func getDockerBin() string {
+	osd := os.Getenv("DOCKER_BIN")
+	if osd != "" {
+		return osd
+	}
+	wd, err := exec.Command("which", "docker").Output()
```
ddollar: Just `"docker"` should have `PATH` do the right thing here.

MiguelMoll: Are you saying it's safe to assume `docker` will be in the users `PATH`?

cmd/convox/stdcli/stdcli.go
```diff
@@ -264,6 +266,19 @@ func queryExecCommand(bin string, args ...string) ([]byte, error) {
 	return exec.Command(bin, args...).CombinedOutput()
 }
 
+func getDockerBin() string {
```
MiguelMoll: I think this function can go in the `main` package and be used by the various sub commands. Could also use the same method as with the COMPOSE_FILE where you have a `dockerBin` var.

cmd/convox/stdcli/stdcli.go
ddollar: If `which docker` works then `docker` was in the `PATH` to begin with.

MiguelMoll: That went over my head. Got it 👍

cmd/convox/stdcli/stdcli.go
soulshake: Good point; done.

soulshake:
[DOCKER_BIN](https://trello.com/c/QN2Mp60A/1826-docker-bin)

PR #1666, opened 2017-02-02
build args

ddollar:
This change allows a user to specify build args in their `Dockerfile` like so:

```
FROM ubuntu:16.04

ARG RAILS_ENV=production
ARG FOO
```

These args will be populated from the local environment during `convox start` and the app's environment during `convox build`


ddollar:
[build args](https://trello.com/c/kA9GHq69/1788-build-args)

soulware:
Does this open up the possibility of using build args in docker-compose.yml or is this completely orthogonal?

```
build:
  context: .
  args:
    password: sekret
```

ddollar:
That's not currently covered by this PR but it seems like it wouldn't be too hard to add on top.

dlanger:
Hooking this up to `docker-compose` (and `.env` files, if that's not already done) would make this great for build-time configuration of minor things.

mwarkentin:
@ddollar Any chance we could get a pre-release version of this to try out? I think this will be useful for a few of the problems we had, but would like to play around with it to verify.

ddollar:
It is hooked up to `.env` but not to `docker-compose.yml`

mattmanning:
[Build arguments](https://trello.com/c/oFUaHPqL/69-build-arguments)

PR #1667, opened 2017-02-02
Add "Accept: application/json" header to client requests (blocked by console #794)

soulshake:
- Add "Accept: application/json" header to client requests
- Add test for nonexistent API endpoints (blocked by console #794)

P.S. Hardcoding the expected reply in `TestApiGet404` felt like it defeated the purpose, so I did it the hard way, by making an actual request for a nonexistent page (which is why the test will fail until console #794 is merged). There must be an easier way to validate responses from Console. Unfortunately, I don't know what it is yet.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -433,6 +433,7 @@ func (c *Client) Request(method, path string, body io.Reader) (*http.Request, er
 	req.SetBasicAuth("convox", string(c.Password))
 
 	req.Header.Add("Content-Type", "application/json")
+	req.Header.Add("Accept", "application/json")
 	req.Header.Add("Version", c.Version)
 
 	if c.Rack != "" {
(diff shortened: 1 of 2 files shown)


soulshake:
[ERROR: invalid character '<' looking for beginning of value](https://trello.com/c/oBoZA554/1787-error-invalid-character-looking-for-beginning-of-value)

soulshake:
@ddollar Sorry, I'm having trouble seeing how to mock the test request without hardcoding the header I'm trying to test for. Is there an existing test you can point me to that I could copy from?

ddollar:
Seems like you only need the second half of your 404 test (the mocked version) and not the needs-internet version.

soulshake:
I updated it to include only the second half of the test, based on what I understood from your comment. It passes, but I really don't see how the test is useful in its current form.

Once #1684 has been merged I'll be able to set the Headers on test.Http, but that still only proves we can set headers on the _test_ request...

I'm truly sorry, I must be missing something :(

soulshake:
Update: good to go now.

MiguelMoll:
With a rebase we can get this one out.

Issue #1626, opened 2017-01-27
Use COMPOSE_FILE environment variable as default file when set


amsb:
Docker Compose uses the COMPOSE_FILE environment variable as the default when set:

https://docs.docker.com/compose/reference/envvars/#composefile

Doing the same with Convox would make it a bit easier to manage test/build/deploy scripts.

ddollar:
Doing a bit of feature request culling here. Please use the `-f` flag to `convox start` or `convox build`

PR #1670, opened 2017-02-03
Respect COMPOSE_FILE envvar

soulshake:
Respect `COMPOSE_FILE` environment variable in `convox start`, `convox doctor`, `convox build` and `convox deploy`.

Addresses #1626.

### DIFF
--- a/cmd/convox/doctor.go
+++ b/cmd/convox/doctor.go
@@ -149,23 +149,24 @@ func cmdDoctor(c *cli.Context) error {
 	}
 
 	stdcli.Writef("\n\n### Build: Service\n")
-	startCheck("<file>docker-compose.yml</file> found")
-	_, err := os.Stat("docker-compose.yml")
+	dcm := helpers.DetectComposeFile()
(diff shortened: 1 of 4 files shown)


cmd/convox/doctor.go
```diff
@@ -149,23 +149,24 @@ func cmdDoctor(c *cli.Context) error {
 	}
 
 	stdcli.Writef("\n\n### Build: Service\n")
-	startCheck("<file>docker-compose.yml</file> found")
-	_, err := os.Stat("docker-compose.yml")
+	dcm := helpers.DetectComposeFile()
```
MiguelMoll: This is nitpicking, but perhaps we could have one `dcm = helpers.DetectComposeFile()` variable up in the `var()` scope.

soulshake:
[Respect COMPOSE_FILE envvar](https://trello.com/c/sNt3fuDs/1807-respect-compose-file-envvar)

idris:
@soulshake will this work if you set `COMPOSE_FILE` in `.env`?

soulshake:
@idris No, `.env` is only applied to the containers' environment.

PR #1672, opened 2017-02-03
Don't treat latest version as required Rack update if user has requested a specific version

soulshake:
I'm concerned about how easily this would allow a user to skip a required update, but maybe that's enough of an edge case that it's OK.


soulshake:
[`convox rack update` to branch release forces incorrect required release](https://trello.com/c/qkYQ282M/1692-convox-rack-update-to-branch-release-forces-incorrect-required-release)

MiguelMoll:
I'm not sure we want to skip required updates even if a version is specified. That could prevent the Rack from updating and causing it to roll back. A required update has some stop gap changes for versions going forward.

ddollar:
We should take a different approach here. Releases that are actually required should always be stopped at or the Rack could end up permanently broken for certain weird changes.

We should instead make sure that the "latest" release is not automatically marked as "required" also which seems to be what is happening here.

soulshake:
@MiguelMoll @ddollar is this better?

PR #1677, opened 2017-02-05
Dedicated build instance

ddollar:
This change creates a new Rack parameter called `BuildInstance`. Set this parameter to an instance type to create a dedicated build instance (AutoScalingGroup of size 1).

`convox rack params set BuildInstance=c4.large`

### DIFF
--- a/provider/aws/dist/rack.json
+++ b/provider/aws/dist/rack.json
@@ -1375,6 +1375,7 @@
             "bootcmd:\n",
             "  - mkswap /dev/xvdb\n",
             "  - swapon /dev/xvdb\n",
+            "  - until yum install -y aws-cli; do echo \"Waiting for network\"; done;\n",
             "  - [ cloud-init-per, instance, docker_storage_setup, /usr/bin/docker-storage-setup ]\n",
             "  - echo ECS_CLUSTER=", { "Ref": "BuildCluster" }, " >> /etc/ecs/ecs.config\n",
             "  - echo ECS_ENGINE_AUTH_TYPE=docker >> /etc/ecs/ecs.config\n",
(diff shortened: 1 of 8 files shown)


provider/aws/dist/rack.json
```diff
@@ -1321,6 +1329,128 @@
...
+            "  - mkdir -p /etc/convox\n",
+            "  - echo -e '/var/log/docker {\\n  rotate 7\\n  daily\\n  nocompress\\n  copytruncate\\n}' >> /etc/logrotate.d/docker\n",
+            "  - export INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)\n",
+            "  - /opt/aws/bin/cfn-signal --stack ", { "Ref": "AWS::StackName" }, " --region ", { "Ref":"AWS::Region" }, " --resource BuildInstances\n"
+          ] ] }
+        }
+      }
+    },
+    "BuildInstances": {
+      "DependsOn": [ "AvailabilityZones", "Subnet0", "Subnet1" ],
```
mattmanning: Do we need something conditional for private subnets in private racks?

ddollar: It automatically happens by being referenced if the `Private` condition is met.

ddollar:
This should ideally always be 1 instance for best caching. If you need more space use a bigger instance.

mwarkentin:
Agreed, I think getting the caching benefits will be huge, and you can use a beefy instance type if needed. Could the instance potentially leverage an EBS volume to persist the cache even if the instance is terminated, etc.?

I can imagine that we'll be configuring nightly scale downs for this build "cluster" (maybe even in prod?).. could be nice if that didn't blow away the cache.

ddollar:
[dedicated build instance](https://trello.com/c/JpqmYNZn/1867-dedicated-build-instance)

provider/aws/dist/rack.json
```diff
@@ -1935,7 +2066,7 @@
               "AWS_REGION": { "Ref": "AWS::Region" },
               "AWS_ACCESS": { "Ref": "KernelAccess" },
               "AWS_SECRET": { "Fn::GetAtt": [ "KernelAccess", "SecretAccessKey" ] },
-              "CLIENT_ID": { "Ref": "ClientId" },
```
MiguelMoll: Was this removed on purpose? This was needed for build notifications.

MiguelMoll:
Needed these permissions while testing:

KernelUserGeneralPolicy

- autoscaling:CreateAutoScalingGroup
- ecs:CreateCluster

KernelUserPolicy 

```
        {
            "Action": "ecs:DeleteCluster",
            "Resource": [
                "arn:aws:ecs:*:*:cluster/{ "Ref": "AWS::StackName" }*"
            ],
            "Effect": "Allow",
            "Sid": "LimitedECSAccess"
        }
```

This one should be locked down
```
autoscaling:DeleteAutoScalingGroup
```

PR #1680, opened 2017-02-06
Prevent terminal from hanging on ssh error

MiguelMoll:
When trying to SSH into an instance that doesn't exist or no ssh key, the terminal would hang. This change injects the status code prefix in the error so the client can parse and act accordingly.


MiguelMoll:
[Do not hang terminal on no ssh key error](https://trello.com/c/ZBGGjkrP/1820-do-not-hang-terminal-on-no-ssh-key-error)

MiguelMoll:
Going in a different direction via #1686

Issue #1520, opened 2017-01-06
'convox apps delete' doesn't respect the --rack flag


soulshake:
In the commands below, my active rack is `personal/legit`:
```
$ convox rack
Name     legit
Status   running
[...]
```

`personal/legit` has an app called `cv-soulshake-net`.
I want to delete an app by the same name on a different rack, `personal/dev`.

```
$ convox apps --rack personal/dev
APP               STATUS
console-staging   running
cv-soulshake-net  failed

$ convox apps --rack personal/dev delete cv-soulshake-net
Deleting cv-soulshake-net... DELETING
```

The app is deleted on `personal/legit` instead of `personal/dev`:

```
$ convox apps --rack personal/dev
APP               STATUS
console-staging   running
cv-soulshake-net  failed

$ convox apps 
APP                 STATUS
blog-soulshake-net  running
cv-soulshake-net    deleting
flask               running
```

dlanger:
Related to https://github.com/convox/rack/issues/1107?

soulshake:
Thanks @dlanger 
Closing as dupe of #1107

Issue #1202, opened 2016-09-08
the services create cli command doesn't support the rackflag


awsmsrc:
this is because we intentionally skip flag parsing due to all the different flags the different service types handle

this is the cause of: https://github.com/convox/console/issues/513

redroot:
+1 just experienced this as well

soulshake:
Why do we skip flag parsing instead of using `--` like in bash builtins?

From `man bash`:

```
--        A -- signals the end of options and disables further option processing.  Any arguments
          after  the  -- are treated as filenames and arguments.
```

ddollar:
Deupe of #1107

Issue #1107, opened 2016-08-19
--rack argument isn't always respected


dlanger:
There are a number of `convox` commands which are documented (according to `convox help`) to accept the `--rack` argument, but which silently ignore it and use the current rack.

Here's an example with the `convox ssl` command. In it, I:
1. Log into the `stage` rack using `convox switch`
2. Update the SSL certificate on the `stage` rack
3. Try and update the SSL certificate on the `kraken` rack, using the `--rack` argument - but get an error saying there's nothing to be done
4. Switch to the `kraken` rack using `convox switch`
5. Issue the same command as in #3, but in this case the update is successful

I'd expect #2 to work, as I'm using the `--rack` command to specify which rack I'd like the command to impact.

```
$ convox --version
client: 20160818150650
$ convox switch
wave/stage
$ convox ssl update main:443 acm-9b81e07ebc40
Updating certificate... OK
$ convox ssl --rack kraken update main:443 acm-9b81e07ebc40
Updating certificate... ERROR: ValidationError: No updates are to be performed.
        status code: 400, request id: b445b025-6663-11e6-bbe2-a10804931323
$ convox switch kraken
Switched to wave/kraken
$ convox ssl --rack kraken update main:443 acm-9b81e07ebc40
Updating certificate... OK
```

I've seen this in a few other places which I don't recall - I'll update if I find them again.

dlanger:
Also #1202

dlanger:
The order in which you use the `--rack` is also meaningful, and doesn't provide helpful feedback when you do it wrong.

As an example, this:
```
convox api --rack RACKNAME get /apps/APPNAME
```
and
```
convox api get /apps/APPNAME --rack RACKNAME
```
do different things.

The first one makes the API call against whatever your current rack is, and ignores the `--rack` argument passed in (and doesn't tell you that it's being ignored). The second one makes the API call against `RACKNAME`.

soulshake:
This should be resolved by #1684, please let us know if you encounter the misbehavior again.

PR #1684, opened 2017-02-06
Recover --rack from os.Args if urfave/cli has dropped it

soulshake:
## Flag handling changes

**Current behavior**: If `--rack` is passed in position 0, convox (via urfave/cli) ignores it. 
**PR behavior:** When coalescing the active rack, if urfave/cli thinks the rack flag is empty, we double-check in the os.args via `getRackFlag()`.

@noah added `rack` and `app` app.Flags at the CLI level.
(There might therefore be some redundancy in `getRackFlag()`; need to check.) 

## Testing changes

To test this, we needed to add a `rack` header to the testServer. That wasn't possible before, so @noah added `Headers` to the `Http` struct in `test/http.go`. (I think we could have accomplished this via the `CONVOX_RACK` env var, but being able to manipulate headers is useful anyway.)

Resolves:
* #1107
* #1202
* #1520 
* https://github.com/convox/console/issues/513

### DIFF
--- a/api/controllers/apps_test.go
+++ b/api/controllers/apps_test.go
@@ -24,7 +24,7 @@ func TestAppList(t *testing.T) {
 	aws := test.StubAws(test.DescribeStackCycleWithoutQuery("convox-test-bar"))
 	defer aws.Close()
 
-	body := test.HTTPBody("GET", "http://convox/apps", nil)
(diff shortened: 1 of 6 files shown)


soulshake:
[Respect --rack flag regardless of position](https://trello.com/c/XI6kT9aj/1825-respect-rack-flag-regardless-of-position)

PR #1686, opened 2017-02-06
Handle case where terminal doesn't enter raw mode

mattmanning:
In some cases when `terminal.MakeRaw` is called the terminal will not go into raw mode and `nil` will be returned as the state. If `terminal.Restore` is later called with a nil state the CLI will panic. This change checks for a nil state and does not call `terminal.Restore`.

This also adds back a call to close a PipeWriter in the case of validation errors, etc so that it doesn't hang open.


mattmanning:
[Handle case where terminal doesn't enter raw mode](https://trello.com/c/mT7fKvhO/1821-handle-case-where-terminal-doesn-t-enter-raw-mode)

PR #1688, opened 2017-02-06
Prevent terminal from hanging on ssh error

MiguelMoll:


MiguelMoll:
[SSH key not found hangs terminal](https://trello.com/c/I3iODkON/1822-ssh-key-not-found-hangs-terminal)

PR #1692, opened 2017-02-07
drain container instance on termination

MiguelMoll:
- [x] Test cron/one-off jobs
- [x] Needs time out? Lambda has a 5 minute max


provider/aws/lambda/lifecycle/main.go
```diff
@@ -114,6 +129,32 @@ func handle(r Record) error {
 	return nil
 }
 
+func waitForInstanceDrain(cluster, ci string) error {
+
```
ddollar: Is it possible to use one of the Waiters in the SDK for this?

MiguelMoll: I believe we can and worth a shot. I went this route first to avoid any state changes that happen when calling the Waiter with tasks grabbed before or after instance change to `DRAINING`.

MiguelMoll:
SDK states:
```Any PENDING or RUNNING tasks that do not belong to a service are not affected; you must wait for them to finish or stop them manually.```

Have to handle this case for any scheduled jobs or one off processes that happen to be running. We might have to stop them, as unfortunate as that is. Maybe wait for the lifecycle/lambda (10mins/5mins) timeout?

MiguelMoll:
[Container Instance Draining](https://trello.com/c/Aq0eGhgC/1847-container-instance-draining)

PR #1693, opened 2017-02-07
make NATs depend on KernelUser

mattmanning:
screen_shot_2017-02-06_at_10 52 16_am


mattmanning:
[Make NATs depend on KernelUser](https://trello.com/c/aIL1ZriR/1836-make-nats-depend-on-kerneluser)

PR #1694, opened 2017-02-07
Paginate stack listing

mattmanning:
Accounts with large amounts of CloudFormation stacks can see only partial results for `convox resources` or `convox apps`.

This change pages through the full listing of stacks to make sure the result is complete.


mattmanning:
[List more than 30 resources](https://trello.com/c/g49p4NyX/1727-list-more-than-30-resources)

PR #1909, opened 2017-02-08
build cache

ddollar:
This change allows you to persist certain artifacts between docker builds. This is useful for speeding up things like `bundle install`.

The new behavior:

* Any files residing in `/var/cache/build` in your image will be extracted after your build is complete and stored at the application level.
* Future builds will see the contents of this cached directory injected into `.cache/build` during the build process.

An example of how to use this with `bundle install`:

````
# bundle install
COPY Gemfile      /app/Gemfile
COPY Gemfile.lock /app/Gemfile.lock
COPY .cache/build/gems /var/cache/build/gems
RUN bundle install --path /var/cache/build/gems
````

The next time you need to touch your Gemfile.lock you won't need to completely rebuild your entire Gemfile as the local artifacts will already be present.

### DIFF
--- a/api/cmd/build/main.go
+++ b/api/cmd/build/main.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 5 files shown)


nzoschke:
Some next steps:

- [ ] Review design of .cache directory
- [ ] QA on Mac. Start and build of a rails app
- [ ] Testing on Windows
- [ ] Testing on Linux
- [ ] Replace shell outs with go code

mwarkentin:
I'd be happy to test this out with some of our django apps as well if there's a pre-release build available.

MiguelMoll:
Completely agree. Forgot to remove it during testing.

MiguelMoll:
Things to consider:
- [ ] Should we parse the Docker file for `COPY .cache/build/foo` to indicate if we should use build case. Otherwise misleading (worrisome?) errors are displayed to the user: ```build    │ Error response from daemon: lstat /var/lib/docker/aufs/mnt/00a07473df28d3a1cf02e83a408d9b5870b36634271ec308b660b29129f271f8/var/cache/build: no such file or directory
build    │ cache error: exit status 1```

- [ ] Funky order to get this working unless I'm missing something. The cache has to be generated the first time (e.g. `RUN bundle install --path /var/cache/build/gems`) before the Dockerfile should specify the `COPY .cache` command. This goes for `convox start` and `convox build`.

nzoschke:
[Build Cache](https://trello.com/c/w68PF75n/1900-build-cache)

nzoschke:
[Build Cache](https://trello.com/c/w68PF75n/1900-build-cache)

nzoschke:
How does this work with a `COPY . /app` command?

`myapp/.cache` gets picked up in that copy and is noticeably slower on a big gemset.

Adding `.cache/` to `.dockerignore` means a `COPY .cache/build/ /var/cache/build/` doesn't work.

FWMatt:
I think this is a great idea.

My initial idea when building a Rails app was to use a volume from docker-compose.yml to store gems. However volumes don't appear to be accessible from the Dockerfile?

nzoschke:
@FWMatt that's right, volumes aren't available in Dockerfile and during `docker build`. So we are experimenting with a bit of external tooling so we can `COPY` the cache into and out of the build images

nzoschke:
QA Checklist

- [ ] build CLI from build-cache branch
- [ ] `rails new --database=postgres`
- [ ] `convox init` generates Dockerfile with build-cache COPY directives
- [ ] `convox start` builds gems
- [ ] ~/.convox/ has archive of gems
- [ ] Touch Gemfile.lock
- [ ] `convox start` re-builds gems instantly due to cache

nzoschke:
It's working pretty well for rails. Not so great for node.js. Some things I observed

- There is a noticeable delay before the first build output while cache is being moved around. Maybe a friendly `build | Copying build cache...` message first.
- The myapp/.cache directory can be confusing. @ddollar suggests copying app and cache data together into a /tmp directory to avoid polluting the actual app dir
- On a node app ([ghost-on-heroku](https://github.com/cobyism/ghost-on-heroku)), the second build is very slow on the `mv /var/cache/build/buildpack/node/node_modules /app/node_modules` buildpack step inside the container
- The second build fails with a `sh: 1: ncp: not found`

FWMatt:
Is this being merged or is there a replacement?

ddollar:
The changes were carried over into convox/praxis

PR #1916, opened 2017-02-09
only log the command to run

MiguelMoll:


MiguelMoll:
[Only log command to run](https://trello.com/c/cvtAohIH/1868-only-log-command-to-run)

awsmsrc:
I'm ok with this, but it seems a little weird that this will output 

`running: docker` whether it is a push, pull, tag or whatever

ddollar:
This is going to make it harder to debug Convox in some cases. It was handy to be able to run the docker commands yourself to see what was going on.

awsmsrc:
how about printing only if DEBUG is turned on, otherwise printing nothing?

MiguelMoll:
Think this should be printed if `DEVELOPMENT` == true

soulshake:
I think the `running:` line is useful information. It's only one line, and it really helps make it clear what's going on under the hood. I think we should leave it in, personally.

Another thing: is it `DEBUG` or `CONVOX_DEBUG`? I see references to both in the code. `CONVOX_DEBUG` seems more appropriate, IMO.

MiguelMoll:
I believe `CONVOX_DEBUG` is being phased out. Only see two instances of it in `sync.go`.

ddollar:
Maybe w could test the verbose vs non-verbose output?

PR #1917, opened 2017-02-09
build events includes id, timestamp and from

nzoschke:


provider/aws/event.go
```diff
@@ -26,6 +26,13 @@ import (
 func (p *AWSProvider) EventSend(e *structs.Event, err error) error {
 	e.Timestamp = time.Now().UTC()
 
+	if e.Data["timestamp"] != "" {
+		t, err := time.Parse(time.RFC3339, e.Data["timestamp"])
+		if err != nil {
```
MiguelMoll: Think this is `err == nil` instead?

nzoschke: Good catch. Fixed

PR #1920, opened 2017-02-10
Error when duplicate cron labels are present.

soulshake:
The warning isn't easily visible in (and doesn't match the formatting of) the output of `convox start`. Is there a helper for printing warnings?

### DIFF
--- a/manifest/yaml.go
+++ b/manifest/yaml.go
@@ -280,6 +280,9 @@ func (l *Labels) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 			switch len(parts) {
 			case 2:
+				if _, ok := (*l)[parts[0]]; ok {
+					return fmt.Errorf("duplicate cron label %v", parts[0])
+				}
 				(*l)[parts[0]] = parts[1]
 			default:
 				return fmt.Errorf("cannot parse label: %v", t)
--- a/cmd/convox/deploy_test.go
+++ b/cmd/convox/deploy_test.go
@@ -56,3 +56,20 @@ func TestDeployInvalidCronInManifest(t *testing.T) {
 		},
 	)
 }
+
+func TestDeployDuplicateCronInManifest(t *testing.T) {
+	ts := testServer(t,
+		test.Http{Method: "GET", Path: "/apps/foo", Code: 200, Response: client.App{Name: "foo", Status: "running"}},
+	)
+
+	defer ts.Close()
+
+	test.Runs(t,
+		test.ExecRun{
+			Command: "convox deploy --file docker-compose.duplicate-cron-label.yml --app foo",
+			Dir:     "../../manifest/fixtures",
+			Exit:    1,
+			Stderr:  "ERROR: invalid docker-compose.duplicate-cron-label.yml: error loading manifest: duplicate cron label convox.cron.myjob\n",
+		},
+	)
+}
--- a/cmd/convox/start_test.go
+++ b/cmd/convox/start_test.go
@@ -23,3 +23,19 @@ func TestStartInvalidManifest(t *testing.T) {
 		},
 	)
 }
+func TestStartDuplicateCronInManifest(t *testing.T) {
+	ts := testServer(t,
+		test.Http{Method: "GET", Path: "/apps/foo", Code: 200, Response: client.App{Name: "foo", Status: "running"}},
+	)
+
+	defer ts.Close()
+
+	test.Runs(t,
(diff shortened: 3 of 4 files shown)


soulshake:
[Validate against duplicate cron job names](https://trello.com/c/gu0lLNGA/1813-validate-against-duplicate-cron-job-names)

awsmsrc:
@soulshake this is fine but with regards to your question we could add stdcli.Warn?

look at cmd/convox/stdcli/write.go

the number represents the color: https://www.google.co.uk/imgres?imgurl=http://misc.flogisoft.com/_media/bash/colors_format/256_colors_bg.png&imgrefurl=http://misc.flogisoft.com/bash/tip_colors_and_formatting&h=444&w=642&tbnid=93SjXqsTjP1yxM:&vet=1&tbnh=146&tbnw=210&docid=Lw9El7UXPqDhLM&usg=__4eSZdOqbXcCNUMYRzNLZOOjw4Vg=&sa=X&ved=0ahUKEwixt-3v_oXSAhVHDsAKHfDqA4cQ9QEIITAA#h=444&imgrc=93SjXqsTjP1yxM:&tbnh=146&tbnw=210&vet=1&w=642

soulshake:
+1 @awsmsrc I created a separate task to implement that.

MiguelMoll:
Any thoughts on returning an error instead? Seems odd to have CLI output in the Unmarshal func. Maybe we can keep it here for now as a deprecation notice?

soulshake:
@MiguelMoll Good point...

I'm not seeing a way to "warn" the user outside of `UnmarshalYAML` without aborting completely. I'm having trouble even find a place where I could do an extra check on the error returned by `UnmarshalYAML`, like `if strings.Contains(err, "Duplicate label found"){ [warn user but don't abort] }`

Do you have a suggestion?

MiguelMoll:
I believe it would be returned in one of these two calls: https://github.com/convox/rack/blob/master/manifest/manifest.go#L44-L50

I believe yaml.Unmarshal checks if a type implements the UnmarshalYAML, if so it calls that.

soulshake:
@MiguelMoll I took another approach; is this any better?

Unrelated, but something I just noticed: after loading the manifest in `convox start`, we run `Validate()` on it [here](https://github.com/convox/rack/blob/master/cmd/convox/start.go#L91). We load the manifest in a similar way in `convox deploy`, but we don't `Validate()` it [here](https://github.com/convox/rack/blob/master/cmd/convox/deploy.go#L60
). Is that intentional?

(By the way, `Validate()` sounds like it would be a nice place to do the check for duplicate labels, but the duplicates are already long discarded by then.)

MiguelMoll:
Good catch, `convox deploy` should call `Validate()` too.

My fault though as I should've been more specific. My original idea was for `UnmarshalYAML` to return an error instead of warning and let that bubble up to the user as any other error. That way they can make any changes as needed. And we don't have to check everywhere for this specific error.

```
case 2:
 if _, ok := (*l)[parts[0]]; ok {
    return fmt.Errorf("duplicate label found for: %s\n", s)
}
(*l)[parts[0]] = parts[1]
```

soulshake:
@mattmanning I think you had mentioned a warning was preferable to an outright error in the case of duplicate cron labels. I think I agree with @MiguelMoll that an error is probably best here, given the complexity that would have to be added to make a warning possible. WDYT?

mattmanning:
@soulshake I'm good with an error, and it's probably actually better UX anyway.

manifest/yaml.go
```diff
@@ -280,6 +280,9 @@ func (l *Labels) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 			switch len(parts) {
 			case 2:
+				if dup, ok := (*l)[parts[0]]; ok {
+					return fmt.Errorf("duplicate cron label %v", dup)
```
MiguelMoll: Could we use `parts[0]` (the key) instead of `dup`? 

`duplicate cron label convox.cron.myjob` seems easier to read than `duplicate cron label * * * * ? bin/myjob1`

mattmanning: +1 was just testing this and thinking the same thing

mattmanning:
Works great for me now. Thanks!

MiguelMoll:
This PR caused a few conflicts so it missed this release. When you can, rebase so we can get it in the next one.

soulshake:
@MiguelMoll rebase done

PR #1924, opened 2017-02-10
guard against nil values from old task definitions

ddollar:


MiguelMoll:
Is this ready for review? Also looks like it needs a rebase.

PR #1936, opened 2017-02-14
Validate manifest file when deploying

soulshake:
Before:

```
$ convox start
ERROR: Cron task my_job is not valid (cron names can contain only alphanumeric characters, dashes and must be between 4 and 30 characters)

$ convox deploy
Deploying flask
Creating tarball... OK
[...]
```

After:

```
$ convox start
ERROR: Cron task my_job is not valid (cron names can contain only alphanumeric characters, dashes and must be between 4 and 30 characters)

$ convox deploy
ERROR: Cron task my_job is not valid (cron names can contain only alphanumeric characters, dashes and must be between 4 and 30 characters)
```

Ref: As discussed in https://github.com/convox/rack/pull/1920#issuecomment-279531409

### DIFF
--- a/cmd/convox/deploy.go
+++ b/cmd/convox/deploy.go
@@ -61,8 +61,17 @@ func cmdDeploy(c *cli.Context) error {
 	}
 
 	// validate docker-compose
-	if _, err := manifest.LoadFile(c.String("file")); err != nil {
+	m, err := manifest.LoadFile(c.String("file"))
+	if err != nil {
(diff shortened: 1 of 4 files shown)


soulshake:
[`convox deploy` should fail if envvars are missing](https://trello.com/c/Gd3pg2QZ/1923-convox-deploy-should-fail-if-envvars-are-missing)

soulshake:
[Validate manifest when deploying](https://trello.com/c/dENQljzO/1934-validate-manifest-when-deploying)

PR #1937, opened 2017-02-14
Install flags for a build instance

MiguelMoll:


MiguelMoll:
[Install flags for build instance](https://trello.com/c/F4y5eU6I/1935-install-flags-for-build-instance)

cmd/convox/install.go
```diff
@@ -83,6 +83,12 @@ func init() {
 				Value: "",
 				Usage: "custom AMI for rack instances",
 			},
+			cli.StringFlag{
+				Name:   "build-instance",
```
nzoschke: This is a style call, but this option is parallel to the `instance-type` option which raises some questions.

- [ ] Flag could be `build-instance-type`
- [ ] EnvVar could be RACK_BUILD_INSTANCE_TYPE
- [ ] Why an EnvVar for this at all?
- [ ] Why do some EnvVars use `RACK_` and others don't?

ci/region.sh
```diff
@@ -12,6 +13,7 @@ case "$1" in
   2)
     echo AWS_DEFAULT_REGION=us-west-1
     echo AWS_REGION=us-west-1
+    echo RACK_BUILD_INSTANCE=m4.xlarge
```
nzoschke: I'm guessing it's intentional to not use a build instance in some regions, and to use different sizes here? Maybe a comment stating this intent?

nzoschke:
Also, what about the other things to support this PR:

- [ ] Console install advanced option
- [ ] Docs

nzoschke:
I don't see why we need RACK_ or CONVOX_ prefixes on any of the EnvVar options. The could be simplified to:

- EMAIL
- PASSWORD
- BUILD_INSTANCE_TYPE
- PRIVATE
- STACK_NAME
- VERSION

and I'm not sure why BUILD_INSTANCE_TYPE has an EnvVar. Is that just for CI?

MiguelMoll:
I like the the RACK_ prefix and acts as a namespace to not clash with other env vars. Same as AWS_ is used. Should definitely pick one or the other. RACK vs CONVOX

RACK_BUILD_INSTANCE has a env var for completeness, which we happen to use for CI too. I believe most if not all of the env vars could be removed and just use CLI args. But that seems weird and it's nice to give the user an option of one over the other. I'm indifferent to the actual name. I considered appending type but seemed very long.

PR #1939, opened 2017-02-15
Improve uninstall error

ianmalott:
If the stack corresponding to the target rack is not found, mention
AWS environment variables and AWS CLI credentials file in case the
wrong AWS account was accidentally used.

### DIFF
--- a/cmd/convox/uninstall.go
+++ b/cmd/convox/uninstall.go
@@ -91,7 +91,7 @@ func cmdUninstall(c *cli.Context) error {
 
 	// verify that rack was detected
 	if len(stacks.Rack) == 0 || stacks.Rack[0].StackName != rackName {
-		return stdcli.Error(fmt.Errorf("can not find rack named %s", rackName))
+		return stdcli.Error(fmt.Errorf("can not find stack named %s\nAre you authenticating with the correct AWS account?\nSee AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and ~/.aws/credentials.", rackName))
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/convox/uninstall.go
```diff
@@ -91,7 +91,7 @@ func cmdUninstall(c *cli.Context) error {
 
 	// verify that rack was detected
 	if len(stacks.Rack) == 0 || stacks.Rack[0].StackName != rackName {
-		return stdcli.Error(fmt.Errorf("can not find rack named %s", rackName))
+		return stdcli.Error(fmt.Errorf("can not find stack named %s\nAre you authenticating with the correct AWS account?\nSee AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and ~/.aws/credentials", rackName))
```
MiguelMoll: Believe this should be `Rack` instead of `stack` as it's an implementation detail. The latter might confuse users who are not familiar with the term.

ianmalott: Happy to revert the change. I switched to "stack" to match the CLI help text for `uninstall`:

```
❯ convox help uninstall
convox uninstall: uninstall a convox rack

Usage:
  convox uninstall <stack-name> <region> [credentials.csv]
...
```

Seems like we kind of force users to be familiar with the implementation details here already. I will switch it back and let others discuss it later, though.

MiguelMoll: That's good point. At some time, a CLI pass would be nice to have consistency. Ideally remove references to stack.

PR #1941, opened 2017-02-15
allow customization of ELB Connection Draining Timeout

### LINKED ISSUE #1350: allow configuration of ELB ConnectionDraining Timeout
One way to speed up deploys is to lower the `ConnectionDraining.Timeout` attribute of the ELB from its [default of 5 minutes](http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/config-conn-drain.html#enable-conn-drain).  Let's expose another docker label that allows me to customize this value.

(I plan to work up a PR)

bobzoller:
fixes #1350 

- still defaults to 60 seconds
- uses Docker label `convox.draining.timeout`
- see also http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/config-conn-drain.html#enable-conn-drain


MiguelMoll:
This looks great. 👍 I'll run it through some tests and hopefully get this through.

We currently have a `convox.balancers` label. Any thoughts using `convox.balancers.draining` for this setting?

bobzoller:
cool :+1:

happy to have you change the name of the label... no preference here.

ddollar:
`convox.draining.timeout` has parity with `convox.idle.timeout` and `convox.health.timeout`

PR #1943, opened 2017-02-16
Don't use own ECR as private registry

soulshake:
Before:
```
$ convox registries add 332653055745.dkr.ecr.us-east-1.amazonaws.com --username foo --password bar
ERROR: unable to authenticate
```

After:
```
$ convox registries add 332653055745.dkr.ecr.us-east-1.amazonaws.com --username foo --password bar
ERROR: can't add own ECR host as a private registry: 332653055745.dkr.ecr.us-east-1.amazonaws.com
```

Notes:

- This won't work if the ECR host has a `http://` or `https://` prefix.
- Maybe the error is a bit too specific and something like "invalid registry host" would be better.
- I'm having a hard time with the cycles on the provider-side tests.
- I'm having a hard time with manual testing since I'm not even sure what credentials one would need to provide in order to add one's own Rack ECR as a registry.

### DIFF
--- a/cmd/convox/registries_test.go
+++ b/cmd/convox/registries_test.go
@@ -1,0 +1,182 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/convox/rack/client"
+	"github.com/convox/rack/test"
+)
+
+func TestRegistriesList(t *testing.T) {
+	tr := testServer(t,
+		test.Http{
+			Method: "GET",
+			Path:   "/registries",
+			Code:   200,
+			Response: client.Registries{
+				client.Registry{
+					Server:   "testRegistryServer",
+					Username: "testRegistryUser",
+					Password: "testRegistryPassword",
+				},
+			},
+		},
+	)
+
+	defer tr.Close()
+
+	test.Runs(t,
+		test.ExecRun{
+			Command: "convox registries",
+			Exit:    0,
+			Stdout:  "SERVER              USERNAME\ntestRegistryServer  testRegistryUser\n",
+		},
+	)
+}
+
+/*
+func TestRegistriesAddStdin(t *testing.T) {
+	tr := testServer(t,
+		test.Http{
+			Method: "POST",
+			Path:   "/registries",
+			Body:   "email=&password=testRegistryPassword&server=https%3A%2F%2Findex.docker.io%2Fv1%2F&username=testRegistryUser",
+			Code:   200,
+		},
+	)
+
+	defer tr.Close()
+
+	test.Runs(t,
+		test.ExecRun{
+			Command: "convox registries add https://index.docker.io/v1/",
+			Stdin:   "testRegistryUser\ntestRegistryPassword\n",
+			Exit:    0,
+			Stdout:  "Username: Password: \nDone.\n",
+		},
+	)
+}
+*/
+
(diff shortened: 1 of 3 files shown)


soulshake:
[Validate against adding a Rack's own ECR as a private registry](https://trello.com/c/1xq6Lpfu/1766-validate-against-adding-a-rack-s-own-ecr-as-a-private-registry)

provider/aws/registries.go
```diff
@@ -36,8 +57,12 @@ func (p *AWSProvider) RegistryAdd(server, username, password string) (*structs.R
 	// validate login
 	switch {
 	case regexpECRHost.MatchString(server):
+		if t := isOurOwnECR(p, server); t == true {
+			return nil, fmt.Errorf("can't add own ECR host as a private registry: %s", server)
```
ddollar: `can't add the rack's internal registry`

provider/aws/registries.go
```diff
@@ -13,6 +13,27 @@ import (
 	docker "github.com/fsouza/go-dockerclient"
 )
 
+func isOurOwnECR(p *AWSProvider, server string) bool {
```
ddollar: `host := fmt.Sprintf("%s.dkr.ecr.%s.amazonaws.com", a.Outputs["RegistryId"], p.Region)`

Found this builds.go in the provider. Should be able to use that to match specifically.

provider/aws/registries.go
MiguelMoll: Slight tangent, do we mind if it's `rack's` or `Rack's`?

soulshake: @MiguelMoll From what I see based on `egrep -r Errorf ./*  | grep -i rack` it looks like it's lowercase almost everywhere.

provider/aws/registries.go
```diff
@@ -36,6 +36,17 @@ func (p *AWSProvider) RegistryAdd(server, username, password string) (*structs.R
 	// validate login
 	switch {
 	case regexpECRHost.MatchString(server):
+		system, err := p.describeStack(p.Rack)
+		if err != nil {
+			return nil, err
+		}
+		sid := regexpStackID.FindStringSubmatch(*system.StackId)[2]
```
ddollar: Missing bounds check

provider/aws/registries.go
```diff
@@ -36,6 +36,17 @@ func (p *AWSProvider) RegistryAdd(server, username, password string) (*structs.R
...
+		host := fmt.Sprintf("%s.dkr.ecr.%s.amazonaws.com", sid, p.Region)
```
ddollar: This diverges from how the build process determines the app's registry. This seems in danger to diverge. Should fix one or both sides or at least add clarifying comments to both sides.

cmd/convox/registries_test.go
```diff
@@ -0,0 +1,182 @@
...
+		test.ExecRun{
+			Command: "convox registries",
+			Exit:    0,
+			Stdout:  "SERVER              USERNAME\ntestRegistryServer  testRegistryUser\n",
+		},
+	)
+}
+
+/*
+func TestRegistriesAddStdin(t *testing.T) {
```
ddollar: What's up with this test?

soulshake:
Updated to address @ddollar feedback

Issue #1185, opened 2016-09-02
check cloudformation status before going through with deploy


contra:
Situation: CF has in a failed state, user runs deploy.

Current result: Convox spends a bunch of time building docker images and whatnot, then fails at the very end.

Ideal result: Convox checks the CF status before doing anything, and fails immediately with a nice message.

( as per discussion w/ @ddollar )

PR #1944, opened 2017-02-16
block release promotion if app is updating

awsmsrc:
fixes https://github.com/convox/rack/issues/1185

this behavior was largely already in place

### DIFF
--- a/cmd/convox/deploy.go
+++ b/cmd/convox/deploy.go
@@ -51,9 +51,9 @@ func cmdDeploy(c *cli.Context) error {
 
 	switch a.Status {
 	case "creating":
-		return stdcli.Error(fmt.Errorf("deployment blocked, %s is still being created, check `convox apps info`", app))
+		return stdcli.Error(fmt.Errorf("app %s is still being created, check `convox apps info`", app))
 	case "updating":
-		return stdcli.Error(fmt.Errorf("deployment blocked, %s is already updating, check `convox apps info`", app))
+		return stdcli.Error(fmt.Errorf("app %s is still being updated, check `convox apps info`", app))
 	}
 
 	if !helpers.Exists(c.String("file")) {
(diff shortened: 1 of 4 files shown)


cmd/convox/deploy.go
```diff
@@ -51,9 +51,9 @@ func cmdDeploy(c *cli.Context) error {
 
 	switch a.Status {
 	case "creating":
-		return stdcli.Error(fmt.Errorf("app %s is still being created", app))
+		return stdcli.Error(fmt.Errorf("deployment blocked, %s is still being created, check `convox apps info`", app))
 	case "updating":
-		return stdcli.Error(fmt.Errorf("app %s is being updated", app))
+		return stdcli.Error(fmt.Errorf("deployment blocked, %s is already updating, check `convox apps info`", app))
```
MiguelMoll: nit: `deployment blocked, foo is still being updated, ...`. Seems consistent with the error above. Thoughts?

awsmsrc: was gonna ask @nzoschke's thoughts on the same thing here, the important thing is pushing towards "check convox apps info"

nzoschke: Yeah we don't need "deployment blocked"

ERROR: app foo is still being created, check `convox apps info`.

cmd/convox/releases.go
```diff
@@ -139,6 +139,15 @@ func cmdReleasePromote(c *cli.Context) error {
 		return stdcli.Error(err)
 	}
 
+	a, err := rackClient(c).GetApp(app)
+	if err != nil {
+		return stdcli.Error(err)
+	}
+
+	if a.Status != "running" {
+		return stdcli.Error(fmt.Errorf("promotion blocked, %s is already updating, check `convox apps info`", app))
```
nzoschke: same here

ERROR: foo is already updating, check `convox apps info`

PR #1945, opened 2017-02-16
app/rack status reads failed when rollback fails and more

### LINKED ISSUE #1266: status "rollback_failed" shows as "running"
When in a state when an app deploy fails and a rollback also fails, CloudFormation is in status `rollback_failed`. However, convox shows that as `running`, and doesn't prevent new deploys.

Of course, they can't deploy, but `convox apps info` should reflect that status so you know something is wrong and you have to go into cloudformation and force the rollback.

Ideally, this rollback forcing should be automatic or at least doable from convox CLI...

### LINKED ISSUE #1866: `convox env` help text improvements
The help text says to run `convox env get` to "get all environment variables":

```
$ convox env -h | grep get
  get		get all environment variables
```

But it doesn't work as expected:

```
$ convox env get
ERROR: No variable specified
```

The help text should be updated here. The commands work as we wish, the help text should just say that `get` is for getting a specific variable. `convox env` with no subcommand or other arguments gets all the variables.

https://trello.com/c/lxbVU9uW/369-convox-env-help-text-improvements

MiguelMoll:
Fixes #1266
Fixes #1866

### DIFF
--- a/cmd/convox/env.go
+++ b/cmd/convox/env.go
@@ -24,7 +24,7 @@ func init() {
 		Subcommands: []cli.Command{
 			{
 				Name:        "get",
-				Description: "get all environment variables",
+				Description: "get an environment variable",
 				Usage:       "VARIABLE",
 				Action:      cmdEnvGet,
 				Flags:       []cli.Flag{appFlag, rackFlag},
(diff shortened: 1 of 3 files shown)


MiguelMoll:
[rollback status, env help msg, and policy added](https://trello.com/c/qmNYe2Nf/1949-rollback-status-env-help-msg-and-policy-added)

PR #1949, opened 2017-02-16
display the output from compile on error

MiguelMoll:


MiguelMoll:
[Display buildpack compile output on error](https://trello.com/c/8IVeaDlN/1959-display-buildpack-compile-output-on-error)

PR #1953, opened 2017-02-21
Add eu-west-2

soulshake:
Tested manually, rack seems to have installed correctly and app deployed successfully.


mattmanning:
[eu-west-2 support](https://trello.com/c/p0dsqrWA/1895-eu-west-2-support)

ci/region.sh
```diff
@@ -43,6 +43,10 @@ case "$1" in
     echo AWS_REGION=ap-southeast-1
     echo RACK_PRIVATE=yes
     ;;
+  9)
+    echo AWS_DEFAULT_REGION=eu-west-2
```
MiguelMoll: Could we group the regions together here. That is make `eu-west-2` number 5 and move the others down. `ap-southeast-2` is floating up there too.

soulshake:
@MiguelMoll is this better?

PR #1958, opened 2017-02-22
allow switching to internal balancers

ddollar:
Change the `LoadBalancerName` when switching to `Internal=Yes` so that CloudFormation can make the change.


ddollar:
[allow switching to internal balancers](https://trello.com/c/ATCU7AqW/1996-allow-switching-to-internal-balancers)

mattmanning:
I've found one case where this doesn't work exactly as expected:

If you have an app that was previously created as Internal with an existing load balancer and then you switch it to Internal=No, redeploying the app doesn't change the load balancer to public as expected.

However, you can work around this by dropping the ports from `docker-compose.yml` deploying, then adding them back and deploying again.

I would be ok with pushing this change through since it fixes many other cases and fixing this quirk on a subsequent PR.

PR #1959, opened 2017-02-22
use /app when calling convox/init

MiguelMoll:
Clean print statements too.


MiguelMoll:
[Switch convox/init to use /app dir](https://trello.com/c/w3nJUB46/1998-switch-convox-init-to-use-app-dir)

PR #1960, opened 2017-02-22
Print a warning when custom values aren't passed to --vpc-cidr or --subnet-cidrs when installing into an existing VPC

soulshake:
Since the CLI fills in a default value for `--vpc-cidr` and `--subnet-cidrs`, I don't think it's possible to tell when a user hasn't passed those flags.

If this PR is accepted, the CLI will display a warning if the user has passed `--existing-vpc` and the value of either `--vpc-cidr` or `--subnet-cidrs` corresponds to the default:

vpc-cidr

### DIFF
--- a/cmd/convox/install.go
+++ b/cmd/convox/install.go
@@ -54,8 +54,10 @@ We recommend that you create a new set of credentials exclusively for this
 install/uninstall process and then delete them once the installer has completed.`
 
 var (
-	distinctID   = ""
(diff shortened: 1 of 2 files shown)


soulshake:
[Require --vpc-cidr for an existing vpc install](https://trello.com/c/fA6zjZsa/1968-require-vpc-cidr-for-an-existing-vpc-install)

PR #1961, opened 2017-02-22
Allow running a single test with 'make test'/'TESTRUN'

soulshake:
```
$ make test TEST=github.com/convox/rack/cmd/convox TESTRUN=TestRequiredFlagsWhenInstallingIntoExistingVPC
env PROVIDER=test CONVOX_WAIT= bin/test
Running single test TestRequiredFlagsWhenInstallingIntoExistingVPC
ok  	github.com/convox/rack/cmd/convox	0.010s	coverage: 1.7% of statements
```

Accepting this PR will enable behavior similar to `go test -test.run` but with `make test`.

### DIFF
--- a/Development.md
+++ b/Development.md
@@ -277,6 +277,12 @@ $ ci/cleanup.sh
 
 This generally uses a specific version number (e.g. 20160323164322) that has been released but not published. Passing integration tests offer confidence that the release can be published.
 
+## Testing
+
+* To run the test suite locally, run `make test`
+* To run a subset of tests, provide `TEST`, e.g. `make test TEST=github.com/convox/rack/cmd/convox`
+* To run a single test, provide `TEST` and `TESTRUN`, e.g. `make test TEST=github.com/convox/rack/cmd/convox TESTRUN=TestRequiredFlagsWhenInstallingIntoExistingVPC`
+
 ## Gotchas
 
 We compile some templates into `.go` files using `go-bindata`. If you make any changes to template files you will often need to run `make templates` to propagate your changes to the `.go` files. If it seems like your template changes aren't doing anything, try running `make templates.
--- a/bin/test
+++ b/bin/test
@@ -7,6 +7,12 @@ go get -t ./...
 rm -f coverage.txt
 
 if [ ! -z ${TEST} ]; then
+    if [ ! -z ${TESTRUN} ]; then
+        echo "Running single test $TESTRUN"
+        go test -coverprofile=profile.out -covermode=atomic "$TEST" -test.run "$TESTRUN"
+        exit
+    fi
+
     go test -coverprofile=profile.out -covermode=atomic "$TEST"
     exit
 fi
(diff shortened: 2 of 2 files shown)


soulshake:
[Allow running a single test](https://trello.com/c/5cnojttG/2001-allow-running-a-single-test)

Development.md
```diff
@@ -277,6 +277,12 @@ $ ci/cleanup.sh
 
 This generally uses a specific version number (e.g. 20160323164322) that has been released but not published. Passing integration tests offer confidence that the release can be published.
 
+## Testing
+
+* To run the test suite locally, run `make test`
+* To run a subset of tests, provide `TEST`, e.g. `make test TEST=github.com/convox/rack/cmd/convox`
+* To run a single test, provide `TEST` and `TESTRUN`, e.g. `make test TEST=github.com/convox/rack/cmd/convox TESTRUN=TestRequiredFlagsWhenInstallingIntoExistingVPC`
```
MiguelMoll: Any thoughts on renaming:
TEST -> PKG (this would indicate its the golang package the test belongs too)
TESTRUN -> TEST (actual test to run)

bin/test
```diff
@@ -7,6 +7,12 @@ go get -t ./...
 rm -f coverage.txt
 
 if [ ! -z ${TEST} ]; then
+    if [ ! -z ${TESTRUN} ]; then
+        echo "Running single test $TESTRUN"
+        go test -coverprofile=profile.out -covermode=atomic "$TEST" -test.run "$TESTRUN"
```
MiguelMoll: I'm not familiar with `-test.run`. I've used `go test pkg -run`. Any known differences?

soulshake: No known differences. I didn't know about `-run`; it looks like `-run` could be enough by itself since it takes `regexp`.

Development.md
soulshake: Works for me; will do

bin/test
soulshake: Update: I couldn't get this to work in 3 minutes, so I just renamed them as requested in your comment above.

bin/test
```diff
@@ -6,8 +6,14 @@ go get -t ./...
...
-if [ ! -z ${TEST} ]; then
-    go test -coverprofile=profile.out -covermode=atomic "$TEST"
+if [ ! -z ${PKG} ]; then
+    if [ ! -z ${TEST} ]; then
+        echo "Running single test $TEST"
+        go test -coverprofile=profile.out -covermode=atomic "$PKG" -test.run "$TEST"
+        exit
+    fi
+
+    go test -coverprofile=profile.out -covermode=atomic "$PKG"
```
ianmalott: 10-16: would this be a little easier to follow as an `if...else`?

ianmalott: Could add an `echo` for this line like the one you have at 11. I think that's a nice touch.

ianmalott:
Another thought: if `TEST` is passed without `PKG`, we could error and print some usage guidance. Don't want to scope creep here, though.

PR #1962, opened 2017-02-22
Use http client.Get instead of http.Get when testing HTTP response codes of docString URLs

soulshake:
Before: 

```
--- FAIL: TestUrls (10.03s)
	Error Trace:	install_test.go:313
	Error:		Received unexpected error "Get https://convox.com/docs/about-resources/: net/http: TLS handshake timeout"
```

After: Hopefully happy tests.


soulshake:
[Increase TLSHandshakeTimeout in tests](https://trello.com/c/l9y7vUMI/2002-increase-tlshandshaketimeout-in-tests)

cmd/convox/install_test.go
```diff
@@ -308,8 +309,14 @@ func TestUrls(t *testing.T) {
 		"http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html",
 	}
 
+	tr := &http.Transport{
+		TLSHandshakeTimeout: 30 * time.Second,
```
MiguelMoll: Thirty seconds seems way to high. Maybe 5? Anything higher than that, something might be up and we should investigate.

soulshake: Fixed

PR #1966, opened 2017-02-23
do not calculate capacity of a draining instance

MiguelMoll:


MiguelMoll:
[Do not calculate draining instance resources](https://trello.com/c/Z8wiWxHc/2007-do-not-calculate-draining-instance-resources)

PR #1963, opened 2017-02-23
Make instance tenancy refer to Tenancy instead of being hardcoded to default

soulshake:
Tested this again (after closing https://github.com/convox/rack/pull/1934) and it seems to be working fine (I was testing with a local Rack instead of uploading rack.json to CF web interface; maybe that was the problem?).

Tested manually with:

### Standard

```
Ami	
ApiCpu	128
ApiMemory	128
Autoscale	Yes
BuildCpu	0
BuildImage	
BuildInstance	
BuildMemory	1024
ClientId	dev@convox.com
ContainerDisk	10
* CustomTopicRuntime	nodejs4.3
Development	No
Encryption	Yes
ExistingVpc	
InstanceBootCommand	
InstanceCount	3
InstanceRunCommand	
* InstanceType	m4.large
InstanceUpdateBatchSize	1
Internal	No
InternetGateway	
Key	
* Password	****
Private	No
PrivateApi	No
Subnet0CIDR	10.0.1.0/24
Subnet1CIDR	10.0.2.0/24
Subnet2CIDR	10.0.3.0/24
SubnetPrivate0CIDR	10.0.4.0/24
SubnetPrivate1CIDR	10.0.5.0/24
SubnetPrivate2CIDR	10.0.6.0/24
SwapSize	5
* Tenancy	dedicated
* Version	20170216230221
VolumeSize	50
VPCCIDR	10.0.0.0/16
```

### Internal

```
Ami	
ApiCpu	128
ApiMemory	128
Autoscale	Yes
BuildCpu	0
BuildImage	
BuildInstance	
BuildMemory	1024
ClientId	dev@convox.com
ContainerDisk	10
* CustomTopicRuntime	nodejs4.3
Development	No
Encryption	Yes
ExistingVpc	
InstanceBootCommand	
InstanceCount	3
InstanceRunCommand	
* InstanceType	m4.large
InstanceUpdateBatchSize	1
* Internal	Yes
InternetGateway	
Key	
* Password	****
Private	No
PrivateApi	No
Subnet0CIDR	10.0.1.0/24
Subnet1CIDR	10.0.2.0/24
Subnet2CIDR	10.0.3.0/24
SubnetPrivate0CIDR	10.0.4.0/24
SubnetPrivate1CIDR	10.0.5.0/24
SubnetPrivate2CIDR	10.0.6.0/24
SwapSize	5
* Tenancy	dedicated
* Version	20170216230221
VolumeSize	50
VPCCIDR	10.0.0.0/16
```

[...]

PR #1970, opened 2017-02-23
Add dedicated iam permissions

soulshake:


MiguelMoll:
Since we're breaking the permissions into a separate PR, this one **MUST** go out before #1963. And this PR shouldn't include the `"InstanceTenancy": { "Ref": "Tenancy" },` as that one belongs in #1963.

PR #1963 should also change it's base branch to use `add-dedicated-iam-permissions` instead of master.

Trying to document as much of this as possible to get a process around IAM permissions.

soulshake:
[Add dedicated IAM permissions](https://trello.com/c/SuYPstAa/2018-add-dedicated-iam-permissions)

MiguelMoll:
Are these permissions needed for install too? If so, we'd have to create a PR for console as well updating the [integration](https://github.com/convox/console/blob/master/models/aws_provider.go#L2455) permissions.

MiguelMoll:
Should have checked before making a comment, but Console already has these permissions for install so thats good.

PR #1977, opened 2017-02-24
No updates user errors

### LINKED ISSUE #1935: ERROR: No updates are to be performed
An error message is displayed on safe, non-operations like promoting the active release or setting a parameter to its current value.

Let's make this error more friendly. The error message should point to a next action. And these non-operations should not trigger a release started, failed or error webhook.

## Desired behavior

```
$ convox releases promote RFHEGMEIIKJ --app python-getting-started
ERROR: Release RFHEGMEIIKJ is already active. Try `convox releases`.

$ convox apps params set DatabaseFormation=1,0,256
ERROR: DatabaseFormation is unchanged. Try `convox apps params`.
```

## Old

```
$ convox releases promote RFHEGMEIIKJ --app python-getting-started
Promoting RFHEGMEIIKJ... UPDATING

$ convox releases promote RFHEGMEIIKJ --app python-getting-started
Promoting RFHEGMEIIKJ... ERROR: No updates are to be performed.
```

```
$ convox apps params | grep DatabaseFormation
DatabaseFormation            1,0,256

$ convox apps params set DatabaseFormation=1,0,256 --app python-getting-started
Updating parameters... ERROR: ValidationError: No updates are to be performed.
    status code: 400, request id: 77f4e5d3-f2fb-11e6-ae71-f3d4c2e66e79
```

awsmsrc:
If an action results in no updates the user should see a helpful message

fixes #1935

### DIFF
--- a/cmd/convox/apps.go
+++ b/cmd/convox/apps.go
@@ -267,6 +267,27 @@ func cmdAppParamsSet(c *cli.Context) error {
 		params[parts[0]] = parts[1]
 	}
 
+	existing, err := rackClient(c).ListParameters(app)
+	if err != nil {
+		return stdcli.Error(err)
+	}
+
+	hasChange := false
+	for k, v := range params {
+		res, ok := existing[k]
+		if !ok {
+			hasChange = true
+			break
+		} else if res != v {
+			hasChange = true
+			break
+		}
+	}
+
+	if !hasChange {
+		return stdcli.Error(fmt.Errorf("No updates are to be performed"))
+	}
+
 	stdcli.Startf("Updating parameters")
 
 	err = rackClient(c).SetParameters(app, params)
--- a/cmd/convox/apps.go
+++ b/cmd/convox/apps.go
@@ -267,31 +267,13 @@ func cmdAppParamsSet(c *cli.Context) error {
 		params[parts[0]] = parts[1]
 	}
 
-	existing, err := rackClient(c).ListParameters(app)
-	if err != nil {
-		return stdcli.Error(err)
-	}
-
-	hasChange := false
-	for k, v := range params {
-		res, ok := existing[k]
-		if !ok {
-			hasChange = true
-			break
-		} else if res != v {
-			hasChange = true
-			break
-		}
-	}
-
-	if !hasChange {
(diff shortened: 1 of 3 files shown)


awsmsrc:
[user friendly cli errors for non actionable stack requests](https://trello.com/c/vLxFRGTT/2020-user-friendly-cli-errors-for-non-actionable-stack-requests)

MiguelMoll:
Any way to add simple tests?

cmd/convox/apps.go
```diff
@@ -267,6 +267,27 @@ func cmdAppParamsSet(c *cli.Context) error {
...
+			hasChange = true
+			break
+		} else if res != v {
+			hasChange = true
+			break
+		}
+	}
+
+	if !hasChange {
+		return stdcli.Error(fmt.Errorf("No updates are to be performed"))
```
MiguelMoll: Could we direct the user to `convox apps params` to see existing values? Same with the rack params.

cmd/convox/rack.go
```diff
@@ -205,6 +205,27 @@ func cmdRackParamsSet(c *cli.Context) error {
...
+		if !ok {
+			hasChange = true
+			break
+		} else if res != v {
+			hasChange = true
+			break
+		}
+	}
+
+	if !hasChange {
```
ddollar: A parameter change is not the only thing that can cause a stack to change. Sometimes we update the underlying templates and the user will get changes even without parameter changes.

Let's not try to reimplement CloudFormation here. Are we just looking for a nicer display of their error?

awsmsrc: This is specifically in the `rack params set` command so I thought that would be the only way to update the stack here?

The goal I was assigned was to validate the request and if it wouldn't make a change don't make a request to the rack and trigger failure requests

ddollar: AWS will tell us if it won't make a change. Let's look for that error message and leave the logic here up to the CF engine.

MiguelMoll:
Original intention was to clean up rack metrics around updates. Keeping the PR to just showing user friend error messages.

MiguelMoll:
Went out via https://github.com/convox/rack/pull/2006

PR #1982, opened 2017-02-27
configure login transport to allow for proxies

awsmsrc:
The implementation of retrieving the console user id in the login function broke the login command for users using proxies.

This should fix that.

### DIFF
--- a/client/auth.go
+++ b/client/auth.go
@@ -7,6 +7,9 @@ import (
 	"net/http"
 )
 
+//Auth is a request that simply checks whether the password is valid
+//in the case of console the users id will be returned, a rack will
+//return an empty string
 func (c *Client) Auth() (string, error) {
 	req, err := http.NewRequest("GET", fmt.Sprintf("https://%s/auth", c.Host), nil)
 	if err != nil {
@@ -35,7 +38,8 @@ func (c *Client) Auth() (string, error) {
 
 	err = json.Unmarshal(body, &data)
 	if err == nil {
-		//if bad JSON is returned it is probably a legacy app
+		//if bad JSON is returned it is probably a legacy rack
+		//which used to return the plain text string 'OK'
 		id = data["id"]
 	}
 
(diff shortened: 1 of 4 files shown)


MiguelMoll:
Could we change the controller to return JSON instead of just OK? What would break? Out of scope?

awsmsrc:
it shouldn't really break anything changing the controller to return json, as people won't be testing against that string, but login would still break with older racks

client/auth.go
```diff
@@ -1,29 +1,43 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	var id string
+	var data map[string]string
+
+	err = json.Unmarshal(body, &data)
+	if err == nil {
+		//if bad JSON is returned it is probably a legacy app
```
ddollar: What does legacy app mean in this context?

Should probably document the "", nil potential return case here as it's out of the value-or-error normal.

client/auth.go
```diff
@@ -1,29 +1,43 @@
...
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 )
 
-func (c *Client) Auth() error {
+func (c *Client) Auth() (string, error) {
```
ddollar: Does /auth return more than just id in that JSON hash? If so this function should be returning a struct.

client/auth.go
ddollar: Does this "legacy app" situation happen enough to justify swallowing errors blindly here?

awsmsrc: questions 1 + 3) app is meant to be rack, i.e anything before this release, will update the comment

all racks will return "", nil - that is the current released behaviour (user ids are a console concept), i will add a comment

ddollar:
[convox login uses proxies so requires client code but still needs console id](https://trello.com/c/TqGSH7U2/2052-convox-login-uses-proxies-so-requires-client-code-but-still-needs-console-id)

MiguelMoll:
https://github.com/convox/rack/pull/2006

PR #1983, opened 2017-02-27
check if instance to drain has any tasks

MiguelMoll:


MiguelMoll:
[QA convox init and buildpacks](https://trello.com/c/pgBddwun/1964-qa-convox-init-and-buildpacks)

PR #1988, opened 2017-03-01
Add more defaults for a rails app

### LINKED ISSUE #1985: `convox init` for Rails remaining issues
### Address binding in development

When I run my app in development mode (RAILS_ENV=development RACK_ENV=development), the ruby server (puma in this case) binds to localhost rather than 0.0.0.0.

I see 3 potential solutions for this:

1. Edit the generated `command` in docker-compose.yml to add `-b 0.0.0.0`.
2. Inject the environment variable SERVER_NAME=0.0.0.0
3. Inject the environment variable HTTP_HOST=0.0.0.0

1 seems like the best option that would work everywhere. 2 and 3 are puma-specific. It still bugs me that I can't tell why this behavior is RAILS_ENV specific, but these solutions work in development and production.

### Build arg for Bundler

Users need to be able to specify which gem groups to exclude when bundling gems this can be done by adding:

`ARG BUNDLE_WITHOUT="development:test"`

to the Dockerfile. `development:test` matches the Heroku ruby buildpack's default value. Making it a build arg means users can override it locally in their `.env` and in production using `convox env set` if desired.

### .env in .gitignore

Can we inject .env into the .gitignore if .gitignore exists and doesn't already have .env?

### Installing new gems

This is more of an open question for me right now. How do we want users to install new gems when developing. We previously suggested

1. Edit the Gemfile
2. Run `convox start` which will install the gems and generate changes to `Gemfile.lock`
3. Code-sync will sync the updated `Gemfile.lock` back to the host

With the Heroku buildpack this results in:

```
 The command '/bin/sh -c /tmp/heroku-buildpack-ruby/bin/compile /app /tmp/cache' returned a non-zero code: 1
ERROR: build error: exit status 1
```

I'm pretty sure the buildpack wants Gemfile and Gemfile.lock to already match. Do we tell users they have to run `bundle install` on the host to generate the correct Gemfile.lock, or is there some way to make it happen in the dev container?

MiguelMoll:
Fixes #1985 (See for background)

- Init will edit the generated command in docker-compose.yml to add -b 0.0.0.0
- `ARG BUNDLE_WITHOUT="development:test"` added to Dockerfile for all rails apps
- .env in .gitignore if .gitignore does not exist (this is keeping with not touching a file if it exists. thoughts?)
- I added a shell trick to display the error if there is one while running the buildpack
    - Verified by writing `gen 'pg'`

The Gemfile issue is still up for discussion. Maybe this can go out for now? I tried running `convox start web bundle update` after editing the Gemfile on my host. I didn't see the error mentioned. I saw:

```
build  │ -----> Compiling Ruby/Rails -----> Using Ruby version: ruby-2.2.6 -----> Installing dependencies using bundler 1.13.7 Running: bundle install --without none --path vendor/bundle --binstubs vendor/bundle/bin -j4 --deployment Don't run Bundler as root. Bundler can ask for sudo if it is needed, and installing your bundle as root will break this application for all non-root users on this machine. You are trying to install in deployment mode after changing your Gemfile. Run `bundle install` elsewhere and add the updated Gemfile.lockto version control. You have added to the Gemfile: Dockerfile Gemfile Gemfile.lock README.md Rakefile app bin config config.ru db docker-compose.yml lib log public test vendor pg Bundler Output: Don't run Bundler as root. Bundler can ask for sudo if it is needed, and installing your bundle as root will break this application for all non-root users on this machine. You are trying to install in deployment mode after changing your Gemfile. Run `bundle install` elsewhere and add the updated Gemfile.lock to version control. You have added tothe Gemfile: Dockerfile Gemfile Gemfile.lock README.md Rakefile app bin config config.ru db docker-compose.yml lib log public test vendor pg ! ! Failed to install gems via Bundler.!
build  │  ---> 5d6f37dcc79
```

and

 ```
web    │ You are trying to install in deployment mode after changing
web    │ your Gemfile. Run `bundle install` elsewhere and add the
web    │ updated Gemfile.lock to version control.
web    │
web    │ If this is a development machine, remove the /app/Gemfile freeze
web    │ by running `bundle install --no-deployment`.
web    │
web    │ You have added to the Gemfile:
web    │ * source: rubygems repository https://rubygems.org/
web    │ * rails (~> 5.0.1)
web    │ * sqlite3
web    │ * pg
...
```


MiguelMoll:
[QA convox init and buildpacks](https://trello.com/c/pgBddwun/1964-qa-convox-init-and-buildpacks)

MiguelMoll:
[Cloudformation events in Rack and App Logs](https://trello.com/c/EB80X3Io/2036-cloudformation-events-in-rack-and-app-logs)

PR #1992, opened 2017-03-01
Avoid AWS rate limits for apps with lots of processes

### LINKED ISSUE #1990: AWS rate limit errors when running `convox ps` on an app with many processes
I have an app running 300 containers. When I run `convox ps` on it I get the following error:

```
ERROR: RequestLimitExceeded: Request limit exceeded.
	status code: 503, request id: 11113443-bb1e-4021-a86c-75e71c7ec8d3
```

This is likely related to pagination of API calls. Previously `convox ps` could only fetch 100 processes max, so it probably couldn't make enough API calls to get rate-limited.

mattmanning:
Add a short delay between pages to prevent AWS rate limiting.

Fixes #1990

### DIFF
--- a/provider/aws/processes.go
+++ b/provider/aws/processes.go
@@ -230,6 +230,7 @@ func (p *AWSProvider) stackTasks(stack string) ([]string, error) {
 				for _, arn := range page.TaskArns {
 					tasks = append(tasks, *arn)
 				}
+				time.Sleep(100 * time.Millisecond)
 				return true
 			},
 		)


mattmanning:
[Avoid AWS rate limits on `convox ps` for apps with lots of processes](https://trello.com/c/okPBUlXZ/2046-avoid-aws-rate-limits-on-convox-ps-for-apps-with-lots-of-processes)

mattmanning:
I'm pulling this one. Doesn't actually work in later testing. Looks like we'll need to do some sort of retry + backoff strategy instead of a simple delay.

PR #1993, opened 2017-03-01
elastic search 

slajax:
Overview
===

I really wanted to be able to create managed elastic search resources via the convox cli so I added it. 

It is working right now using a local rack. It set it up so can white list a local IP for `es:*` (super admin) and define a user ARN so that you can sign your requests with it's credentials for `es:ESHttp*` (search, insert etc). 

By default your ECS containers won't even have access to the elastic search resource. You have to sign requests using the ARN key and secret using Signature Version 4 signing as described [here](https://aws.amazon.com/blogs/security/how-to-control-access-to-your-amazon-elasticsearch-service-domain/). I've tested it manually by exec'ing into a container and this is working.

I also haven't added any tests yet, but I'm going to look into that next - I may need some help on that - this is the first time I have ever worked with go and I'm learning as I go _(see what I did there?)_.

Example
===

`❯ convox resources create elasticsearch --domain-name testing --user-arn "arn:aws:iam::123456789101:user/es" --whitelist-ip "8.8.8.8"`

You can also over write the other defaults:

`elasticsearch  [--domain-name] [--user-arn] [--whitelist-ip=127.0.0.1/0] [--version=2.3] [--dedicated-master=true] [--master-instance-type=m3.medium.elasticsearch] [--master-instance-count=3] [--instance-type=m3.medium.elasticsearch] [--instance-count=2] [--iops=0] [--volume-size=20]`

Tasks
===
- [x] create working cloudformation template
- [x] sync variables with cli
- [x] use sane defaults
- [x] default whitelist to 127.0.0.1/0 
- [x] allow overrides
- [x] add to correct places in cli logic
- [ ] links
- [ ] allow fluentd to forward logs to es
- [ ] tests - *handing off to convox team*


slajax:
@ianmalott please ping me in slack if you get a moment tomorrow.

provider/aws/templates/resource/elasticsearch.tmpl
```diff
@@ -0,0 +1,128 @@
...
+    "Conditions": {
+    },
+    "Parameters": {
+      "DomainName": {
+        "Type": "String",
+        "Description": "Name of elasticsearch domain"
+      },
+      "UserArn": {
+        "Type": "String",
+        "Description": "Name of elasticsearch domain"
```
ddollar: Looks like a description accidentally got copy/pasted here

provider/aws/templates/resource/elasticsearch.tmpl
```diff
@@ -0,0 +1,128 @@
+{{ define "resource" }}
+{
+  "AWSTemplateFormatVersion" : "2010-09-09",
+    "Conditions": {
+    },
+    "Parameters": {
```
ddollar: Would you please alphabetize these parameters? Makes it easier to manage as things grow.

cmd/convox/services.go
```diff
@@ -37,6 +37,10 @@ func init() {
 			"[--automatic-failover-enabled] [--database=db-name] [--instance-type=cache.t2.micro] [--num-cache-clusters=1] [--private]",
 		},
 		{
+			"elasticsearch",
+			"[--domain-name] [--user-arn] [--whitelist-ip=127.0.0.1/0] [--version=2.3] [--dedicated-master=true] [--master-instance-type=m3.medium.elasticsearch] [--master-instance-count=3] [--instance-type=m3.medium.elasticsearch] [--instance-count=2] [--iops=0] [--volume-size=20]",
```
ddollar: Should be `[--domain-name=example-domain-here]` etc for parameters that take values.

slajax:
@ddollar here are the changes you requested. I tried creating a new es resource while there is an existing one in a "deleting" state and it looks like I'm getting this: 

`ERROR: ValidationError: Circular dependency between resources: [ElasticsearchDomain]`

I'm going to test again once it's deleted to confirm, but this seems flawed. I should be able to create multiple ES Domains at a given time no?

slajax:
Hmm, nope, looks like it's consistent now. Even if I roll back to the previous commit that was working last night. I'm kind of blocked here I think. Let me know if you have any ideas.

slajax:
This is working well now, it's just missing the tasks as described in issue. The rest of it is a bit over my head, but let me know if I can be of any help with this.

ddollar:
Closing for now since this is quite old but will borrow this cloudformation template for Praxis. Thanks!

Issue #1860, opened 2017-02-08
CLI description, argument, help consistency and cleanup


ddollar:
data: {u'text': u'https://github.com/convox/rack/issues/1779', u'list': {u'name': u'Rack', u'id': u'5899e3224255b666d6e7d9e9'}, u'board': {u'shortLink': u'XxvWeICx', u'id': u'587425638297353d93b2d4af', u'name': u'Roadmap'}, u'card': {u'idShort': 26, u'shortLink': u'efteFk3N', u'id': u'5874589bb341d597857671a0', u'name': u'CLI description, argument, help consistency and cleanup'}}
date: 2017-02-08T13:15:21.057Z
id: 589b19e9a027db69e2ceb965
idMemberCreator: 4f3c59c1d9d3e56b446861d8
memberCreator: {u'username': u'ddollar', u'fullName': u'David Dollar', u'avatarHash': u'4b941446da3c56990164d470ea22bb2f', u'id': u'4f3c59c1d9d3e56b446861d8', u'initials': u'DD'}
type: commentCard

https://trello.com/c/efteFk3N/26-cli-description-argument-help-consistency-and-cleanup

MiguelMoll:
Fixed via #1996

PR #1996, opened 2017-03-01
Check for help on every command

soulshake:
Before:

```
$ convox ps -h
ERROR: no such app: rack
```

After:

```
$ convox ps -h
convox ps: list an app's processes

Usage:
  convox ps <command> [args...]

Subcommands: (convox ps help <subcommand>)
  info		show info for a process (note: you can obtain the process id via 'convox ps')
  stop		stop a process by its id
  help, h	
  
Options:
  --app value, -a value	app name inferred from current directory if not specified
  --rack value		rack name
  --stats		display process cpu/memory stats
  --help, -h		show help
```

Currently, treatment of `-h` and `--help` is very inconsistent. It's not as simple as adding a global --rack flag (as in [this PR](https://github.com/convox/rack/pull/1684/files#diff-641894c246f9a42ae17552d302cdb082)) because the golang flag library panics if `-h` gets redefined.

This PR attempts to unify handling of the help flag as well as the  number of required arguments by adding two methods to `stdcli`: `NeedHelp` and `NeedArg`.

Addresses:

- #2018
- #1860

### DIFF
--- a/cmd/convox/apps.go
+++ b/cmd/convox/apps.go
@@ -20,14 +20,16 @@ func init() {
 			{
 				Name:        "cancel",
 				Description: "cancel an update",
-				Usage:       "",
+				Usage:       "[options]",
+				ArgsUsage:   "",
 				Action:      cmdAppCancel,
 				Flags:       []cli.Flag{appFlag, rackFlag},
 			},
 			{
 				Name:        "create",
 				Description: "create a new application",
-				Usage:       "<name>",
+				Usage:       "[name] [options]",
+				ArgsUsage:   "[name] (inferred from current directory if not specified)",
 				Action:      cmdAppCreate,
 				Flags: []cli.Flag{
 					rackFlag,
@@ -56,13 +58,15 @@ func init() {
 				Name:        "params",
 				Description: "list advanced parameters for an app",
 				Usage:       "[name]",
+				ArgsUsage:   "",
 				Action:      cmdAppParams,
 				Flags:       []cli.Flag{appFlag, rackFlag},
 				Subcommands: []cli.Command{
 					{
 						Name:        "set",
 						Description: "update advanced parameters for an app",
-						Usage:       "NAME=VALUE [NAME=VALUE]",
+						Usage:       "NAME=VALUE [NAME=VALUE] ... [options]",
+						ArgsUsage:   "NAME=VALUE",
 						Action:      cmdAppParamsSet,
 						Flags:       []cli.Flag{appFlag, rackFlag},
 					},
@@ -73,14 +77,8 @@ func init() {
 }
 
 func cmdApps(c *cli.Context) error {
-	if len(c.Args()) > 0 {
-		return stdcli.Error(fmt.Errorf("`convox apps` does not take arguments. Perhaps you meant `convox apps create`?"))
-	}
-
-	if c.Bool("help") {
-		stdcli.Usage(c, "")
-		return nil
-	}
+	stdcli.NeedHelp(c)
+	stdcli.NeedArg(c, 0)
 
 	apps, err := rackClient(c).GetApps()
 	if err != nil {
@@ -98,6 +96,9 @@ func cmdApps(c *cli.Context) error {
 }
 
 func cmdAppCancel(c *cli.Context) error {
+	stdcli.NeedHelp(c)
+	stdcli.NeedArg(c, 0)
+
 	_, app, err := stdcli.DirApp(c, ".")
 	if err != nil {
 		return stdcli.Error(err)
@@ -119,12 +120,16 @@ func cmdAppCancel(c *cli.Context) error {
 }
 
 func cmdAppCreate(c *cli.Context) error {
+	stdcli.NeedHelp(c)
+
 	_, app, err := stdcli.DirApp(c, ".")
 	if err != nil {
 		return stdcli.Error(err)
 	}
 
 	if len(c.Args()) > 0 {
+		// accept no more than 1 argument
+		stdcli.NeedArg(c, 1)
 		app = c.Args()[0]
 	}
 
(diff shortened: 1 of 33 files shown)


nzoschke:
`convox`                      -> app template
`convox help`              -> app template
`convox ps`                 -> command template
`convox ps help`         -> command template
`convox ps stop`         -> subcommand template
`convox ps stop help` -> subcommand template

nzoschke:
I like the direction of this. I can see a comprehensive CLI output testing drive:

* Better user experience
* Consistency / polish of the CLI
* Safer and faster changes to CLI commands
* Safer and faster changes to CLI stdout/stderr formatting

The various `help` commands is a nice way to establish coverage since its very user-centric (getting missing, bad or confusing help is not helpful) and doesn't require any test server stuff.

cmd/convox/apps.go
```diff
@@ -73,14 +77,8 @@ func init() {
...
-	if len(c.Args()) > 0 {
-		return stdcli.Error(fmt.Errorf("`convox apps` does not take arguments. Perhaps you meant `convox apps create`?"))
-	}
-
-	if c.Bool("help") {
-		stdcli.Usage(c, "")
-		return nil
-	}
+	stdcli.NeedHelp(c)
+	stdcli.NeedArg(c, 0)
```
nzoschke: It's unfortunate we need these in every handler. It seems like the CLI framework should be able to handle this in its command parsing / dispatching stuff.

But what I do like about this PR is that more comprehensive around the expected CLI output makes the output expectations explicit, which will allow confident changes to the CLI framework now or in the future.

cmd/convox/stdcli/stdcli.go
```diff
@@ -309,3 +306,51 @@ func ParseOpts(args []string) map[string]string {
...
+
+	if len(c.Args()) < count || count < 0 {
+		Errorf("missing required %s %s",
+			expected,
+			c.Command.ArgsUsage)
+		Usage(c, c.Command.Name)
+	}
+
+	if len(c.Args()) > count {
+		msg := fmt.Sprintf("expected %d %s %s; got '%s'. Perhaps you meant to use a subcommand or option?",
```
nzoschke: My first impression of this output is that it's redundant, because the CLI usage that's printed out should imply the problem.

```
$ git init foo bar
usage: git init [-q | --quiet] [--bare] [--template=<template-directory>] [--shared[=<permissions>]] [<directory>]
```

I also like it when we the CLI can hint at something more specific:

```
$ heroku app info
 !    `app` is not a heroku command.
 !    Perhaps you meant `api` or `apps`.
 !    See `heroku help` for a list of available commands.
```

However I do like:

* The consistency
* Tests that assert the stderr output

cmd/convox/apps.go
soulshake: I tested with another approach using the `Before` field in the `cli.Command` struct, but it has some annoying issues that I can't for the life of me remember right now.

cmd/convox/stdcli/stdcli.go
soulshake: I see where you're coming from, but I think it's better to be specific if we possibly can.

soulshake:
[`convox ps` -h, --help not respected](https://trello.com/c/dDYkVzYb/1944-convox-ps-h-help-not-respected)

cmd/convox/ps.go
```diff
@@ -27,15 +28,19 @@ func init() {
 		Subcommands: []cli.Command{
 			{
 				Name:        "info",
-				Description: "show info for a process",
-				Usage:       "<id>",
+				Description: "show info for a process (note: you can obtain the process id via 'convox ps')",
```
nzoschke: style: I don't think we need the note

cmd/convox/ps.go
```diff
@@ -14,8 +14,9 @@ func init() {
 	stdcli.RegisterCommand(cli.Command{
 		Name:        "ps",
 		Description: "list an app's processes",
-		Usage:       "",
 		Action:      cmdPs,
+		Usage:       "[subcommand] [args] [options]",
```
nzoschke: I see

```
$ convox ps --help
convox ps: list an app's processes

Usage:
  convox ps <command> [subcommand]
```

I'm not sure what the `<command>` means here. I think it should just be `convox ps [subcommand]`

cmd/convox/ps.go
```diff
@@ -44,6 +49,9 @@ func init() {
 }
 
 func cmdPs(c *cli.Context) error {
+	stdcli.NeedHelp(c)
+	stdcli.NeedArg(c, 0)
```
nzoschke: The expected args logic makes sense but I think we should not display the red `ERROR` warning. It seems unnecessary with good Usage output and can be a bit confusing on its own.

```
$ convox ps web --app site-staging
ERROR: expected 0 arguments; got 3 arguments (web --app site-staging). Perhaps you meant to use a subcommand or option?
convox ps: list an app's processes
```

soulshake: Pushed an update to only display the error if `DEBUG` or `CONVOX_DEBUG` are non-empty. Would that work?

cmd/convox/ps.go
soulshake: Fixed

cmd/convox/ps.go
nzoschke: That works, however I'm still seeing the behavior

```
$ convox releases --app myapp thunk
ERROR: expected 0 arguments; got 1 argument (ass). Perhaps you meant to use a subcommand or option?
```

nzoschke:
It's looking good to me. One more nitpick...

```
$ convox -h
convox: command-line application management

Usage:
  convox <command> [subcommand] [options...] [args...]

Subcommands: (convox help <subcommand>)
  api                  make a rest api call to a convox endpoint
  ...
```

I think the usage string should say "Commands: (convox help <command>)" if the main usage string is `convox <command> [subcommand]`

nzoschke:
I think CONVOX_DEBUG is reversed:

```
$ CONVOX_DEBUG=true convox releases --app myapp thunk
convox releases: list an app's releases

$ convox releases --app myapp thunk
ERROR: expected 0 arguments; got 1 argument (thunk). Perhaps you meant to use a subcommand or option?
convox releases: list an app's releases
```

cmd/convox/ps.go
soulshake: Fixed

nzoschke:
Latest changes LGTM

PR #1998, opened 2017-03-01
encrypt ebs volumes

ddollar:


MiguelMoll:
Code is simple enough. Could we describe in the PR what this means and what users can expect? ssh into an instance works as usual?

ddollar:
Should be completely invisible other than the EBS volumes will now be encrypted at rest.

PR #1930, opened 2017-02-13
Add CF Resource metadata

ianmalott:
Add Description, Name, or Tags to CF Resources where possible.

Before this is released, https://github.com/convox/rack/pull/2005 should be released in a required release and https://github.com/convox/console/pull/927 should be deployed to Console.

PR #2005, opened 2017-03-03
IAM permissions needed for tags and alias update

mattmanning:
Permissions to support https://github.com/convox/rack/pull/1930

Should go out in a required release before https://github.com/convox/rack/pull/1930 is released.


mattmanning:
[Permissions for Tags PR](https://trello.com/c/6zgUSwXz/2065-permissions-for-tags-pr)

PR #2034, opened 2017-03-08
Don't panic when user runs 'convox install/uninstall' with no AWS credentials in env or config

soulshake:
Before:

```
$ cx uninstall one two


     ___    ___     ___   __  __    ___   __  _
    / ___\ / __ \ /  _  \/\ \/\ \  / __ \/\ \/ \
   /\ \__//\ \_\ \/\ \/\ \ \ \_/ |/\ \_\ \/>  </
   \ \____\ \____/\ \_\ \_\ \___/ \ \____//\_/\_\
    \/____/\/___/  \/_/\/_/\/__/   \/___/ \//\/_/


Using AWS CLI for authentication...
You appear to have the AWS CLI installed but have not configured credentials.
You can configure credentials by running `aws configure`.

This installer needs AWS credentials to install/uninstall the Convox platform into
your AWS account. These credentials will only be used to communicate between this
installer running on your computer and the AWS API.

We recommend that you create a new set of credentials exclusively for this
install/uninstall process and then delete them once the installer has completed.
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xd15da6]

goroutine 1 [running]:
main.readCredentials(0x0, 0x0, 0x1, 0xfa, 0x0)
	/home/aj/.go/src/github.com/convox/rack/cmd/convox/install.go:772 +0x2d6
main.cmdUninstall(0xc4201817c0, 0x0, 0xc4201817c0)
	/home/aj/.go/src/github.com/convox/rack/cmd/convox/uninstall.go:80 +0x249
github.com/convox/rack/vendor/gopkg.in/urfave/cli%2ev1.HandleAction(0xe322c0, 0x1097d78, 0xc4201817c0, 0xc42001fb00, 0x0)
	/home/aj/.go/src/github.com/convox/rack/vendor/gopkg.in/urfave/cli.v1/app.go:484 +0xd4
github.com/convox/rack/vendor/gopkg.in/urfave/cli%2ev1.Command.Run(0x1063010, 0x9, 0x0, 0x0, 0x0, 0x0, 0x0, 0x108845f, 0x27, 0x0, ...)
	/home/aj/.go/src/github.com/convox/rack/vendor/gopkg.in/urfave/cli.v1/command.go:207 +0xb72
github.com/convox/rack/vendor/gopkg.in/urfave/cli%2ev1.(*App).Run(0xc4201e2820, 0xc420010600, 0x4, 0x4, 0x0, 0x0)
	/home/aj/.go/src/github.com/convox/rack/vendor/gopkg.in/urfave/cli.v1/app.go:249 +0x7d0
main.main()
	/home/aj/.go/src/github.com/convox/rack/cmd/convox/main.go:70 +0x17b
```

After:

```
$ cx uninstall one two


     ___    ___     ___   __  __    ___   __  _
    / ___\ / __ \ /  _  \/\ \/\ \  / __ \/\ \/ \
   /\ \__//\ \_\ \/\ \/\ \ \ \_/ |/\ \_\ \/>  </
   \ \____\ \____/\ \_\ \_\ \___/ \ \____//\_/\_\
    \/____/\/___/  \/_/\/_/\/__/   \/___/ \//\/_/


Using AWS CLI for authentication...
You appear to have the AWS CLI installed but have not configured credentials.
You can configure credentials by running `aws configure`.

This installer needs AWS credentials to install/uninstall the Convox platform into
your AWS account. These credentials will only be used to communicate between this
installer running on your computer and the AWS API.

We recommend that you create a new set of credentials exclusively for this
install/uninstall process and then delete them once the installer has completed.
AWS Access Key ID: ^C
```

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -6,6 +6,9 @@ go get -t ./...
 
 rm -f coverage.txt
 
+# Print urfave/cli template parsing errors
+export CLI_TEMPLATE_ERROR_DEBUG=true
+
 if [ ! -z ${PKG} ]; then
     if [ ! -z ${TEST} ]; then
         echo "Running single test $TEST"
--- a/cmd/convox/install_test.go
+++ b/cmd/convox/install_test.go
@@ -17,11 +17,11 @@ import (
 )
 
 func TestConvoxInstallSTDINCredentials(t *testing.T) {
-	stackId := "arn:aws:cloudformation:us-east-1:123456789:stack/MyStack/aaf549a0-a413-11df-adb3-5081b3858e83"
+	stackID := "arn:aws:cloudformation:us-east-1:123456789:stack/MyStack/aaf549a0-a413-11df-adb3-5081b3858e83"
 	cycles := []awsutil.Cycle{
 		{
 			awsutil.Request{"GET", "/", "", "/./"},
-			awsutil.Response{200, `<CreateStackResult><StackId>` + stackId + `</StackId></CreateStackResult>`},
+			awsutil.Response{200, `<CreateStackResult><StackId>` + stackID + `</StackId></CreateStackResult>`},
 		},
 		{
 			awsutil.Request{"GET", "/", "", ""},
@@ -42,7 +42,7 @@ func TestConvoxInstallSTDINCredentials(t *testing.T) {
 			Exit:    0,
 			Env:     map[string]string{"AWS_ENDPOINT_URL": s.URL, "AWS_REGION": "test"},
 			Stdin:   `{"Credentials":{"AccessKeyId":"FOO","SecretAccessKey":"test","Expiration":"2015-09-17T14:09:41Z"}}`,
-			Stdout:  Banner + "\nInstalling Convox (" + latest + ")...\nUsing AWS Access Key ID: FOO\n" + stackId + "\n",
+			Stdout:  Banner + "\nInstalling Convox (" + latest + ")...\nUsing AWS Access Key ID: FOO\n" + stackID + "\n",
 		},
 	)
 }
@@ -154,11 +154,11 @@ func TestConvoxInstallFileCredentialsWithEnvCredentials(t *testing.T) {
 }
 
 func TestConvoxInstallValidateStackName(t *testing.T) {
-	stackId := "arn:aws:cloudformation:us-east-1:123456789:stack/MyStack/aaf549a0-a413-11df-adb3-5081b3858e83"
+	stackID := "arn:aws:cloudformation:us-east-1:123456789:stack/MyStack/aaf549a0-a413-11df-adb3-5081b3858e83"
 	cycles := []awsutil.Cycle{
 		{
 			awsutil.Request{"GET", "/", "", "/./"},
-			awsutil.Response{200, `<CreateStackResult><StackId>` + stackId + `</StackId></CreateStackResult>`},
+			awsutil.Response{200, `<CreateStackResult><StackId>` + stackID + `</StackId></CreateStackResult>`},
 		},
 		{
 			awsutil.Request{"GET", "/", "", ""},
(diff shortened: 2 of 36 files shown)


soulshake:
[Don't panic when user runs 'convox install' with no AWS credentials in env or config](https://trello.com/c/awkDU0fW/2083-don-t-panic-when-user-runs-convox-install-with-no-aws-credentials-in-env-or-config)

cmd/convox/uninstall_test.go
```diff
@@ -0,0 +1,82 @@
...
+   
+`
+
+var errUninstallNeedArg = `ERROR: at least 2 arguments are required: <stack-name> <region>
+`
+var errNoSuchHost = `ERROR: RequestError: send request failed
+caused by: Post https://cloudformation.invalidRegion.amazonaws.com/: dial tcp: lookup cloudformation.invalidRegion.amazonaws.com on 192.168.1.1:53: no such host
+`
+
+func TestUninstallHelpFlag(t *testing.T) {
```
soulshake: Touche, renamed

cmd/convox/install_test.go
```diff
@@ -16,12 +16,52 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+var installUsage = `convox install: install convox into an aws account
```
soulshake: I want to test:

- that usage is printed when requested
- that usage is printed when there are invalid, missing or extra args
- that missing credentials files are handled
- that hostname resolution failures and "no such file or directory" errors are handled as expected
- that we don't panic in any case

I would be ok with a less strict usage test; for now I was just following the same pattern as in #1996.

cmd/convox/uninstall_test.go
```diff
@@ -0,0 +1,82 @@
...
+			Exit:    0,
+			Stdout:  Banner + "\n",
+			// Stderr: errNoSuchHost,
+		},
+
+		// FIXME: this command actually exits with a 'no such file or directory' error
+		test.ExecRun{
+			Command: "convox uninstall rackArg regionArg credentialsArg",
+			Stdout:  Banner + "\nReading credentials from file credentialsArg\n",
+			//Stderr: `ERROR: open credentialsArg: no such file or directory`,
```
soulshake: As I mentioned in the comment directly above `test.ExecRun`, when I run this command for real, the CLI exits 1 with `no such file or directory`. The test, however, doesn't capture that in either Stderr or Stdout, and I'm not sure why. Removing the commented `Stderr` anyway though.

.travis.yml
```diff
@@ -17,6 +17,7 @@ go:
   - 1.7
 
 before_install:
+  - pip install --user awscli
```
MiguelMoll: Not sure why this was added. Something change?

bin/test
```diff
@@ -6,6 +6,9 @@ go get -t ./...
 
 rm -f coverage.txt
 
+unset AWS_ACCESS_KEY_ID
+unset AWS_SECRET_ACCESS_KEY
```
MiguelMoll: Same here, what changed that these keys are removed?

.travis.yml
soulshake: I added this because the CLI prints a different message if the AWS CLI hasn't been installed than the one it prints if the AWS CLI is installed but not configured. I don't know how to mock-toggle this in local tests, so in other words, the test output will be different locally (since we all have the AWS CLI installed) vs in Travis (which doesn't).

bin/test
soulshake: If these aren't unset, then they are present in the test suite environment. In order to test rack installs when no credentials are found, we need there to not be any credentials in the host environment.

MiguelMoll: I could be missing some context but seems like the test case should handle that via `os.Unsetenv("VAR")`. Wondering about the side effects of including this in the test script.

soulshake: I don't think there are any side effects (since we're executing `bin/test`, not sourcing it). The envvar isn't unset in the host environment permanently, only in the environment of the bash process that's running `bin/test`.

IMO in any test where an env var is accessed, we should set it explicitly, e.g. via:

- `Command: "FOO=bar convox install -h"`
- `os.Setenv(...)`
- in the `Env` field of the `ExecRun` struct)

...rather than relying on the environment of the user running the tests (with the obvious exceptions where we _want_ the envvar to affect the test suite, like `PKG` and `TEST`).

MiguelMoll: I concur. 👍

cmd/convox/install_test.go
soulshake: Ok, I'm only testing one line of the help output in my latest commit.

PR #2037, opened 2017-03-09
add cf notifcation env var to the monitor

MiguelMoll:
Also send a notification event only if the system or formation updates are successful.


MiguelMoll:
[Autoscaling issues](https://trello.com/c/0ZqtvpTU/2091-autoscaling-issues)

PR #2039, opened 2017-03-09
'convox scale' should not exit silently if app not found

soulshake:
Before:

```
$ convox scale
$ echo $?
0
```

After:
```
$ convox scale
ERROR: rack not found
$ echo $?
1
```

### DIFF
--- a/cmd/convox/scale.go
+++ b/cmd/convox/scale.go
@@ -65,11 +65,17 @@ func cmdScale(c *cli.Context) error {
 			return stdcli.Error(fmt.Errorf("missing process name"))
 		}
 
-		displayFormation(c, app)
+		err = displayFormation(c, app)
+		if err != nil {
+			return stdcli.Error(err)
+		}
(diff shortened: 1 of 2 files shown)


soulshake:
['convox scale' should not exit silently if app not found #2039](https://trello.com/c/mxhUQrbH/2092-convox-scale-should-not-exit-silently-if-app-not-found-2039)

cmd/convox/scale.go
```diff
@@ -65,11 +65,17 @@ func cmdScale(c *cli.Context) error {
 			return stdcli.Error(fmt.Errorf("missing process name"))
 		}
 
-		displayFormation(c, app)
+		err = displayFormation(c, app)
+		if err != nil {
```
MiguelMoll: nit for consistency:  would we wrap these err checks:
```
if err := displayFormation(c, app); err != nil{
    return nil
}
```

PR #2040, opened 2017-03-10
Document build image in Development.md

soulshake:


soulshake:
[Document build image in Development.md on Rack #2040](https://trello.com/c/2Qa7GcMP/2094-document-build-image-in-development-md-on-rack-2040)

Development.md
```diff
@@ -116,6 +116,15 @@ i-146c2f97  on     active  2 hours ago  1   0.00%  3.21%
 i-c7de605c  on     active  2 hours ago  0   0.00%  0.00%
 ```
 
+## Build Image
+
+If you're working on the builder, you can set the [`BuildImage` Rack param](https://convox.com/docs/rack-parameters#buildimage) to a Docker image for the builder:
+
+```
+$ make builder
+$ convox rack params set BuildImage=convox/build:$(whoami)
```
MiguelMoll: Only users of the Convox docker hub would be able to use `make builder` without changes. Could we be more explicit there?

soulshake: @MiguelMoll do you have a suggestion? Or would it be enough just add `Only users of the Convox docker hub would be able to use make builder without changes.` in a comment there?

MiguelMoll: Something along the lines of:
```
If you're working on the builder, you can set the [`BuildImage` Rack param](https://convox.com/docs/rack-parameters#buildimage) to a Docker image for the builder. This is primarily used for development purposes only. General users should not set this parameter.

A developer with access to the Convox DockerHub organization could run:

make builder

Which creates an image of the build script and uploads it to DockerHub. It can be set vai

convox rack params set BuildImage=convox/build:$(whoami)
```

As a side not, I think https://convox.com/docs/rack-parameters#buildimage should make it explicit that this param is intended for developers only and general users should not set it. Seems users think its related to `BuildInstance` and set both of them.

PR #2041, opened 2017-03-10
Allow for shorter app names

ravenac95:
Turns out the rule for docker tags to be between 4-30 characters was not maintained for very long on docker. The rules have now changed to be 1-128 characters. See: https://github.com/docker/docker/pull/8447

I'm not sure how to add tests for this so some guidance on that would be great.

### DIFF
--- a/api/models/app.go
+++ b/api/models/app.go
@@ -101,7 +101,7 @@ func getAppByStackName(stackName string) (*App, error) {
 	return app, nil
 }
 
-var regexValidAppName = regexp.MustCompile(`\A[a-zA-Z][-a-zA-Z0-9]{2,29}\z`)
+var regexValidAppName = regexp.MustCompile(`\A[a-zA-Z][-a-zA-Z0-9]{0,127}\z`)
 
 func (a *App) IsBound() bool {
 	if a.Tags == nil {
--- a/api/models/app.go
+++ b/api/models/app.go
@@ -101,7 +101,7 @@ func getAppByStackName(stackName string) (*App, error) {
 	return app, nil
 }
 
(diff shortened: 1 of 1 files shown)


MiguelMoll:
Thanks for the PR @ravenac95, always appreciated. Curious if the length of the app name is blocking you in anyway? For the time being, we do not plan on changing the character limit for an app name. Docker was one limitation but there's also AWS(Cloudformation, ELB, etc) limits in play. Rack also appends other strings to the app name in various use scenarios.

I'll close the PR for now, but please feel free to continue the conversation here.

mwarkentin:
Just an extra data point - we had an app called `api` that we needed to rename because of this restriction, might be a common app name. :)

ravenac95:
@mwarkentin That's almost the exact reason. My company has a few repos with 3 letter acronyms. Also @MiguelMoll how would we feel about just allowing shorter names. I've actually tested it all locally and I know shorter names work. In terms of longer names I haven't done anything extensive with it. I imagine other people have three letter apps tho.

ravenac95:
@MiguelMoll It's not a blocker more like a major inconvenience because i have to add `-a appname` every time and would have to make sure the team knows what app name makes to what repo as opposed to being immediately discoverable

ravenac95:
The need for the rule change started here: https://github.com/convox-archive/kernel/issues/110 but then docker addressed it here: https://github.com/docker/docker/pull/8447. I don't see anything that addresses issues with AWS however at least in terms of making the name shorter. I can understand about long names however. More than 30 characters for an application name seems like more an inconvenience anyhow.

mwarkentin:
@ravenac95 FYI, you can "set" a different convox app name default than the current dir by setting the [`.convox/app` file](https://convox.com/docs/cli/). Hope that helps!

ravenac95:
@mwarkentin Heck yes. that I didn't know.

Issue #2008, opened 2017-03-06
Comments in .env cause convox env set to fail


SonOfBytes:
I get that this is probably a utility issue rather than a strict bug.  However we are using multiple env files injected by our build pipeline that have comments in them to help people update them.  Blank lines also aid in readability.

Having comments (lines that start with #) in .env work with `docker-compose` and `convox start`

```
$ convox env set < .env
Updating environment... ERROR: Environment variables should be defined in key=value format. You specified: # A line comment
```

Blank lines also cause problems

PR #2043, opened 2017-03-10
Importing .env with comments and blank lines breaks env set

SonOfBytes:
Importing .env with comments and blank lines breaks
`convox env set <.env`

This creates a new parsing function to validate .env lines and has unit tests in env_ut_test.go.

A separate _ut_test file was use to keep it apart from the integration tests in _test

This should support various empty/whitespace only lines, and simple comment lines that begin with #

See issue https://github.com/convox/rack/issues/2008


soulshake:
[Review Rack PR (Importing .env with comments and blank lines breaks env set #2043)](https://trello.com/c/dxJ8BVas/2096-review-rack-pr-importing-env-with-comments-and-blank-lines-breaks-env-set-2043)

cmd/convox/env_ut_test.go
```diff
@@ -0,0 +1,44 @@
+package main
+
+import (
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestParseEnvLine(t *testing.T) {
```
MiguelMoll: nit: This test can go with the other env tests inside `env_test.go`. Nice to keep them together in one file.

cmd/convox/env.go
```diff
@@ -282,3 +290,25 @@ func cmdEnvUnset(c *cli.Context) error {
 
 	return nil
 }
+
+func parseEnvLine(line string) (key string, value string, err error) {
+	// Deal with empty lines
+	var m bool
+	if m = regexp.MustCompile(`^\s*$`).MatchString(line); m {
+		return
```
MiguelMoll: nit: Seems the variables declared in the return statement are not used anywhere. To be consistent with the return statements, can we remove the named return variables and `return "","", nil` here? Same with line 303.

cmd/convox/env_test.go
```diff
@@ -160,7 +160,7 @@ func TestEnvSetStdin(t *testing.T) {
 		test.Http{
 			Method: "POST",
 			Path:   "/apps/myapp/environment",
-			Body:   "foo=bar\nping=pong\n",
+			Body:   "# Comment followed by blank line\n\nfoo=bar\nping=pong\n",
```
MiguelMoll: This line should be reverted back as `foo=bar\nping=pong\n` is what's returned by the test server.

cmd/convox/env.go
SonOfBytes: agreed - I liked named vars for the self documenting properties, you are right to it should match the style of the rest of the code.  I have added a 1 liner documenting description.

cmd/convox/env_test.go
SonOfBytes: I did update the fixture `manifest\fixtures\env-test.env` which I believe is what is fed to the test server - does this mean that this is coded elsewhere?

cmd/convox/env_ut_test.go
SonOfBytes: done

cmd/convox/env_test.go
SonOfBytes: ok - I see this breaks travis-ci :(

MiguelMoll:
Thanks for the changes. Also, I just remembered that there is some logic server side to handle some edge cases. If it's OK with you, I can take it from here and move your changes server side so every env set call can benefit from these changes (not just the CLI.

SonOfBytes:
Great! If I can see the changes server side I will be able to learn from your carrying of the baton 😀

MiguelMoll:
Moving these changes forward via #2055. Thanks again for the PR!

PR #2048, opened 2017-03-11
Respect --app flag regardless of position

soulshake:
Before (convox drops `-a`, `--app` flag if passed in position 0):

```
$ convox -a foo releases
ERROR: no such app: rack
```

After:

```
$ convox -a foo releases
ERROR: no such app: foo
```

We were already doing this with the `--rack` flag via `getRackFlag()`. I replaced this with a more generic function `RecoverFlag` which takes a list instead of a string (because we need to check for both `-a` and `--app`).
**Important:** I also modified `stdcli.ParseOpts`: it previously only handled double-dash prefix, now it treats an item with only a single dash prefix as a flag. I don't _think_ this will cause issues, but maybe you know something I don't.

### DIFF
--- a/cmd/convox/stdcli/stdcli.go
+++ b/cmd/convox/stdcli/stdcli.go
@@ -115,6 +115,23 @@ func Debug() bool {
 	return false
 }
 
+// RecoverFlag allows us to capture things like --app FOO which would otherwise be discarded by urfave/cli if passed in position 0
+func RecoverFlag(c *cli.Context, flagNames []string) string {
+	for _, flagName := range flagNames {
+		f := c.String(flagName)
+		if f != "" {
+			return f
+		}
+
+		f = ParseOpts(os.Args)[flagName]
+		if f != "" {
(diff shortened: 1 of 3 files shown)


soulshake:
[Respect --app flag regardless of position #2048](https://trello.com/c/nIdf1T3w/2106-respect-app-flag-regardless-of-position-2048)

cmd/convox/stdcli/stdcli.go
```diff
@@ -115,6 +115,23 @@ func Debug() bool {
 	return false
 }
 
+// RecoverFlag allows us to capture things like --app FOO which would otherwise be discarded by urfave/cli if passed in position 0
+func RecoverFlag(c *cli.Context, flagNames []string) string {
```
MiguelMoll: We could change the flags to a variadic variable:

`RecoverFlag(c *cli.Context, flags ...string) string`

This way when calling the function, we can `RecoverFlag(c, "a", "app")`

soulshake: Good point, fixed

PR #1989, opened 2017-03-01
[RELEASE] 20170228

MiguelMoll:
## Pull Requests
  - closes #1632 Update cli pkg and fix panic in container env [@awsmsrc]
  - closes #1958 allow switching to internal balancers [@ddollar]
  - closes #1983 check if instance to drain has any tasks [@MiguelMoll]
  - closes #1995 Fix `convox instances` showing 0% CPU usage [@ianmalott]

## Milestone Release
- [x] Release branch
- [x] Pass CI
- [x] Code review
- [ ] Merge into master
- [ ] Close milestone
- [ ] Release master
- [ ] Record release number: 
- [ ] Pass CI
- [ ] Update demo
- [ ] Deploy demo/httpd
- [ ] Deploy demo/rails
- [ ] Update staging
- [ ] Deploy staging/site-staging
- [ ] Deploy staging/console-staging
- [ ] Update [release notes](https://github.com/convox/rack/releases)
- [ ] Documentation review
- [ ] Publish release
- [ ] Release CLI

PR #2049, opened 2017-03-11
Return error if CONVOX_DEBUG, CONVOX_WAIT or RACK_PRIVATE aren't set to valid boolean values

soulshake:
Before:
```
$ CONVOX_WAIT=foo convox switch
squirrels/dev
```

After:
```
$ CONVOX_WAIT=foo convox switch
WARNING: 'foo' is not a valid value for environment variable CONVOX_WAIT (expected: [true false 1 0 ])
squirrels/dev
```

### Context

This has something to do with urfave/cli not accepting `yes`/`no` as valid boolflag values, I think. See also [this rack change](https://github.com/convox/rack/pull/1989/files#diff-effb1c59e116f169d0acdeb0a44d1f60) and [this site change](https://github.com/convox/site/pull/632).

### DIFF
--- a/cmd/convox/stdcli/stdcli.go
+++ b/cmd/convox/stdcli/stdcli.go
@@ -104,8 +104,21 @@ func New() *cli.App {
 	}
 
 	app.Writer = DefaultWriter
+	app.Before = ValidatePreconditions(CliCheckEnv)
 
 	return app
+}
+
+// ValidatePreconditions runs one or more cli.BeforeFuncs where called in Command.Before
+func ValidatePreconditions(preconditions ...cli.BeforeFunc) cli.BeforeFunc {
+	return func(c *cli.Context) error {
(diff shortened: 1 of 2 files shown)


soulshake:
[Print a warning if CONVOX_DEBUG, CONVOX_WAIT or RACK_PRIVATE aren't set to valid boolean values #2049](https://trello.com/c/BIa5czZl/2107-print-a-warning-if-convox-debug-convox-wait-or-rack-private-aren-t-set-to-valid-boolean-values-2049)

MiguelMoll:
This is a great addition. Any reason to warn instead of error-ing out? Seems a little risky to continue the command if the value type isn't what the cli pkg expects.

soulshake:
No particular reason for a warning vs. an error. I'll make it return an error.

PR #2050, opened 2017-03-11
Use CONVOX_DEBUG rather than DEBUG

soulshake:
`DEBUG` is too generic to rely on for CLI debug printing purposes. If this PR is accepted, the CLI will use `CONVOX_DEBUG` instead.

### DIFF
--- a/provider/aws/aws.go
+++ b/provider/aws/aws.go
@@ -120,7 +120,7 @@ func (p *AWSProvider) config() *aws.Config {
 		config.Endpoint = aws.String(p.Endpoint)
 	}
 
-	if os.Getenv("DEBUG") == "true" || os.Getenv("CONVOX_DEBUG") == "true" {
+	if os.Getenv("DEBUG") == "true" {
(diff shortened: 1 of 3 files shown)


provider/aws/aws.go
```diff
@@ -120,7 +120,7 @@ func (p *AWSProvider) config() *aws.Config {
 		config.Endpoint = aws.String(p.Endpoint)
 	}
 
-	if os.Getenv("DEBUG") == "true" {
+	if os.Getenv("DEBUG") == "true" || os.Getenv("CONVOX_DEBUG") == "true" {
```
MiguelMoll: I'm OK going with either or, but we should pick one. Having both can cause unexpected behavior while debugging.

Believe @ddollar made some changes in this regards a while back. Any thoughts?

MiguelMoll: `CONVOX_DEBUG` it is. Lets remove it from this line though as it won't have any affect in a rack and could lead to confusion. We could in another PR change this throughout the rack.

PR #2052, opened 2017-03-13
Restricted Permissions for Custom Topic Role

ravenac95:
A few things bothered me about the permissions the lambda function has for convox apps. Namely, the S3 environment was open to the public in order to support the lambda function, so that led me down the road of discovery and eventually I figured out that the lambda function was also an admin for everything in a single account. 

This PR introduces a few changes to remedy those concerns:

* The IAM Role for the Lambda Function is now restricted to just the commands it calls. I may have missed something but in all my manual tests it seems to work.
* The env file uploaded to S3 is no longer being uploaded with a `public-read` ACL

I'm not sure how to run the integration tests, but I did run the tests on my own development rack.


MiguelMoll:
Nice improvements and thanks for the PR. I'm going to test it out and dig in a little more later today. After a quick, the changes look promising.

MiguelMoll:
[Restricted Permissions for Custom Topic Role](https://trello.com/c/pg9hsdLH/2146-restricted-permissions-for-custom-topic-role)

MiguelMoll:
Using #2105 to carry these changes forward. @ravenac95 Thanks for these improvements. 👍

PR #2057, opened 2017-03-13
Ami updates

MiguelMoll:
Seems something changed in the docs that the `current-ecs-amis` script wasn't returning anything. After time boxing it for a while, this is my best effort with jq. Any improvements is appreciated.


MiguelMoll:
[AMI Updates](https://trello.com/c/74e5clrY/2121-ami-updates)

PR #2064, opened 2017-03-16
bash completion script initial commit

soulshake:


soulshake:
[Bash completion](https://trello.com/c/WvtT4MZs/1894-bash-completion)

nzoschke:
I tested this out. It's working well on OS X and Homebrew bash completion package. App name, build id, argument completion is sick!

My main question is how this interacts with the bash completion stuff that [urfave/cli offers](https://github.com/urfave/cli#bash-completion) and [we document here](https://convox.com/docs/cli/#osx--homebrew).

Is this a replacement? How does the convox script get generated?

Some other things I noticed...

- [ ] I was logged out from Console and was getting silent failures on app name completion. Logging in cleared it up.
- [ ] If i'm not in an app directory, build id completion doesn't work



- [ ] Docs review. https://convox.com/docs/cli/#osx--homebrew is out of date

soulshake:
Thanks for feedback @nzoschke!

> My main question is how this interacts with the bash completion stuff that [urfave/cli offers](https://github.com/urfave/cli#bash-completion) and [we document here](https://convox.com/docs/cli/#osx--homebrew).

We can (and, I think, should) easily move at least some of the completion logic to the CLI. `urfave/cli` allows you to hook a custom function in whenever `--generate-bash-completion` is passed, allowing completion support for anything that can be printed from `cli.Context` which would include most of what we need (apps, resources, etc). I don't think it would be any faster since it would be making the same calls under the hood, but maybe the more we can move over, the better (because having the completion logic in a separate script is harder to maintain, there might be some shell-specific issues, etc).

On the other hand, having it in a separate file will mean less dependency on this version of urfave/cli in case we decide to upgrade or change. And I'm not sure how straightforward it will be to move this to the CLI, but I can investigate anytime (though preferably after #1996 is dealt with).

> Is this a replacement? How does the convox script get generated?

This can be a replacement, or supplemental. We can take better advantage of `--generate-bash-completion` by hardcoding things like the list of API endpoints in the CLI rather than in the completion script.

It doesn't get generated at all; I did it manually. I don't think it would be very easy to automate it either, though maybe it's possible.

>  I was logged out from Console and was getting silent failures on app name completion. Logging in cleared it up.

Pushed a commit that makes it autocomplete a message `NO_AUTH_FOUND_PLEASE_LOG_IN`, WDYT?

>  If i'm not in an app directory, build id completion doesn't work

Pushed a fix for this, it should work now.

PR #2065, opened 2017-03-16
Prevent a panic if no host config is found with a nicer error message

MiguelMoll:
New output would look like:

```
convox apps
ERROR: no host config found, try `convox login`
```

### DIFF
--- a/cmd/convox/main.go
+++ b/cmd/convox/main.go
@@ -112,12 +112,10 @@ func currentRack(c *cli.Context) string {
 func rackClient(c *cli.Context) *client.Client {
 	host, password, err := currentLogin()
 	if err != nil {
-		stdcli.Error(err)
-		return nil
+		stdcli.Errorf("%s, try `convox login`", err)
 	}
 
 	cl := client.New(host, password, c.App.Version)
-
 	cl.Rack = currentRack(c)
 
 	return cl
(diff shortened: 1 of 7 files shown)


MiguelMoll:
[Rack segfault when not logged in](https://trello.com/c/PZXakHrK/2099-rack-segfault-when-not-logged-in)

MiguelMoll:
[Rack segfault when not logged in](https://trello.com/c/PZXakHrK/2099-rack-segfault-when-not-logged-in)

cmd/convox/main.go
```diff
@@ -112,12 +112,11 @@ func currentRack(c *cli.Context) string {
 func rackClient(c *cli.Context) *client.Client {
 	host, password, err := currentLogin()
 	if err != nil {
-		stdcli.Error(err)
-		return nil
```
awsmsrc: i get why this was changed like this, that function is called everywhere.. Should it just return an error "the go way?"

awsmsrc:
I've added a comment, I can see how this fixes the problem. 

I'm ok with the change but it isn't the standard "the go way"

cmd/convox/main.go
MiguelMoll: At first, I went with this function returning an error. But as you mentioned, it's called everywhere in a chained fashion. All those calls would have to add error checking, which is fine in itself but then reduces the usefulness of this helper IMHO.

So I went with the function exiting out as the CLI can't do anything without a working client.

MiguelMoll:
Fixed via https://github.com/convox/rack/pull/2074

PR #2066, opened 2017-03-16
Listing processes use a cache to prevent throttling

### LINKED ISSUE #1990: AWS rate limit errors when running `convox ps` on an app with many processes

### LINKED ISSUE #2021: `convox ps info` doesn't work
```
$ convox ps
ID            NAME  RELEASE      STARTED       COMMAND
dfecd78e33c7  wopr  RRANZYKRPSM  17 hours ago  sh -c /src/bin/wopr

$ convox ps info dfecd78e33c7
ERROR: response status: 404 404 page not found
```

MiguelMoll:
When gathering information for a large number of processes (e.g > 300), API throttling becomes an issue. This is addressed by caching various resources that are unlikely to change while gathering the process data. That includes container definitions and instances that are shared amongst the processes. I was able to gather over ~420 processes without being throttled back to back. Fixes #1990

Other fixes include:
- Bumping CircleCI golang to 1.7.5 as rack itself
- Added the `ProcessGet` interface so that `convox ps info <ID>` works. Fixes #2021
- Added some logging to the `SystemGet` interface

### DIFF
--- a/api/controllers/processes.go
+++ b/api/controllers/processes.go
@@ -13,22 +13,6 @@ import (
 	"golang.org/x/net/websocket"
 )
 
-func ProcessList(rw http.ResponseWriter, r *http.Request) *httperr.Error {
-	app := mux.Vars(r)["app"]
-
-	ps, err := models.Provider().ProcessList(app)
-	if provider.ErrorNotFound(err) {
-		return httperr.NotFound(err)
-	}
-	if err != nil {
-		return httperr.Server(err)
-	}
-
-	sort.Sort(ps)
-
-	return RenderJson(rw, ps)
-}
-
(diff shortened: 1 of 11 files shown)


MiguelMoll:
[Avoid AWS rate limits on `convox ps` for apps with lots of processes](https://trello.com/c/okPBUlXZ/2046-avoid-aws-rate-limits-on-convox-ps-for-apps-with-lots-of-processes)

PR #2068, opened 2017-03-16
Parameterize ebs encryption

mattmanning:
This makes EBS encryption configurable instead of hard-coded to on. It defaults to off, but if you want to turn it on you can run:

```
convox rack params set EncryptEbs=Yes
```

### DIFF
--- a/provider/aws/dist/rack.json
+++ b/provider/aws/dist/rack.json
@@ -1409,7 +1409,7 @@
           {
             "DeviceName": "/dev/xvdb",
             "Ebs": {
-              "Encrypted": { "Ref": "EncryptEbs" },
+              "Encrypted": { "Fn::If": [ "EncryptEbs", "true", "false" ] },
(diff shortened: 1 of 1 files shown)


mattmanning:
[Parameterize EBS encryption](https://trello.com/c/KTbEsM69/2145-parameterize-ebs-encryption)

MiguelMoll:
Question that comes to mind is will users that currently can't update because the current default is true, will they be able to update to this version and take advantage of the parameter?

I see a few work arounds, but it might not be straight forward.

mwarkentin:
@MiguelMoll has a point here, I think we'd be in a chicken / egg situation where we wouldn't be able to get to the version with the parameter in order to set it. Maybe do a single required release where it defaults to off and introduces the param?

mattmanning:
Changed the default to "No". I think we could do this one as a required release and then set it back to "Yes" for future releases.

PR #2069, opened 2017-03-16
build instance support for run and boot command

### LINKED ISSUE #2046: Build Instances don't use InstanceRunCommand
We use `InstanceRunCommand` to download credentials (since we don't have rack-level credentials), add additional SSH keys, and kick off a bootstrap script that configures monitoring and security tooling.

Experimentation shows that builder instances don't run this command, meaning we're running blind (save CloudWatch) on the state of those instances. 

From Slack:

> ddollar [11:47 AM] 
> @dlanger it appears not, though i cant think of a reason why they shouldnt
> 
> ddollar [11:47]  
> https://github.com/convox/rack/blob/master/provider/aws/dist/rack.json#L1560   https://github.com/convox/rack/blob/master/provider/aws/dist/rack.json#L1418

MiguelMoll:
Fixes #2046

### DIFF
--- a/provider/aws/dist/rack.json
+++ b/provider/aws/dist/rack.json
@@ -1450,7 +1450,19 @@
             ] ] },
             "  - mkdir -p /etc/convox\n",
             "  - echo -e '/var/log/docker {\\n  rotate 7\\n  daily\\n  nocompress\\n  copytruncate\\n}' >> /etc/logrotate.d/docker\n",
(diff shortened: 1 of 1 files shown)


MiguelMoll:
[Make build instances respect InstanceRunCommand](https://trello.com/c/FnL87RGY/2102-make-build-instances-respect-instanceruncommand)

PR #2077, opened 2017-03-20
Respect CONVOX_WAIT and --wait on 'convox rack params set'

soulshake:


soulshake:
[Respect CONVOX_WAIT and --wait on 'convox rack params set'](https://trello.com/c/tVeycToV/2158-respect-convox-wait-and-wait-on-convox-rack-params-set)

cmd/convox/rack.go
```diff
@@ -215,7 +221,19 @@ func cmdRackParamsSet(c *cli.Context) error {
 		return stdcli.Error(err)
 	}
 
-	fmt.Println("OK")
+	if c.Bool("wait") {
+		stdcli.Startf("Waiting for completion")
+
+		// give the rack a few seconds to start updating
+		time.Sleep(5 * time.Second)
```
MiguelMoll: nit: Seems like a long wait time. Was two or three seconds tested?

cmd/convox/rack.go
```diff
@@ -215,7 +221,19 @@ func cmdRackParamsSet(c *cli.Context) error {
...
+
+		if err := waitForRackRunning(c); err != nil {
+			return stdcli.Error(err)
+		}
+
+		stdcli.OK()
```
MiguelMoll: Think the `fmt.Println("OK")` (or something to that effect) should be incorporated for when wait isn't set.

MiguelMoll:
Rebase would be needed too.

cmd/convox/rack.go
soulshake: Sorry, I don't see what you mean here. I was trying to use the same style as [here](https://github.com/convox/rack/blob/master/cmd/convox/rack.go#L304-L315), is that not the way to go?

cmd/convox/rack.go
soulshake: I just used the same sleep time as [here](https://github.com/convox/rack/blob/master/cmd/convox/rack.go#L304-L315). I haven't tested with 2-3 seconds, but can if you think 5 seconds is too long here.

MiguelMoll: That's a good example. Thinking something like:

```
stdcli.Startf("Updating parameters... ")
// stuff happens here
stdcli.Wait("OK")
```

MiguelMoll: Sounds good to me. 👍

cmd/convox/rack.go
MiguelMoll: That's a good example. Thinking something like:

```
stdcli.Startf("Updating parameters... ")
// lines 216 -222 go here
stdcli.Wait("OK")
```

soulshake: Oh I see. Pushed again, is that better?

PR #2080, opened 2017-03-21
default to two cpu shares

MiguelMoll:


MiguelMoll:
[Investigate strange CPU performance behavior.](https://trello.com/c/5295pDXG/1969-investigate-strange-cpu-performance-behavior)

nzoschke:
LGTM, but it's tricky so let me make sure I understand...

The minimum value of CPU shares according to Linux is "2", and the max is "1024".

A value of "2" means that a container is guaranteed only 2 CPU shares, but can burst up to 1024 shares as long as it doesn't step on other container guarantees. This means that all the containers with a value of "2" should be very "nice" and not hog CPU resources.

A value of "1024" means that the container is effectively given an entire CPU core. Though Linux still has some rules about two containers with "1024" sharing a single core fairly.

The value of "2" is a reasonable default for every container, unless a user explicitly wants to enable a container to guarantee more CPU resources if there is contention.

On one ECS agent, a TD value of "null" or "0" used to be converted to "1024", and a value of "1" used to be converted to "2".

On the latest ECS agent, "null", "0", "1" and "2" are converted to "2".

This logic defaults "null", "0", "1" and "2" explicitly to "2" to provide the reasonable default for any version of the ECS agent.

PR #2087, opened 2017-03-21
Automatically update Lambda runtime version in regions that support updates

nzoschke:
In some regions we can now update the custom topic runtime from Node.js 0.10 to Node.js 4.3. These CF template changes automatically update to Node.js 4.3 in those regions.

## Release Playbook
- [x] Rebase against master
- [x] Release branch (20170322150845-custom-topic-update)
- [x] Pass CI (https://circleci.com/gh/convox/rack/2110)
- [x] Code review
- [ ] Merge into master
- [ ] Release master
- [ ] Pass CI
- [ ] Publish release

For this one we should also:

- [ ] Mark release as required


api/models/fixtures/agent_label.json
```diff
@@ -75,6 +75,42 @@
         },
         "Yes"
       ]
+    },
+    "RegionHasCustomTopicUpdate": {
```
ddollar: This conditional name doesn't really make sense. Would probably be better to add a LambdaRuntime attribute to one of the region attribute mappings (this is how we determine EFS, etc)

ddollar: Just looked at rack.json. Should add a `LambdaRuntime` attribute next to `Ami` in `RegionConfig`

mattmanning: @ddollar I don't see how to avoid a conditional of some kind. We can't just set the runtime to whatever is in the RegionConfig map because it would override the Parameter setting for newer Racks.

We still need to be able to do an If like this somehow:

```
          "Fn::If": [
             "RegionHasCustomTopicUpdate",
             "nodejs4.3",
             {
               "Ref": "CustomTopicRuntime"
             }
           ]
```

nzoschke: The name started by copying the `RegionHasEFS` conditional in rack.json.

I do see how the mapping config makes cleaner CF json.

@mattmanning we'd still have a conditional like `EventRuleForECS` that uses `Fn::FindInMap`.

Seems like Ami, ThirdAvailabilityZone, EventRuleForECS, EFS, and this Lambda logic could all be in one map.

Why do `AvailabilityZoneConfig` and `RegionConfig` not cover the same regions?

I'll take a stab at cleaning this up more.

ddollar: If you need a conditional then it should be named something that describes it better. This conditional seems to suggest that it is whether or not the region allows custom topic updates but that is not the case. This conditional describes whether or not the region allows the runtime attribute of a lambda function to be updated.

nzoschke: > This conditional seems to suggest that it is whether or not the region allows custom topic updates

That is the case.

We have always been able to update the runtime attribute of Lambda, and already did so for things like the lifecycle handler in https://github.com/convox/rack/pull/1519/files.

We haven't been able to update the runtime attribute of CF custom handlers due to a CloudFormation limitation. That limitation of CF is now gone in some regions.

When it is gone in every region we can remove the mapping and conditional logic altogether, and hardcode 'nodejs4.3' for all lambda functions, custom topics and otherwise.

nzoschke:
@ddollar I moved everything over to a mapping, which feels like much cleaner CF but could use a close review.

I renamed the conditional to `RegionHasLambdaUpdate` but now thinking that name is even worse because of https://github.com/convox/rack/pull/2087/files/a2b123ec6e537bc38bf2d42a2fd2fd644e21d4d5#r107316706

ddollar:
That is not the case. The change being rolled out allows the runtime
attribute of lambda functions to be updated without requiring replacement
of the resource. Replacement of custom topics is not allowed (and is still
not allowed, even in these regions)

nzoschke:
Ah that makes sense. The other handlers we could update to nodejs4.3 because a replacement was ok.

Does `RegionHasLambdaUpdate` and `LambdaUpdate` make sense then?

nzoschke:
Just caught something in app.tmpl, with 2 mappings blocks. Updating...

nzoschke:
[Prepare for Lambda runtime change](https://trello.com/c/JE9V5wTT/2184-prepare-for-lambda-runtime-change)

nzoschke:
This passed CI in all regions except for one due to a transcription bug. Bug is addressed in https://github.com/convox/rack/pull/2087/commits/9c8f82245c62672115b55c1b709dcbbc39bc3dfc and running CI again

nzoschke:
- [ ] Code review around CF map
- [ ] Conditional name
- [ ] Required release

PR #2088, opened 2017-03-21
use stat.Size() alternate instead of seek

MiguelMoll:
Even though os.Stdin is of type io.ReadSeeker, it's illegal to call Seek() on it. This change uses the file's stat size and falls back to ReadSeeker if needed.

This allows for both:
```
> convox builds import < export.tgz
Uploading: 65.19 MB / 65.19 MB [========================] 100.00 % 53s
Importing build... Release: RDGJVQIQXNB
```
and
```
> convox builds export BWJQAAKDFVA | convox builds import
Exporting BWJQAAKDFVA... OK
Release: RETMQURAGRT
```

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -150,14 +150,28 @@ func (c *Client) PostMultipart(path string, opts PostMultipartOptions, out inter
 
 		var size int64
 		for _, file := range opts.Files {
-			if rs, ok := file.(io.ReadSeeker); ok {
-				s, err := io.Copy(ioutil.Discard, rs)
+			switch f := file.(type) {
+			// Seek() is illegal for pipded /dev/stdin so lets try to get the stat size first
+			case *os.File:
+				stat, err := f.Stat()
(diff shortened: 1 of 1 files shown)


MiguelMoll:
Another option would be to load all data into a buffer regardless of type and read the size that way. Seems inefficient but it's an option.

ddollar:
These images can be multiple gigs. The rack process is hard limited to
128MB of RAM

MiguelMoll:
Even being a client side change, can't assume it won't run in a rack so I concur. Any thoughts on the change itself?

SonOfBytes:
I'd advise against in memory. If the concern is lack of feedback then maybe a /-\| (and repeat) spinner every MB to show something is happening?

MiguelMoll:
[Build import/export flakiness](https://trello.com/c/T5v6zHd0/1843-build-import-export-flakiness)

PR #2089, opened 2017-03-22
use stack's resource for log group

MiguelMoll:
This allows for logs to be availabe if the stack is in rollback-in-progress state. Output values are not avaliable during that time.


MiguelMoll:
[convox logs unavailable during rollback](https://trello.com/c/V8hslfMP/2143-convox-logs-unavailable-during-rollback)

PR #2093, opened 2017-03-23
Fix UDP ports before TCP ports

### LINKED ISSUE #2013: ELB creation should not be dependent on UDP port ordering
If a UDP port is specified first in the list of ports that also include TCP ports in `docker-compose.yml` it causes a CloudFormation error

spencerhakim:
`LoadBalancers` should only consider TCP ports. Should fix #2013.


MiguelMoll:
Quick rebase when you get a chance please.

spencerhakim:
@MiguelMoll Done.

MiguelMoll:
Moving forward via #2118.

PR #2097, opened 2017-03-24
remove env vars from dynamo for new releases, only use encrypted s3

MiguelMoll:


MiguelMoll:
[Encrypt env in dynamo](https://trello.com/c/euahY6OX/2212-encrypt-env-in-dynamo)

nzoschke:
Patch looks good. This one should have some careful QA. Perhaps something like:

* On old version, create app and create release (R1) with `convox env set FOO=bar`
* Deploy app and verify `convox exec env`
* Update rack
* Verify `convox env`
* Create new release (R2) with `convox env set FOO=baz`
* Deploy app and verify `convox exec env`
* Roll back to R1 and verify `convox exec env`
* Roll forward to R2 and verify `convox exec env`

MiguelMoll:
Checklist plays out as expected. 👍

provider/aws/releases.go
```diff
@@ -45,7 +45,30 @@ func (p *AWSProvider) ReleaseGet(app, id string) (*structs.Release, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if a.Parameters["Key"] != "" {
+		cr := crypt.New(p.Region, p.Access, p.Secret)
+
+		if d, err := cr.Decrypt(a.Parameters["Key"], data); err == nil {
+			data = d
+		}
```
nzoschke: is this missing an error check?

Issue #2015, opened 2017-03-07
Better visibility for dedicated build instances


mattmanning:
Dedicated build instances don't show up in the `convox instances` output and don't seem to be tagged in any unique way that makes them easy to identify.

soulshake:
Closing as duplicated of https://github.com/convox/rack/issues/1948

PR #2100, opened 2017-03-25
Add BuildInstance to 'convox rack' output

soulshake:
## Before

```
$ convox rack
Name			dev
Status			running
Version			20170324201702
Region			us-east-1
Count			3
Type			t2.small
```

## After

```
$ convox rack
Name			dev
Status			running
Version			20170324201702
Region			us-east-1
Count			3
Type			t2.small
BuildInstance		t2.small
```

Addresses #2015

### DIFF
--- a/api/controllers/system_test.go
+++ b/api/controllers/system_test.go
@@ -17,13 +17,13 @@ import (
 func TestSystemShow(t *testing.T) {
 	models.Test(t, func() {
 		system := &structs.System{
-			Count:         3,
-			Name:          "test",
-			Region:        "us-test-1",
-			Status:        "running",
-			Type:          "t2.small",
-			Version:       "dev",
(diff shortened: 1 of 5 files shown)


soulshake:
[Add BuildInstance to 'convox rack' output](https://trello.com/c/4Fxbq0X6/2219-add-buildinstance-to-convox-rack-output)

MiguelMoll:
Code is simple enough. Only thing I can nitpick about is displaying `Type` vs `BuildInstance`. Should the former be something like `ClusterInstance`? The latter `BuildInstanceType`? Not really a big deal but the disparity is really standing out.

mwarkentin:
Ideally we also see something in `convox instances` as well, along with the capability to use stuff like `convox instances ssh <build instance id>` so that it's up to par w/ the regular instance types.

PR #2102, opened 2017-03-27
Add support for CONVOX_WAIT to resource creation

SonOfBytes:
This adds support for using CONVOX_WAIT to resources creation, update, deletion.  However, --wait flag may not work with create and update (only the env var).  This may be due to "SkipFlagParsing: true" on these subcommands.

### DIFF
--- a/cmd/convox/flags.go
+++ b/cmd/convox/flags.go
@@ -11,3 +11,9 @@ var rackFlag = cli.StringFlag{
 	Name:  "rack",
 	Usage: "rack name",
 }
+
+var waitFlag = cli.BoolFlag{
+	Name:   "wait",
+	EnvVar: "CONVOX_WAIT",
+	Usage:  "wait for change to finish before returning",
+}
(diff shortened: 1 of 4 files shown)


SonOfBytes:
I believe this works - tho the support for the --wait flag via options smells a little.  Ideas?

MiguelMoll:
Thinking out loud about this one. Does it make sense to wait for a resource? Usually they are created ahead of time. What would be the use case for a wait? Scripting comes to mind.

Normally using `options` would be a red flag, but I think its its a good work around given `SkipFlagParsing`.

SonOfBytes:
@MiguelMoll good question.

This came out of a need to automate the provisioning of multiple environments in a deployment pipeline.  However, containers were not able to come up because they did not have the resource endpoint of the RDS instance they were trying to use.  This meant that a script had to keep polling the status of the convoy RDS resource before proceeding to deploying the app.

SonOfBytes:
@MiguelMoll is this dead in the water given all efforts are going into praxis?

MiguelMoll:
@SonOfBytes Whenever you can, could you update this branch? Thanks.

SonOfBytes:
hopefully done correctly @ddollar

PR #2105, opened 2017-03-27
Restrict custom topic role

MiguelMoll:
Taking #2052 forward with changes. Changed a few nitpicks. Also added some more app and rack permissions that require an `*` resource. @ravenac95 Any feedback is welcomed.


MiguelMoll:
[Restricted Permissions for Custom Topic Role](https://trello.com/c/pg9hsdLH/2146-restricted-permissions-for-custom-topic-role)

ravenac95:
@MiguelMoll changes don't seem bad but I'm curious, are some of the wildcard requirements not capable of being restricted to a specific account? I'm not that adamant about it but I would definitely prefer something more restrictive that more permissive.

MiguelMoll:
That sounds reasonable. I can try messing around with the `Condition` element. AWS docs and my testing pretty much left little else with `Resource`.

ravenac95:
Ahh ok. Ya I think as it stands the IAM role is worlds better than what it was previously.

nzoschke:
This looks really good in general, but we need more time to carefully review the reduced IAM to make sure there will be no regressions.

MiguelMoll:
Better approach would be to handle this with Praxis and it's custom role. Editing the IAM at this point could break existing features.

PR #2108, opened 2017-03-27
Reduce number of API calls during uninstall

MiguelMoll:
This change allows uninstall to make API calls only when needed, specifically for stack buckets and events. Another fix is to uninstall both a `service` and it's rename `resource`.

Also removed `distinctId` that was being passed around but not used.


MiguelMoll:
[Uninstall throttling](https://trello.com/c/h8lsRq3P/2228-uninstall-throttling)

PR #2075, opened 2017-03-20
encryption flag for `convox resources` for mysql and postgres

SonOfBytes:
Adds encryption flag to `convox resources` for mysql and postgres.  It updates the convox cmd and cloudformation stacks.  No tests.  Uses Default KMS.

Tested by hand and seems to do the right thing.

This is based on the work by @alex posted in this issue https://github.com/convox/rack/issues/1335

Also tracked here https://github.com/convox/rack/issues/1834

This uses the default KMS - should probably use the convox kms - how best to reference this in cloud formation and is this the right approach?  My feeling is that user custom KMS key should be added when there is a user need.

What would be the best way to test or should we leave the testing as is?

PR #2110, opened 2017-03-28
Encrypted rds volumes

### LINKED ISSUE #1335: add `--encrypted` option for postgres services
Some apps have privacy requirements that require their data to be encrypted at-rest.

Convox currently doesn't support this, so the only way to have an encrypted db is to do so outside of the convox cli and hope it plays nice with convox.

adding `--encrypted` to `convox services create postgres` would enable this

MiguelMoll:
Carrying #2075 forward.

Fixes #1335


MiguelMoll:
[Review encryption flag for `convox resources` for mysql and postgres](https://trello.com/c/xE5kwDCs/2231-review-encryption-flag-for-convox-resources-for-mysql-and-postgres)

provider/aws/templates/resource/mysql.tmpl
```diff
@@ -15,6 +16,16 @@
         "Default" : "app",
         "Description" : "Default database name"
       },
+      "Encryption": {
```
ddollar: We don't need a boolean here. Use the presence/blankness of EncryptionKey

provider/aws/resource.go
```diff
@@ -34,7 +34,25 @@ func (p *AWSProvider) ResourceCreate(name, kind string, params map[string]string
 	var req *cloudformation.CreateStackInput
 
 	switch s.Type {
-	case "memcached", "mysql", "postgres", "redis", "sqs":
+	case "memcached", "redis", "sqs":
+		req, err = p.createResource(s)
+	case "mysql", "postgres":
+		s.Parameters["EncryptionKey"] = ""
```
ddollar: Instead of special-casing for these two, we could parse the resource template, get its parameter list, and always set `EncryptionKey` equal to the rack's KMS key if the parameter exists.

SonOfBytes:
@MiguelMoll is there anyway I can help incorporate the feedback from @ddollar?

robinjoseph08:
@MiguelMoll any way we can help out here? I'd love to see this change get merged!

SonOfBytes:
@MiguelMoll is this dead in the water given all efforts are going into praxis?

SonOfBytes:
😢

MiguelMoll:
Apologies @SonOfBytes. You're correct that the work has been focusing on Praxis and trying to get that into shape. Also using the encrypted there for rds resources. I'll check it out again.

ddollar:
Instead of two parameters let's do one of the following:

* One boolean `Encryption` parameter and have the KMS key be created as part of the resource stack.
* One `EncryptionKey` parameter and derive the encryption status from whether or not it's blank.

PR #2115, opened 2017-03-29
customtopic runtime migrations for remaining regions

MiguelMoll:


nzoschke:
This looks correct to me. My understaning:

With the latest update from AWS we expect to be able to update every lambda function in every region, custom topic or not, to `nodejs4.3`. Meaning we can take out *all* the conditional logic around `nodejs` vs `nodejs4.3` as this patch does.

These have been sensitive changes, so we should be extra careful in CI and QA, but big 👍 .

PR #2117, opened 2017-03-30
Quick fixes

### LINKED ISSUE #871: "convox instances keyroll" failure inconsistency
When a `convox instances keyroll` is run while a rack update is already in progress, the command errors out but the key is not rolled back. Because the key is not rolled back, the instances are not accessible via ssh until the next successful keyroll.

### LINKED ISSUE #1035: It should be hard to delete an app
There should be more friction in the process for deleting an app. Right now, there is no confirmation, so the risk of fat fingering something important is high.

The minimum amount of friction (more is welcome) is asking for confirmation by typing the name of an app. Maybe even have the option of deleting apps disabled on the rack level?

MiguelMoll:
Fixes #1035, #2123, #871

### DIFF
--- a/api/models/instances.go
+++ b/api/models/instances.go
@@ -21,7 +21,7 @@ func InstanceKeyroll() error {
 	switch system.Status {
 	case "running", "converging":
 	default:
-		return fmt.Errorf("unable to keyroll with rack status: %s", system.Status)
+		return fmt.Errorf("unable to keyroll rack with current status: %s, status must be running or converging", system.Status)
 	}
 
 	keyname := fmt.Sprintf("%s-keypair-%d", os.Getenv("RACK"), (rand.Intn(8999) + 1000))
--- a/api/models/instances.go
+++ b/api/models/instances.go
@@ -12,6 +12,18 @@ import (
 )
 
 func InstanceKeyroll() error {
+	system, err := Provider().SystemGet()
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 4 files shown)


MiguelMoll:
[Small triage fixes](https://trello.com/c/TzDkvXFy/2261-small-triage-fixes)

api/models/instances.go
```diff
@@ -12,6 +12,18 @@ import (
...
+	system, err := Provider().SystemGet()
+	if err != nil {
+		return err
+	}
+
+	// only allow running and converging status through
+	switch system.Status {
+	case "running", "converging":
+	default:
+		return fmt.Errorf("unable to keyroll with rack status: %s", system.Status)
```
awsmsrc: perhaps we should say which status they CAN keyroll in too?

awsmsrc:
circle is failing

api/models/instances.go
MiguelMoll: Not sure I follow. The only case states the status that a keyroll can take place.

MiguelMoll:
Circle is failing due to AWS throttling during uninstall. This change prints that error for clarity.

api/models/instances.go
MiguelMoll: After reading it 100x, think you mean the error message should state that. Still waking up...

MiguelMoll:
App deletion prompt was canceled out by https://github.com/convox/rack/pull/2129

PR #2033, opened 2017-03-08
Agent lambda fixes

spencerhakim:
- Fixes an error accessing `undefined`
- Listens for task state changes (for when containers exit unexpectedly)
- Cleaned up/improved some of the logic

PR #2118, opened 2017-03-30
convox.agent and elb-udp fix

MiguelMoll:
Carrying #2033 and #2093 forward. Those pesky template merges aren't fun.

@spencerhakim and @nzoschke Would appreciate extra eyes here. Hopefully I didn't miss something while merging/rebasing.


spencerhakim:
Regarding the compiled templates, is there a reason that they're tracked in git? I think it'd solve a lot of problems if they were ignored and just compiled as part of the normal build process.

MiguelMoll:
[convox.agent and elb-udp fix](https://trello.com/c/puwCYPSe/2295-convox-agent-and-elb-udp-fix)

ddollar:
As much as possible I would like `go get github.com/convox/rack/cmd/convox` and friends to work out of the box which as far as I know means we can't have any required pre-processing before compilation.

PR #2101, opened 2017-03-26
Service groups (like kubernetes pods)

ravenac95:
_This is a pretty massive change. I need to add some additional tests, but this is something pretty critical for our use-case and I imagine other company's use cases._

# Service Groups For Convox

## Summary of Changes

* Introduces Service Groups
* Introduces Secure Environment

### Service Groups

This PR introduces service groups for convox. A service group allows you to create something akin to a pod in kubernetes. It allows you to group sets of docker containers that will be deployed together in a single ECS task. Imagine you have a simple `docker-compose.yml` file that looks like this:

```
version: "2"

services:
  one:
    build: .
    ports:
      - 4444:4444
    labels:
      - convox.port.4444.protocol=http
      - convox.health.path=/status
      - convox.health.port=4444

  two:
    build: .
    ports:
      - 80:80
    labels:
      - convox.balancer=false
  
  three: 
    build: .
    ports:
      - 80:80
    labels:
      - convox.balancer=false

  four:
    build: .
    ports:
      - 80:80
    labels:
      - convox.balancer=false
```

In order to group containers `two` and `three` to container `one`, you would add a `convox.group=[group_name]` label to all of the containers, like so:

```
version: "2"

services:
  one:
    build: .
    ports:
      - 4444:4444
    links:
      - two
      - three
    labels:
      - convox.group=main
      - convox.port.4444.protocol=http
      - convox.health.path=/status
      - convox.health.port=4444

  two:
    build: .
    expose:
      - 80
    labels:
      - convox.group=main
      - convox.balancer=false
  
  three: 
    build: .
    expose:
      - 80
    labels:
      - convox.group=main
      - convox.balancer=false

  four:
    build: .
    ports:
      - 80:80
    labels:
      - convox.balancer=false
```

Now `one`, `two`, and `three` are part of the `main` group. Notice in the second `docker-compose.yml` file that `two` and `three` no longer need to bind ports onto the host, they can simply expose ports and be linked to container `one`.

##### Using `convox ps`

The PIDs for containers is now using a two part string split by a `-`. This allows for us to choose a specific container within a specific ECS task. The first part of the string references the taskArn and the second part references the containerArn.

The output of the command looks like this:

```
$ convox ps
ID                         GROUP           NAME   RELEASE      STARTED    COMMAND
8882440bb4c7-75e2d9874eb7  secondarygroup  four   RBPHEISKCSO  1 day ago  node index.js
[...]

PR #2120, opened 2017-03-31
Secure Environment

ravenac95:
# Secure Environment For Containers

## Summary of Changes

* Introduces a secure environment

## Explanation

_The changes were originally part of #2101 but was separated from the Service Groups to reduce feature complexity for a single PR._

This PR introduces a completely new idea to convox that I call the "Secure Environment" option for containers. This change ensures that the environment will NOT appear on AWS's ECS console. This plus #2052 and #2097 will be the holy grail of a secure environment.

A secure environment can be specified for a container in the `docker-compose.yml` file by adding `convox.secure-env=true` to the labels, like so:

```
version: "2"

services:
  secure:
    build: .
    ports:
      - 4444:4444
    labels:
      - convox.secure-env=true
```

Once the secure environment is enabled, 3 environment variables will be passed into the container at runtime. These variables are: 

* `SECURE_ENVIRONMENT_URL` - The url to the s3 file that contains the environment (encrypted with KMS)
* `SECURE_ENVIRONMENT_TYPE` - The type of the file that is encrypted in s3. This is reserved for the future. For now it only returns `envfile`. _We may want to use some actual mimetype (or create one I guess)
* `SECURE_ENVIRONMENT_KEY` - The arn to the KMS key. Used for decryption of the environment file referenced by  `SECURE_ENVIRONMENT_URL`

The intent is that the container will download and decrypt the secure environment upon start up. The advantage for my team is that the environment can still be updated by the convox CLI and the convox console GUI but it will not be available to AWS's console. 

For ease of consumption, I've also built a tool that will download the S3 document that stores the environment variables and will output a bash compatible script to stdout that can be used to source the environment. 

That is available here: [virtru/secure-environment](https://github.com/virtru/secure-environment)
Example of it's use with a docker container is available here: [virtru/secure-environment-example](https://github.com/virtru/secure-environment-example)


nzoschke:
Thanks for splitting this out!

mattmanning:
@ravenac95 Thanks so much for this PR. I'm closing this PR and have created https://github.com/convox/rack/pull/2150 to carry the changes forward.

The major changes from your original PR are:
- changing the label from `convox.secure-env` to `convox.environment.secure`
- automatically creating an IAM Role and Policy to give the ECS Task permission to use the KMS key

Your utility at https://github.com/virtru/secure-environment is great. May we fork it into the convox GitHub org so we can manage it going forward?

PR #2126, opened 2017-04-04
check app and capacity memory

MiguelMoll:


MiguelMoll:
[Autoscale: better handling of apps over-scaled for instance type](https://trello.com/c/yUOCR1Up/2227-autoscale-better-handling-of-apps-over-scaled-for-instance-type)

Issue #2067, opened 2017-03-16
return process ID from run api


contra:
Right now, running a process just returns `{ "success": true }` which isn't very useful. If the API returned the process ID (at the very least) it would be a lot more useful. This is true for most other endpoints, but let's start with this one.

Our use case is programmatically executing a process w/ specified args, and then being able to kill it later if needed. If we don't know the process ID we can't check the status or kill it later.

contra:
Also, if the process failed to start (immediately died with an error) it would be nice if that was reflected immediately in the API, and would also be extremely useful for the CLI.

contra:
This should be pretty easy after looking through stuff:

Will require a one or two line change. The PID is available in the endpoint handler already as `_` https://github.com/convox/rack/blob/master/api/controllers/processes.go#L119

Just needs to be returned on L130 in the response. I could attempt a PR myself but I'm assuming you'll be able to do it much quicker than reviewing my shitty go code.

ddollar:
Duplicate of #1295

PR #2130, opened 2017-04-05
Add process pid to API process/run output

stevelacy:
### Description
- This adds support for returning the process id in the API response
- Related #2067 

~~Is there a better way to return the `success` status as the output status `success` vs `Success`?~~
Just updated, using JSON encoding for the struct parsing, allows non-breaking changes to the output case.
I was extending a struct for the object interface for `RenderJson`

This adds the following response to successful HTTP calls to `/apps/<app>/processes/<process>/run`:
```json
{
  "pid": "<process-pid>",
  "success": true
}
```
### Before Release
- [x] Rebase against master
- [ ] Get review approval
- [ ] Confirm test coverage
- [ ] Submit documentation PR for convox/site

### DIFF
--- a/api/controllers/processes.go
+++ b/api/controllers/processes.go
@@ -116,7 +116,7 @@ func ProcessRunDetached(rw http.ResponseWriter, r *http.Request) *httperr.Error 
 	command := GetForm(r, "command")
 	release := GetForm(r, "release")
 
-	_, err := models.Provider().ProcessRun(app, process, structs.ProcessRunOptions{
+	pid, err := models.Provider().ProcessRun(app, process, structs.ProcessRunOptions{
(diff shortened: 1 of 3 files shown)


contra:
Anyone had time to look at this? I think the only issue is the case on the response, should be more like

`{ "id": "123", "success": true }`

stevelacy:
Just updated, using JSON encoding for the struct parsing, allows non-breaking changes to the output case

stevelacy:
Rebased with master

MiguelMoll:
Moving the PR forward via #2192

stevelacy:
Thanks! 
I was wondering about the inline interface but assumed you'd want a full struct lol

PR #2152, opened 2017-04-14
fix DescribeInstances throttling

ddollar:
this fixes throttling in the presence of large numbers of containers/instances

process-level stats are temporarily disabled

the `Command` attribute of the process will only be set if there is a custom override (`command:` in the manifest)


ddollar:
DescribeInstance was being called 2-3 times for each container (you can see the effect on the test cycles)

PR #2175, opened 2017-04-28
support privileged in convox start

ddollar:


mattmanning:
[`privileged: true` not working locally with `convox start`](https://trello.com/c/Gn9tfVyL/2302-privileged-true-not-working-locally-with-convox-start)

PR #2176, opened 2017-04-28
add param to disable agent

ddollar:


ddollar:
[Disable agent for SMRT](https://trello.com/c/eO9tK7XJ/2384-disable-agent-for-smrt)

PR #2179, opened 2017-05-02
Determine process run order via topological sort

mattmanning:
Build a directed acyclical graph of links and do a topological sort on it to determine process run order for `convox start`.

## Quick CLI Release Playbook
- [x] Code review
- [x] Merge into master
- [ ] Release CLI

PR #2182, opened 2017-05-02
Determine process run order via topological sort

MiguelMoll:
Build a directed acyclical graph of links and do a topological sort on it to determine process run order for convox start.

Reopening #2179


MiguelMoll:
[`convox start` run order](https://trello.com/c/rCZWOi3t/2397-convox-start-run-order)

PR #2185, opened 2017-05-04
Additional HTTP proxy configuration

nzoschke:
In live debugging we observed that the `HTTP_PROXY` variable was not enabling `yum` to fetch updates / packages.

Exporting it as `http_proxy` (lowercase) enabled yum to work.

This contradicts [the docs](https://www.centos.org/docs/5/html/yum/sn-yum-proxy-server.html) that state upper case should work with yum:

> The http_proxy environment variable is also used by curl and other utilities. Although yum itself may use http_proxy in either upper-case or lower-case, curl requires the name of the variable to be in lower-case.

This patch exports the variable in both cases to be extra sure.


provider/aws/dist/rack.json
```diff
@@ -1505,7 +1510,7 @@
               "  - ", { "Ref": "InstanceRunCommand" }, "\n"
               ] ] }
             ] },
-            "  - /opt/aws/bin/cfn-signal --stack ", { "Ref": "AWS::StackName" }, " --region ", { "Ref":"AWS::Region" }, " --resource BuildInstances\n"
+            "  - /opt/aws/bin/cfn-signal --http-proxy", { "Ref": "HttpProxy" }, "--stack ", { "Ref": "AWS::StackName" }, " --region ", { "Ref":"AWS::Region" }, " --resource BuildInstances\n"
```
nzoschke: this lgtm for testing but does it work with no HttpProxy parameter?

nzoschke: needs a space

mattmanning:
Some additional commits have been added to the original PR to more aggressively set proxy configuration and (no-proxy config where necessary). A `--http-proxy` option was also added to `convox install`.

mattmanning:
[Danone](https://trello.com/c/DqbXyYIh/2409-danone)

provider/aws/dist/rack.json
```diff
@@ -1505,7 +1518,7 @@
               "  - ", { "Ref": "InstanceRunCommand" }, "\n"
               ] ] }
             ] },
-            "  - /opt/aws/bin/cfn-signal --stack ", { "Ref": "AWS::StackName" }, " --region ", { "Ref":"AWS::Region" }, " --resource BuildInstances\n"
+            "  - /opt/aws/bin/cfn-signal --http-proxy \"", { "Ref": "HttpProxy" }, "\" --stack ", { "Ref": "AWS::StackName" }, " --region ", { "Ref":"AWS::Region" }, " --resource BuildInstances\n"
```
MiguelMoll: I noticed cfn-signal has an `--https-proxy` option. Is it safe to omit here?

mattmanning: @MiguelMoll I would rather leave it in for now since this is working for our customer. This is a documented option and leaving it in seems to have no negative effects, just a bit of code redundancy.

PR #2187, opened 2017-05-08
only start a service and its dependencies when specified to convox start

mattmanning:
`convox start` will start all services

`convox start <service-foo>` will start only `service-foo` and its dependencies.

This fixes a bug where all services were started even when `convox start <service-foo>`.

### DIFF
--- a/manifest/manifest.go
+++ b/manifest/manifest.go
@@ -265,9 +265,17 @@ func (m *Manifest) getDeps(root, dep string, deps map[string]bool) error {
 
 // Return the Services of this Manifest in the order you should run them
 func (m *Manifest) runOrder(target string) (Services, error) {
+	deps := make(map[string]bool)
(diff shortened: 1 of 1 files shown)


mattmanning:
[fix `convox start foo` bug where all services are started instead of just foo and its dependencies](https://trello.com/c/ZwOTB3a3/2415-fix-convox-start-foo-bug-where-all-services-are-started-instead-of-just-foo-and-its-dependencies)

manifest/manifest.go
```diff
@@ -276,6 +284,7 @@ func (m *Manifest) runOrder(target string) (Services, error) {
 	// Make a map of service names to services
 	serviceMap := make(map[string]Service, 0)
 
+	// Sort the service nams alphabetically
```
MiguelMoll: small names typo

PR #2196, opened 2017-05-15
Mysql Version

mattmanning:
The engine version for MySQL resources was previously hard-coded to 5.6.27, which is not supported in several AWS regions.

This change:
- Changes the default engine version to 5.7.16 which has broader region support
- Adds a `--version` option to `convox resources create mysql`


mattmanning:
[mysql version](https://trello.com/c/Wwe3As58/2448-mysql-version)

PR #2197, opened 2017-05-15
newline after error msg

nzoschke:
The missing newline causes this output on a new rack:

```
$ convox apps
no apps found, try creating one via `convox apps create`Chappie:~ noah (personal/dev)$ convox apps create foo
```

### DIFF
--- a/cmd/convox/apps.go
+++ b/cmd/convox/apps.go
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"github.com/convox/rack/cmd/convox/stdcli"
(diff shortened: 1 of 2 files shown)


cmd/convox/apps.go
```diff
@@ -7,7 +7,6 @@ import (
 	"time"
 
 	"github.com/convox/rack/cmd/convox/stdcli"
-	"gopkg.in/urfave/cli.v1"
```
MiguelMoll: Was this removed by mistake? Perhaps an editor?

PR #2211, opened 2017-05-30
Allow setting of unhealthy threshold via convox.health.unhealthy docker compose label

mattmanning:
This adds two optional Docker Compose configuration labels:

- `convox.health.threshold.healthy`
- `convox.health.threshold.unhealthy`

Heathy Threshold and the Unhealthy Threshold on a service's ELB: http://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-healthchecks.html

Valid values are 2-10. If no label is set the default value of 2 is used.


mattmanning:
[Unhealthy count Health check param](https://trello.com/c/8u5tQBaM/2501-unhealthy-count-health-check-param)

Issue #2210, opened 2017-05-30
Build `failed to register layer: devmapper`


nzoschke:
I have seen a few reports of device mapper errors during a build. We need to:

- [ ] Root cause device mapper errors
- [ ] Fix root cause
- [ ] Or automate instance replacement

For example:

```
...
628128ca0bac: Pull complete
failed to register layer: devmapper: Thin Pool has 10010 free data blocks which is less than minimum required 10126 free data blocks. Create more free space in thin pool or use dm.min_free_space option to change behavior
ERROR: build error: exit status 1
ERROR: myapp build failed
```

mwarkentin:
@nzoschke That sounds similar to the issue we saw when our custom AMI was configured with the base 20GB volume instead of the 50GB volume provided by convox. Might be worth checking if those reports are also using custom AMIs or not.

Issue #1272, opened 2016-09-28
"not enough memory available to start process" during deploy


contra:
<img width="1276" alt="screen shot 2016-09-28 at 5 02 30 pm" src="https://cloud.githubusercontent.com/assets/425716/18932210/657b125c-859d-11e6-809f-b156d0dd2bac.png">

<img width="460" alt="screen shot 2016-09-28 at 5 03 03 pm" src="https://cloud.githubusercontent.com/assets/425716/18932232/749f47f8-859d-11e6-9e4d-2f0f65ca5241.png">

ddollar:
That means you don't have enough available space on your cluster to run a build (which may also explain some of the other issues you have been seeing)

What is the type/count your Rack is scaled to?

contra:
@ddollar Last screenshot has that info

ddollar:
Sorry about that. You can try `convox rack params set BuildMemory=NNN` where NNN is a number in MB. The default is 1024.

contra:
@ddollar Updated from a t2.micro to a t2.small, which has 2GB of RAM (should be more than enough for the 1GB default build memory) - now getting a hang for 10+ minutes, then CI kills it.

Shouldn't the build memory be based on the size of the instance, with a max of 1GB? Otherwise t2.nano and t2.small will always fail to deploy.

ddollar:
Probably, but trying to keep this already-complicated release as small as possible. The default size works well with the default rack instance type. Planning to update Console to help pick most of these values automatically.

ddollar:
Closing this for now as you should be able to work around this error with the command pasted above.

contra:
@ddollar If the problem is valid and something you want to fix at some point shouldn't the issue stay open? Otherwise I feel like this is going to get lost in the void for good.

contra:
Still happening even on a t2.small (2GB RAM, way more than what should be needed). Any ideas?

<img width="1273" alt="screen shot 2016-09-28 at 10 21 41 pm" src="https://cloud.githubusercontent.com/assets/425716/18938879/fb8e836a-85c9-11e6-9197-4b06ba2dbe6b.png">

contra:
Can I get this reopened since this is still a bug/regression?

contra:
Ping - I haven't been able to deploy for two weeks because of this bug. Can this please be reopened?

nadurthi:
I am also experiencing the same problem.

contra:
[...]

PR #2212, opened 2017-05-31
Build Options

nzoschke:
This adds new ways to tune builds:

- BuildVolumeSize parameter that sets a build instance volume independently of the main cluster
- Build tasks get a soft memory reservation

A way to use this together is:

- BuildInstance=m4.2xlarge
- BuildMemory=256
- BuildVolumeSize=500

This instance has 32 GB of memory, so it will support 128 concurrent builds, each one taking as much memory as it can. The build volume is 10X bigger than default which will greatly reduce the chance of the disk filling up.

This should fix https://github.com/convox/rack/issues/2210 (docker build errors when device mapper is full), and address https://github.com/convox/rack/issues/1272 for build processes (memory reservations blocking new build processes).

### DIFF
--- a/provider/aws/dist/rack.json
+++ b/provider/aws/dist/rack.json
@@ -294,6 +294,11 @@
       "Description": "How much memory should be reserved by the builder",
       "Default": "1024"
     },
+    "BuildVolumeSize": {
+      "Type": "Number",
+      "Description": "Default build disk size in GB",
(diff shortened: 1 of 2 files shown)


nzoschke:
[Build Instance Soft Memory Limit](https://trello.com/c/hWwWylWx/2503-build-instance-soft-memory-limit)

PR #2217, opened 2017-06-04
Reliable build export

MiguelMoll:
During a build export, docker commands in the backend can take some time to complete so we parallelize them and write the response as soon as possible.

Also added some headers and haproxy options so client/servers aren't assuming behavior that caused the connection to close mid-export.


MiguelMoll:
[use waybetter test case to fix build copy](https://trello.com/c/v4RroVsx/2486-use-waybetter-test-case-to-fix-build-copy)

PR #2226, opened 2017-06-16
Apply TaskRole to run TaskDefinition

mattmanning:
If there is a TaskRole set in the app params this will make sure it gets set on `convox run` tasks as well.

### DIFF
--- a/provider/aws/processes.go
+++ b/provider/aws/processes.go
@@ -830,9 +830,12 @@ func (p *AWSProvider) generateTaskDefinition(app, process, release string) (*ecs
 		})
 	}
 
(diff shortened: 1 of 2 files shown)


mattmanning:
[convox run taskrole](https://trello.com/c/EPDBiI2v/2537-convox-run-taskrole)

MiguelMoll:
This looks good to me. Any possible side effects for passing an empty string as a role arn if one is not used?

mattmanning:
@MiguelMoll Just double-checked and the empty string seems to work just fine. App updated fine and I was able to do attached and detached runs without any issues.

PR #2230, opened 2017-06-20
take brackets into account for variable expansion

MiguelMoll:


MiguelMoll:
[interpolation in docker-compose command sequence](https://trello.com/c/iF485yk7/2562-interpolation-in-docker-compose-command-sequence)

PR #2237, opened 2017-06-22
faster code sync

swerling:
- removes 1 code-sync delay and reduces another, causing faster sync

- reduces idle CPU usage by the convox process. Anecdotally, a reduction from ~2.5% CPU  down to ~0.2% was seen.

- better handles file editors that create a 'flurry' of file system events in quick succession during file save, such as vim on linux (in nocompat mode)

- provides a 'fallback code sync' for exceptional setup where file systems events get lost from time to time. Eg. running on linux  over an NFS or fuse-mounted file system.

Turn fallback code sync on by starting convox with the environment var FALLBACK_SYNC_TICK set to some millisecond value, eg.

    FALLBACK_SYNC_TICK=750

This means that every 750 milliseconds, a project subdirectory is rescanned for file changes iff that directory is 'hot'. A directory is 'hot' if some file system event (save,create) has occurred any time in the last 10 minutes.


mattmanning:
[Faster code sync](https://trello.com/c/rHg0Awwq/2602-faster-code-sync)

mwarkentin:
@swerling @mattmanning We're seeing some CPU usage issues with some of our larger apps as developers migrate to them - wondering if this will help. Is there a binary we could test out?

changes/changes.go
```diff
@@ -93,56 +97,92 @@ func readDockerIgnore(file string) ([]string, error) {
...
+
 	}
 
 	return nil
 }
 
-func notify(ch chan Change, from, to map[string]time.Time, base string, ignore []string) {
+// Do a code-sync (notify), then compare the dir contents to the contents from
+// before the sync. Repeat until dir contents dont change for at least 200ms
+// Give up on stabilizing dir after 100 syncs. Also see func comment for
```
MiguelMoll: I believe the comment should say after 10 syncs?

MiguelMoll:
Thanks for the PR. Really good improvements. While testing it out, I did notice on OSX I was getting `fatal error: concurrent map writes` panics fairly consistent and some processes would get sync changes while others wouldn't. I'll take a look at these errors and see if I can knock them out.

Thanks again!

swerling:
> Thanks for the PR. Really good improvements. While testing it out, I did notice on OSX I was getting fatal error: concurrent map writes panics fairly consistent and some processes would get sync changes while others wouldn't

That sounds not so great. Wish I could reproduce that on mine.

swerling:
this is stale, out of sync w/ master

PR #2239, opened 2017-06-27
don't generate random password if one has been provided

mattmanning:
Fixes a bug where the `--password` parameter to `convox resources create` and `convox resources update` was being ignored.


mattmanning:
[--pasword flag for postgres resources](https://trello.com/c/WsSDnuaG/2563-pasword-flag-for-postgres-resources)

PR #2242, opened 2017-06-28
`certs generate` should take minimum 1 argument, not exactly 1

mattmanning:


mattmanning:
[Fix `convox certs generate` with multiple domains](https://trello.com/c/kw5mjlzV/2604-fix-convox-certs-generate-with-multiple-domains)

PR #2247, opened 2017-07-06
Fix Swallowed Errors

alrs:
I found some places in rack where error variables are being collected but not used.

### DIFF
--- a/api/cmd/build/clean.go
+++ b/api/cmd/build/clean.go
@@ -58,6 +58,7 @@ func clean() {
 	records, err := r.ReadAll()
 	if err != nil {
 		fmt.Printf("WARNING: %s\n", err)
+		return
 	}
 	for _, r := range records {
 		// Filter out non-build images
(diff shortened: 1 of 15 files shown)


api/cmd/build/clean.go
```diff
@@ -56,6 +56,9 @@ func clean() {
 	r.FieldsPerRecord = 4
 
 	records, err := r.ReadAll()
+	if err != nil {
+		fmt.Printf("WARNING: %s\n", err)
```
MiguelMoll: There should be a return after the warning so the code doesn't continue in an unknown state.

MiguelMoll:
This is a great PR. Thanks for this. 👍

api/cmd/build/clean.go
alrs: `return` added.

MiguelMoll:
[Fix Swallowed Errors](https://trello.com/c/N1cnpIOS/2667-fix-swallowed-errors)

PR #2252, opened 2017-07-13
Spot Instances Prototype

mattmanning:
This adds an additional LaunchConfiguration and Autoscaling Group to the Rack template to allow the addition of EC2 spot instances to the ECS Cluster.

The following 3 Rack parameters are also introduced:
- `SpotInstanceBid`
- `SpotInstanceCount`
- `SpotInstanceType`

When using spot instances Rack autoscaling should be turned off with `convox rack params set Autoscale=No`. The autoscaler is not yet aware of the spot instances, so it would scale up more on-demand instances than necessary. For now, both on-demand and spot instance counts should be scaled manually.

### DIFF
--- a/api/cmd/monitor/main.go
+++ b/api/cmd/monitor/main.go
@@ -11,6 +11,7 @@ func main() {
 	go workers.StartCluster()
 	go workers.StartHeartbeat()
 	go workers.StartEventQueue()
+	go workers.StartSpotReplace()
 
 	for {
 		time.Sleep(1 * time.Hour)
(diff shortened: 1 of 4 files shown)


mattmanning:
[Doordash cost reduction](https://trello.com/c/xwePshIP/2603-doordash-cost-reduction)

mattmanning:
[Rack spot instances](https://trello.com/c/MaE0xEeL/2680-rack-spot-instances)

mattmanning:
Closing in favor of https://github.com/convox/rack/pull/2265

PR #2257, opened 2017-07-17
dont pass rm flag on run therefore preserve container artifacts

cleblanc87:
When we run tests in CI, the underlying exit code from our test container is not surfaced.
We are able to capture the exit code after the run via scripting, however the current --rm flag prevents the build artifacts from sticking around long enough to inspect.

An alternative is implementing the `--exit-code-from` feature during `convox start` which would be great


MiguelMoll:
I believe we'd like to keep the `--rm` argument as to not have lingering containers on development machines that use `convox start`.

I'd like to hear more about the testing strategy you're using so that we can come up with a solution to help you out. You're running `convox start` in a CI to test the app?

cleblanc87:
@MiguelMoll that is how we are using it. Once convox exits we are able to use the docker cli to determine the exit code of our test container. Possible to make this flag optional? Something like `--no-rm`?

MiguelMoll:
@cleblanc87 Apologies for the delayed response. Are you running tests via `convox start` for each process of an app?

ddollar:
Closing for now, will implement exit code capture in the refactor currently underway

PR #2259, opened 2017-07-18
Default to 200GB disk instead of 50GB disk on dedicated build instances

mattmanning:
This should reduce the frequency of out-of-disk errors for builds on dedicated build instances.


mattmanning:
[Increase BuildVolumeSize default](https://trello.com/c/dX0VWpHz/2691-increase-buildvolumesize-default)

PR #2265, opened 2017-07-21
Spot Instances

mattmanning:
If you set a USD value for the `SpotInstanceBid` rack parameter, Rack will try to run spot instances instead of On-Demand instances in your ECS cluster.

The `OnDemandMinCount` parameter can be set to configure a minimum number of on-demand instances that won't be replaced with spots. Its default value is 3.


mattmanning:
[Rack spot instances](https://trello.com/c/MaE0xEeL/2680-rack-spot-instances)

PR #2268, opened 2017-07-25
Avoid inline Lambda limit

mattmanning:
This change updates the CloudWatch Rule for a cron job to pass the ECS Task Definition ARN instead of having all the ARNs listed in a lookup table in the Lambda function. This reduces the likelihood that a CloudFormation update will fail for an app with lots of services because of the inline Lambda function exceeding 4096 characters.


mattmanning:
[Avoid character limit for cron Lambda functions](https://trello.com/c/iI5cjuI1/2707-avoid-character-limit-for-cron-lambda-functions)

PR #2271, opened 2017-07-31
Add rack instance security group param

william-richard:
We would like to start using convox, but would like to have tight control over the security groups of the ecs instances.  This change allows you to override the default security group, with one managed outside of convox's stack.

I have tested it with one of our internal apps - the only issue I ran into is you need to explicitly allow traffic from the convox app within the security group.  As long as you have those rules, it works well.


mattmanning:
[Security Group customization](https://trello.com/c/HoAFC23D/2727-security-group-customization)

MiguelMoll:
@william-richard Could you rebase on master so we could move this PR forward. Thanks!

william-richard:
Hi @MiguelMoll - I just rebased.  Let me know if there's anything else I can do 😄

PR #2274, opened 2017-08-04
use gorilla/websocket and periodic ping for keep alive

ddollar:


client/client.go
```diff
@@ -1,23 +1,22 @@
...
-	"net"
 	"net/http"
 	"net/url"
 	"os"
 	"strings"
 	"sync"
 	"time"
 
-	"golang.org/x/net/websocket"
+	// "golang.org/x/net/websocket"
```
MiguelMoll: remove comment?

PR #2276, opened 2017-08-08
Support a minor version number for v2 Docker Compose config files

jamesottaway:
I'm trying to use [`docker-compose config`](https://docs.docker.com/compose/reference/config/) to pre-process a Compose file template (`docker-compose.convox.yml`) and generate the Compose file (`docker-compose.<rack>.yml`) that I will pass to `convox build`, but it spits out the version as "2.0" which the builder doesn't seem to like.

```
❯ cat docker-compose.convox.yml | grep version
version: '2'

❯ docker-compose -f docker-compose.convox.yml config > docker-compose.staging.yml

❯ cat docker-compose.staging.yml | grep version
version: '2.0'

❯ convox build --app test -f docker-compose.staging.yml
Creating tarball... OK
Uploading: 3.93 MB / 3.93 MB [==========================] 100.00 % 48s
Starting build... OK
Authenticating nnnnnnnnnnnn.dkr.ecr.us-east-1.amazonaws.com: Login Succeeded
Authenticating https://index.docker.io/v1/: Login Succeeded

ERROR: unknown manifest version: 2.0
ERROR: test build failed
```

If you'd rather not explicitly support v2.1 I'm happy to take that out.

### DIFF
--- a/manifest/fixtures/v2-0.yml
+++ b/manifest/fixtures/v2-0.yml
@@ -1,0 +1,4 @@
+version: "2.0"
+services:
+  web:
+    image: test
--- a/manifest/fixtures/v2-1.yml
+++ b/manifest/fixtures/v2-1.yml
@@ -1,0 +1,4 @@
+version: 2.1
+services:
+  web:
+    image: test
--- a/manifest/fixtures/v2-2.yml
+++ b/manifest/fixtures/v2-2.yml
@@ -1,0 +1,4 @@
+version: '2.2'
+services:
+  web:
+    image: test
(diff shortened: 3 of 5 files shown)


jamesottaway:
I also just noticed that v2.2 came out without me noticing, so I've updated this to support `version: '2.2'`.

As with v2.1, let me know if this isn't desired, although I hope it can be left here for anyone who wants to use those newer versions in their base Compose config files, even if there's no expectation that `convox build` will respect options added after v2.0.

PR #2279, opened 2017-08-15
rds encryption key

MiguelMoll:
Moving https://github.com/convox/rack/pull/2110 along as the branch is too far behind. Also using one parameter via `EncryptionKey`.


jaskinn:
This would be very convenient if it made it into a release.

ddollar:
Closing for now as the design needs work

PR #2286, opened 2017-08-25
set ECS instance attributes on build, primary and spot instances

nzoschke:
This fixes a potential problem by guaranteeing the `spotreplace` worker runs on an on-demand instance.

I've found some questionable things while testing. My parameters are:

```
  AutoScale=No
  InstanceType=m3.medium
  InstanceCount=6
  OnDemandMinCount=3
  SpotInstanceBid=0.0108
```

If I update InstanceCount to 7 I see:

- An initial period of 14 instances because both on demand and spot ASG set desired to instance count (both 7)
- Then spot instances are removed incrementally first. I saw spot=5 and on-demand=7 for example.
- Then some on-demand instances are removed. I'm at spot=4 and on-demand=6.
- Then it keeps changing things for quite a while until it settles down. Approximately 15 minutes
- Finally it stabilizes at spot=4 and on-demand=3.

This doesn't seem ideal to me. I would expect:

- Spot ASG immediately set to 4 (instance count 7 less min on demand 3)
- On Demand ASG adjusted if spots aren't fulfilled or ECS instances are draining
- On Demand ASG stabilizing in < 5m

Why does the Spot ASG Desired Count change at all? Can we set that to the ideal number (InstanceCount - MinOnDemand) and let AWS automatically add and remove spots as bids are won or lost?

This could greatly simplify `spotreplace.go` to something like:

- Update the Spot ASG Desired count only once if not already the ideal number of spots
- Update the OnDemand Desired count based on active spots and draining instances


ddollar:
Wouldn't the monitor worker just be restarted on a different instance?

PR #2287, opened 2017-08-28
tighten up spot replacement algorithm

ddollar:


mattmanning:
@ddollar What happens here if spot instances can't be acquired for the set bid price? (Or are terminated) Do we need some logic to backfill capacity with on-demand instances in that case?

mattmanning:
Oh I see. It looks like the calculation of `onDemandDesired` takes currently running spot count into account. Please disregard my previous question.

ddollar:
What would cause app downtime when this is enabled?

PR #2288, opened 2017-09-12
ALB support

ddollar:
This adds support for ALB to convox v1.

**WARNING: Changing an app in place using this method will cause downtime. The ALB will be assigned a hostname different from the ELB currently in use by your app. To migrate to ALB without downtime configure a second app using this method and then switch your DNS**

* Update your Rack to a release containing this PR
* Modify your `docker-compose.yml` to reference only one port pair for a service and redeploy, eg `443:3000`
* Switch to ALB using the new parameter:`convox apps params set BalancerWebType=ALB`  (use your service name in place of `Web`): 
* The ALB will ignore the first half of your service's port pair and will always listen on port 80. It will also listen on port 443 if you have set a certificate for that port with `convox ssl update`
* After the update is complete your service will be listening on a new hostname. Retrieve it with `convox apps info`


mwarkentin:
@ddollar any chance of cutting a test release for this? Would love to play around with it on our dev racks.

mwarkentin:
@ddollar does this work for internal services as well?

hnrc:
Any updates on this? Is the plan to merge ALB support into `master`?

jaskinn:
I really hope that this can make it into a release in the near future.

mwarkentin:
@ddollar how do we configure routing on this? Is there a way to specify host/path-based routing?

ddollar:
With ALB now supporting multiple certificates this branch is a bit outdated. Work continues on the `refactor` branch

whatyoubendoing:
what happened to this?

mwarkentin:
@whatyoubendoing ALB support has landed in the new 'gen2' applications.

whatyoubendoing:
@mwarkentin to use multiple ssl certs, you have to use gen2 apps? is there a guide of migrating gen 1 to gen 2

whatyoubendoing:
@mwarkentin Its evident based on the docs and chat gen2 apps arent ready for production @ddollar is there anyway we can can use ALB with gen1 apps?

ddollar:
Gen2 apps are ready for production and that is the path to get ALB.

PR #2310, opened 2017-10-07
Add support for secure environment in oneoff tasks

### LINKED ISSUE #2304: [SECURITY] Secure environment ignored for tasks started with `convox run`
I have an app with `convox.environment.secure=true` label set. Service tasks in ECS get the correct set of environment variables (`SECURE_ENVIRONMENT_KEY`, `SECURE_ENVIRONMENT_TYPE`, etc).

However, if a task is executed with `convox run` the corresponding task definition that gets created contains the environment in plain text and the container receives the environment as if `convox.evironment.secure` would have been set to `false`.

Happens on latest rack release `20171002150509`

bladealslayer:
Resolves #2304.

### DIFF
--- a/provider/aws/processes.go
+++ b/provider/aws/processes.go
@@ -721,13 +721,21 @@ func (p *AWSProvider) generateTaskDefinition(app, process, release string) (*ecs
 	sarn := ""
 	sn := fmt.Sprintf("Service%s", upperName(process))
 
+	secureEnvRoleName := ""
+
 	for _, r := range rs.StackResources {
 		if *r.LogicalResourceId == sn {
 			sarn = *r.PhysicalResourceId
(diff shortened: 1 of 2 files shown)


provider/aws/processes.go
```diff
@@ -834,8 +848,14 @@ func (p *AWSProvider) generateTaskDefinition(app, process, release string) (*ecs
 		})
 	}
 
+	cd.Privileged = &s.Privileged
```
ddollar: Make this `cd.Privileged = aws.Boolean(s.Privileged)` rather than a direct reference to avoid possible side effects.

bladealslayer: @ddollar done

PR #2313, opened 2017-10-12
Add extra prompt before deleting resources

jontonsoup:
Add a second prompt so that accidental database deletions happen less frequently. 

I tried running the tests locally but didn't want to set up AWS. If travis fails, I will update the PR. 

UPDATE: tests pass :)

### DIFF
--- a/cmd/convox/resources_test.go
+++ b/cmd/convox/resources_test.go
@@ -127,7 +127,17 @@ func TestResourcesDelete(t *testing.T) {
 		test.ExecRun{
 			Command: "convox resources delete syslog-1234",
 			Exit:    0,
-			Stdout:  "Deleting syslog-1234... DELETING\n",
+			Stdout:  "Are you sure you want to delete syslog-1234? (N/y): Deleting syslog-1234... DELETING\n",
+			Stdin: "y",
+		},
+	)
+
+	test.Runs(t,
+		test.ExecRun{
+			Command: "convox resources delete syslog-1234",
+			Exit:    1,
+			Stdout:  "Are you sure you want to delete syslog-1234? (N/y): ",
+			Stdin: "n",
 		},
 	)
 }
--- a/cmd/convox/services.go
+++ b/cmd/convox/services.go
@@ -13,6 +13,8 @@ import (
 	"github.com/convox/rack/cmd/convox/stdcli"
 	"gopkg.in/urfave/cli.v1"
 	"time"
+	"bufio"
+	"os"
 )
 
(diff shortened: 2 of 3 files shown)


ddollar:
Most of the UX research I've read in the past suggests that adding these sorts of y/N responses almost always leads to training users to blindly hit y after the commands.

More effective would me to make them type the name of the thing they are trying to delete, but they already had to do that in the original command. That being said, any sort of prompt added to this command would break any existing scripts using it.

You could possibly print out the name of the thing they are trying to delete and give a 5 second delay? I don't really love that either.

What is the core problem we are trying to solve? Is it too easy to type the wrong name for deletion? Are users too quick on the enter key?

jontonsoup:
Yes -- retyping the name probably makes more sense. 

The problem I'm trying to solve is this: 

1. `convox resources`
2. Get massive list of things: In our case we have 10 currently and will have over 40 when we are done migrating. 
3. Copy / Paste something in the list into `convox resources delete thing`
4. Realize you have copy / pasted the wrong thing and now have deleted a production database

I'm not sure if this is a high or low likelihood event, but event if it happens once in production its a disaster. 

I'm also not sure if a final snapshot is taken before the delete. So I'm not sure if this is just a downtime mistake or a downtime mistake plus a data loss event. 

Thanks!

ddollar:
Hmm, so if you are pasting you are likely to just paste again into a confirmation dialog. If the user is not stopping to read the command before they push enter how do we force them to read a confirmation dialog?

jontonsoup:
We don't. Ultimately, we can't force someone who is careless to be less careless. 

I think the thing that is bugging me here the most is that most destructive action UIs I've seen have some sort of double opt in. This is a single opt in. 

The big exception here is `rm`. That said, its not hard to find war stories about `rm super-important-file`

The real blocker here is not breaking scripts that depend on this already, but it makes me really uneasy that anyone on my team can accidentally bring down a production DB with one command.

ddollar:
What about adding some sort of “locked” mode that prevents a database from being deleted. Requires manual unlock first. I think we can do this pretty easily on cloud formation with termination protection.

jontonsoup:
That would be awesome! 

Cloud formation is my weakest area right now, but I am happy to learn it for this PR. Do you mind pointing me to where I need to look in the code base and the associated tests? 

Also do you mind writing out the cli incantation you would like to see for this (`convox resources lock XXXXX`)?

ddollar:
Going to close this for now but have added an internal feature request to our list to support CloudFormation termination protection

PR #2316, opened 2017-10-19
add AppCookieStickiness for load balancer

kevinhuang12345678:
From HavenLife


ddollar:
What's the downside of just turning this on by default rather than making it configurable?

mwarkentin:
Some cons mentioned here: https://stackoverflow.com/questions/1553645/pros-and-cons-of-sticky-session-session-affinity-load-blancing-strategy

* depending on "sticky" implementation in your load balancer, may direct unequal load to some servers vs. others
* bringing a new server online doesn't immediately give the new server lots of load-- if you have a dynamic load-balancing system to deal with spikes, stickiness may slow your ability to respond quickly to a spike. That said, this is somewhat of a corner case and really only applies to very large and sophisticated sites.
* if you have relatively few users but a single user's traffic can swamp one server (e.g. complex pages with SSL, AJAX, dynamically-generated images, dynamic compression, etc.), then stickines may hurt end-user response time since you're not spreading a single user's load evenly across servers. If you have a lot of concurrent users, this is a non-issue since all your servers will be swamped!

ddollar:
OK, sounds like it probably shouldn't be on by default.

The config value in this PR doesn't really match the others style-wise (all lowercase, not mixed-case). Making the cookie name configurable doesn't seem to add much.

How about

`convox.port.NNN.sticky=true`

kevinhuang12345678:
can we merge this poll request? thanks.

-Kevin Huang
HaveLife

ddollar:
Hi Kevin,

Please take a look at the most recent comments on the PR and let me know if you have any questions.

kevinhuang12345678:
checked in code to follow the name convention:

convox.port.sticky.cookie=someCookieName

Thanks.

Regards,
Kevin Huang
HaveLife

ddollar:
It seems like there is no advantage to making the cookie name configurable, it's just another decision the user will need to make.

kevinhuang12345678:
In AWS, when "Enable application generated cookie stickiness", "Cookie Name" is required.
Thanks.

Regards,
Kevin

ddollar:
How about just using `convox` as the cookie name.

swisslala:
@ddollar if we default (or hardcode) convox as the name then we have to always override in the app code. For example Express or any other framework that uses connect inderneath would have to be aware of this since the default session id is 'connect.sid' in the headers. I'd advocate to make it free-form like aws allows.

ddollar:
So thinking about that even more what is the advantage of using this app stickiness cookie instead of a load balancer-level cookie?

swisslala:
If your app is managing sessions and controls the timeout then the ELB session management won't work.

ddollar:
But the worst that would happen is that they end up on the same container as before? What's the downside?

swisslala:
@ddollar how so? If you are running multiple instances then they end up in the next round robin selected container. If your app is managing sessions and timeout then you may end up with ghosting issues especially when you sync that session to redis for example. It's basically telling the ELB to delegate session management. Otherwise people are forced to use other tools to fill in the gaps (which we are currently doing...)

ddollar:
Closing this for now as it adds complexity and is not even possible in Gen 2 / ALB

PR #2322, opened 2017-10-21
Updating Lambda permissions required for latest build

denmat:
Latest (https://github.com/convox/rack/releases/tag/20171021144122)  release required the following permmisions for it to succeed.

```
User: arn:aws:iam::<aws-account>:user/convox/content-loadingbay-staging-KernelUser-1PQNW1YM1FF65 is not authorized to perform: lambda:UpdateFunctionCode on resource: arn:aws:lambda:us-east-1:<aws-account>:function:content-loadingbay-stagin-InstancesLifecycleHandle-37WO1FQWQIVG
```
```
AccessDenied. User doesn't have permission to call lambda:GetFunctionConfiguration
```


ddollar:
Hmm. It doesn’t make sense to me why this release would need that new permission. The lambda gets updated with every release.

PR #2333, opened 2017-10-30
Generation 2

ddollar:
This release adds support for Generation 2 applications. Some new features:

* One ALB per Rack, serving all apps on that Rack
* Role-based IAM permissions replace the global Rack IAM user
* Simplified configuration with `convox.yml`
* Rack autoscaler that supports multiple containers from the same service on each instance
* Environment variables now remain in S3/KMS and are loaded by the container at runtime

The goal is to make your Rack faster, simpler, cheaper, and more secure.

You can try it out by creating a new app with `--generation 2`


Dockerfile
```diff
@@ -1,17 +1,6 @@
...
-    git \
-    haproxy \
-    openssh \
-    openssl \
-    python \
-    tar
-
-RUN curl -fsSLO https://get.docker.com/builds/Linux/x86_64/docker-1.13.0.tgz \
-    && tar --strip-components=1 -xvzf docker-1.13.0.tgz -C /usr/local/bin
+RUN apk update && apk add build-base curl docker git haproxy openssh openssl python tar
```
mwarkentin: You can just do `apk add --update build-base ...`.

ddollar: Thanks :) Updated

PR #2338, opened 2017-11-07
alb and s3 log to s3 bucket

ddollar:
configurable with the `LogBucket` rack parameter


mwarkentin:
@ddollar Any chance this is going to be landing soon? Looks like some tests need to have their expected responses updated?

ddollar:
Updated and running through CI now

mwarkentin:
@ddollar Thanks!

Issue #2332, opened 2017-10-26
Add install option for customized cloudformation URL


ffjia:
Hello,

This is a feature request, it's good if you guys could allow customized cloudformation template URL when running `convox install`.

Furthermore, is it possible to use this option to retrieve all related files (CFN tmpl, lambda functions, versions.json) from our own S3 bucket, instead convox.s3.amazonaws.com?

ddollar:
You can do this with the Amazon web ui, CLI, or API for now. More detailed information added in #2343

PR #2343, opened 2017-11-14
add s3-template-url.

ffjia:
Support customized CFN template in user's S3 Bucket, `convox install --stack-name myrack --s3-template-url https://bucket.s3.amazonaws.com/release/20171113014715/formation.json`.

### DIFF
--- a/cmd/convox/install.go
+++ b/cmd/convox/install.go
@@ -162,6 +162,11 @@ func init() {
 				Value: defaultSubnetCIDRs,
 				Usage: "subnet CIDRs",
 			},
+			cli.StringFlag{
+				Name:   "s3-template-url",
+				Value:  "",
+				Usage:  "specify an amazon s3 template url for rack",
+			},
 		},
 	})
 }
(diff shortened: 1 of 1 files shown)


ffjia:
This is for issue #2332.

ddollar:
This is a good start but will only affect the initial Rack installation. Upgrades past that point would still come from the Convox S3 bucket.

ffjia:
Yep, it won't work when using `convox update`. But as a workaround, you could update the CloudFormation directly, with S3 template url enabled.

I found you guys defined fixed template url of Convox S3 bucket in other places, e.g. in `provider/aws/system.go`, do you think it might be possible to get the URL from convox CLI as well? Hints? Thanksl

ddollar:
Closing this for now. If you want to install Rack from a custom template in an S3 bucket you can do so using the AWS console, CLI, or API. Making all of the convox assets come from an S3 bucket could be interesting but would be a much bigger undertaking.

PR #2344, opened 2017-11-15
upgrade docker version to support multistage builds

ddollar:


mwarkentin:
Will this enable docker 17.09 features for all builds (like `--chown`)? I thought we determined that was related to the docker version on the host, but maybe I'm mistaken (hopefully).

PR #2373, opened 2017-11-29
wip: fargate support

ddollar:


ddollar:
waiting for fargate to be released in more regions

PR #2375, opened 2017-11-30
make rack api memory a soft limit

ddollar:


radekl:
What was the reason for that change? Currently we see issues according to the `api` process memory and I wonder if setting it as a hard memory isn't helping a bit with that.

PR #2396, opened 2017-12-06
support internal dns at .convox

ddollar:
Creates internal DNS that routes inside the Rack:

`rack.convox` - points to the Rack API
`<service>.<app>.convox` points to a given service endpoint

App-level endpoints only work for Generation 2 apps.

### DIFF
--- a/cmd/convox/install.go
+++ b/cmd/convox/install.go
@@ -632,6 +632,8 @@ func FriendlyName(t string) string {
 		return "NAT Elastic IP"
 	case "AWS::EC2::InternetGateway":
 		return "VPC Internet Gateway"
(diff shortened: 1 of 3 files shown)


jontonsoup:
@ddollar  what is the expected behavior of this when you have two racks inside of the same VPC? 

I am getting this error when I update my second rack from 20171201042926 to version 20171206173224: 

--
  | 15:11:35 UTC-0500 | CREATE_FAILED | AWS::Route53::HostedZone | HostedZone | The  VPC that you chose, vpc-1fb64667 in region us-east-1, is already  associated with another private hosted zone that has an overlapping name  space, convox..

PR #2407, opened 2017-12-11
Fix for new environment variables set before new build.

radekl:
During creation of new environment variable, when application release exists convox is using existing one as a base.

The issue is that after using the existing one `Id` and `Env` fields are replaced, though `Created` is the same. Later when we are trying to create new build, it is not fetching the release info from stack (great!) though, it is sorting by `app_name.created_time` index - which in this case will fail, because `Created` field for new environment variables is the same as the active release. This causes new environment variables not being visible by new builds. (https://github.com/convox/rack/blob/master/provider/aws/builds.go#L625)

This affects only `build` (`convox build ~`) - not affecting `build import` command as the release creation is not overwritten by the latest application release.

### DIFF
--- a/provider/aws/environment.go
+++ b/provider/aws/environment.go
@@ -2,6 +2,7 @@ package aws
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/convox/rack/crypt"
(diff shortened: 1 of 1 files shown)


radekl:
I was thinking about inserting this set time line into if condition, because it may belong more there than where I put it, though the code looks better this way, so it is seen from first look that we want a current time in this variable.

ddollar:
Thanks!

PR #2436, opened 2018-01-05
Add Build.Manifest to BuildHash

idris:
Two builds with the same path and different manifests (Dockerfiles)
are not equivalent and should not share a hash.

This fixes a bug with https://github.com/convox/rack/pull/2415


ddollar:
Looks good, thanks

Issue #2430, opened 2018-01-03
[GEN2] Suggestions related to domains


ffjia:
Hello,

I found two small issues related to custom domains w/ gen2 application.

1. Need to approve SSL certificate request every time new application deployed. For example, if the custom domain is `myapp.mydomain.tld`, I have to approve this request. Next time if I use `mynewapp.mydomain.tld`, I need to do it again whether I really need the certificate or not.

Is it possible to let user use a wildcard SSL cert like `*.mydomain.tld`? Or add an option to skip the certificate creation proccess?

2. For ALB rules, it looks like Convox will add a rule which uses xxxxxxxxxxxxxx.convox.site domain, even if I specify custom domain in `convox.yml`. The ALB rules limit is 100, which means one ALB can only support less 50 applications. Could you guys help skip the (xxxxxxxxxxxxxx.convox.site route) rules if custom domain(s) is used?

ddollar:
Issue 1 has been covered by a new issue #2470 

Issue 2 has a PR #2451 with some comments

PR #2451, opened 2018-01-16
skip creating default ALB listener rules when having `domains` setup.

ffjia:
Try to fix 2nd issue in  #2430.


ddollar:
This would need to be parameterized at the app stack level as the `convox.site` domain is still useful for testing purposes in many apps.

ffjia:
IIUC, skip creating ALB listener rules won't affect other apps, right? Users specify `domain` in `convox.yml` normally means they'd like to use that `domain(s)` to route the traffic though ALB.

mwisner:
Just my two cents...

I think the default provided domains are a useful tool, especially when first setting up the services when it comes to generally debugging the application.

If I'm understanding correctly... I think @ddollar  is suggesting that instead of just removing the convox.site domain when you have a domain, doing something like this:

```yaml
services:
  web:
    build: .
    command: bin/web
    domain: ${HOST}
    disable_default_domain: true
```

ffjia:
I see your point, (correct me if I'm wrong) but it looks like the host header in ALB rules (`convox.site`) is using different domain name, while the internal domain created in Route53 is something like `rack.${AWS::StackName}.convox.`.

ddollar:
It is helpful to have a hostname that works, even before you have configured DNS.

I would support this PR if it could be triggered with an app-level CloudFormation parameter. Defaults to on, can be toggled off with a CloudFormation parameter. `DefaultDomain=Yes|No`, defaulted to `Yes`

ffjia:
I could add CFN parameter like this:

```
+{{ define "domain-params" }}
+  {{ range .Services }}
+    "{{ upper .Name }}DefaultDomain": {
+      "Type": "String",
+      "Description": "Create Default Domain",
+      "Default": "Yes",
+      "AllowedValues": [
+        "Yes",
+        "No"
+      ]
+    },
+  {{ end }}
+{{ end }}
```

However I couldn't figure out good way to implement if condition checking (for ALB listener rules creation) w/ this parameter in the app.json.tmpl. Hints?

ffjia:
Updated as you required, please help review.

provider/aws/formation/app.json.tmpl
```diff
@@ -213,6 +220,26 @@
   {{ end }}
 {{ end }}
 
+{{ define "domain-condition" }}
+  {{ range .Services }}
+    "{{ upper .Name }}DefaultDomain": { "Fn::Equals": [ { "Ref": "{{ upper .Name }}DefaultDomain" }, "Yes" ] },
```
ddollar: Let's use just one `DefaultDomain` parameter to avoid the parameter proliferation (CloudFormation has a hard limit of 60 parameters per stack)

provider/aws/formation/app.json.tmpl
```diff
@@ -248,7 +275,10 @@
   {{ range .Manifest.Services }}
     "Service{{ upper .Name }}Service": {
       "Type": "AWS::ECS::Service",
-      {{ if .Port.Port }}
```
ddollar: This chunk seems incorrect. It would only create the service if DefaultDomain=Yes?

I don't think this branch is necessary.

ffjia: From my understanding, if user specifies `Port` in convox.yml, and set the `DefaultDomain` to `No`, then the ECS Service will have dependency problem due to no ALB listener rule created?

provider/aws/formation/app.json.tmpl
ffjia: Okay, I will update it with one parameter.

provider/aws/formation/app.json.tmpl
ddollar: With this code if the user specifies `port:` but not `domain:` and sets `DefaultDomain=No` the service would not be created.

ffjia: Yep, since in that case, no ALB listener rules will be created, so (correct me if I'm wrong) it's impossible to add Load Balancer to ECS service. 

Do we want to create the ECS service without Load Balancer? Any better idea?

ddollar:
Handled by #2562

PR #2453, opened 2018-01-17
Add cert create option

ffjia:
Try to fix 1st issue in #2430.


ddollar:
The certificate will be needed regardless. I think the answer here is to try to look for a matching certificate in the account and use that rather than creating a new certificate if possible.

ddollar:
Closing this for now. The way to build this will be to search the account for a certificate that matches the application's custom domain and use that if found.

PR #2465, opened 2018-01-25
allow convox-env to pass through signals

robinjoseph08:
Make `convox-env` forward signals correctly to the wrapped command. Without this, if ECS sends a `SIGTERM` to your container, `convox-env` would get the signal but wouldn't give it to the `CMD`, which means that your application wouldn't be able to gracefully shutdown.

I'm still fairly new to golang, so any and all feedback would be appreciated!

### DIFF
--- a/cmd/convox-env/main.go
+++ b/cmd/convox-env/main.go
@@ -11,9 +11,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
-	"os/exec"
 	"strings"
-	"syscall"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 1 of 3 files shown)


ddollar:
Nice work, thanks!

cmd/convox-env/main.go
```diff
@@ -39,27 +35,16 @@ func main() {
...
-		}
-	case nil:
-		os.Exit(0)
-	default:
+	if err != nil {
 		fmt.Fprintf(os.Stderr, "ERROR: %s\n", err)
 		os.Exit(1)
 	}
+
+	os.Exit(0)
```
ddollar: is this exit necessary?

robinjoseph08: nope, it should just exit cleanly if it gets to this point. good call, will remove

ddollar:
Had to make a quick tweak to `convox-env`, can you please rebase this and I'll get it in the next release.

ddollar:
Here's the conflicting PR for reference: #2467

robinjoseph08:
@ddollar updated!

PR #2499, opened 2018-02-22
allow importing of env as wildcard

ddollar:
allows a service to import the app's entire environment by specifying `*` in the list, e.g.:

```
services:
  web:
    environment:
      - "*"
      - OTHER=override
```


mwarkentin:
@ddollar this is great, thanks!

Issue #2476, opened 2018-01-31
[GEN2] Fail to update Service Name when have Domain set, due to priority conflict


ffjia:
For example, if I need to rename `nginx` to `nginx2`, CFN update will fail, since the ALB Listener Rule Priority is the same.
```
services:
  nginx:
    domain:
      - nginx.domain.tld
```

```
services:
  nginx2:
    domain:
      - nginx.domain.tld
```

PR #2510, opened 2018-03-01
fix ALB Listener Rule Priority conflict.

ffjia:
This is the fix for #2476.


provider/aws/formation/app.json.tmpl
```diff
@@ -80,7 +80,7 @@
         "Actions": [ { "Type": "forward", "TargetGroupArn": { "Ref": "Balancer{{ upper .Name }}TargetGroup{{ if .Internal }}Internal{{ end }}" } } ],
           "Conditions": [ { "Field": "host-header", "Values": [ { "Fn::Join": [ ".", [ "{{$.App}}-{{.Name}}", { "Fn::ImportValue": { "Fn::Sub": "${Rack}:{{ router .Name $.Manifest }}Host" } } ] ] } ] } ],
           "ListenerArn": { "Fn::ImportValue": { "Fn::Sub": "${Rack}:{{ router .Name $.Manifest }}Listener80" } },
-          "Priority": "{{ priority $.App .Name }}"
+          "Priority": "{{ priority $.App (print .Name $service) }}"
```
ddollar: .Name is already the name of the service here. Isn't this just creating duplication?

provider/aws/formation/app.json.tmpl
```diff
@@ -156,7 +156,7 @@
             "Actions": [ { "Type": "forward", "TargetGroupArn": { "Ref": "Balancer{{ upper $service.Name }}TargetGroup{{ if $service.Internal }}Internal{{ end }}" } } ],
               "Conditions": [ { "Field": "host-header", "Values": [ "{{$domain}}" ] } ],
               "ListenerArn": { "Fn::ImportValue": { "Fn::Sub": "${Rack}:{{ router $service.Name $.Manifest }}Listener80" } },
-              "Priority": "{{ priority $.App $domain }}"
+              "Priority": "{{ priority $.App (print $service $domain) }}"
```
ddollar: This should probably be `$service.Name`

provider/aws/formation/app.json.tmpl
ffjia: You are right, will revert it.

PR #2513, opened 2018-03-01
Added export for NatIPs

radekl:
Exported NAT IPs are useful if you want to use those values in other CloudFormation Stacks. 

Personally we use it to create `security-group` and want to allow access from Private Rack (NATed)


ddollar:
It's not possible to export empty values

PR #2527, opened 2018-03-07
add parameter to run timers in fargate

ddollar:
Adds a `FargateTimers` parameter to gen2 apps to run scheduled timers in Fargate instead of on-cluster.


jontonsoup:
@ddollar how does this interact with memory limits for services?

PR #2581, opened 2018-04-04
default convox start to gen2

ddollar:


robatwave:
@ddollar how does this work for cases where an application is Gen 1, but uses a docker-compose file called something other than `docker-compose.yml`?

ddollar:
Run `convox start -g 1`

PR #2600, opened 2018-04-13
remove converging status to avoid ecs throttling

ddollar:


mwarkentin:
Does this remove "converging" from the rack status entirely? I did find it useful for monitoring rack level actions (eg. instance rolls, etc.) however did not find it useful to show converging if any app in the rack was not 100% stable (which I'm assuming is related to the piece about ECS throttling). Any chance we can keep the former behaviour without the latter?

PR #2614, opened 2018-04-21
service autoscaling

ddollar:
```
service:
  web:
    scale:
      count: 1-10
      targets:
        cpu: 70
        memory: 90
        requests: 200
        custom:
          AWS/SQS/ApproximateNumberOfMessagesVisible:
            aggregate: max
            value: 20
            dimensions:
              QueueName: myqueue
```

The following targets are available:

* `cpu`: average cpu utilization (%) across all processes of the service
* `memory`: average memory utilization (%) across all processes of the service
* `requests`: requests per minute per process

The autoscaler will determine the number of containers required to keep the targets at their desired value. If multiple targets are specified the largest number of containers will be used after taking all targets into account.

### DIFF
--- a/manifest/manifest_test.go
+++ b/manifest/manifest_test.go
@@ -130,6 +130,15 @@ func TestManifestLoad(t *testing.T) {
 						Cpu:      50,
 						Memory:   75,
 						Requests: 200,
+						Custom: manifest.ServiceScaleMetrics{
+							{
+								Aggregate:  "max",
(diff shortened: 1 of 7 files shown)


mwarkentin:
Thoughts on allowing for a “custom cloudwatch metric”? Would be nice for things like workers that would let you autoscale on things like queue depth.

ddollar:
Possibly, but trying to keep `convox.yml` cloud-agnostic so would need to think about how to encode that.

PR #2634, opened 2018-05-03
Discussion: reducing API calls for convox run

thieman:
@ddollar No actual code changes here, but wanted to get your thoughts on how we might approach the following problem.

Throttling is still a major issue for us. I was able to throw debug logs on in our development rack while running a simple `convox run`. These logs are seriously hard to parse (streaming the logs generates a bunch of CloudWatch garbage) but I was able to transcribe a portion of this run: https://gist.github.com/thieman/7db3d16d59d04580fa8b58b588db412c

These were all recorded within a 1 second period. In that 1 second period, we made at least 24 calls to ECS.

The comments in this PR line up with areas where we are making multiple ECS calls (most of the time, within loops) within the `convox run` flow. My proposal is to attempt to rewrite some of these to do fewer, less filtered calls rather than many targeted calls. It is a shame that AWS does not reward the API consumer for attempting to be a good citizen by doing filtered calls, but that seems to be the reality we're living in.

If any of these approaches make sense, I can start working on PRs to consolidate the calls.

### DIFF
--- a/provider/aws/processes.go
+++ b/provider/aws/processes.go
@@ -237,6 +237,10 @@ func (p *AWSProvider) stackTasks(stack string) ([]string, error) {
 
 	tasks := []string{}
 
+	// THROTTLING: stackTasks
+	// Calling this in a loop over services, could we instead
+	// make a single call and filter for only the services
+	// we care about in memory?
 	for _, s := range services {
 		err := p.ecs().ListTasksPages(&ecs.ListTasksInput{
 			Cluster:     aws.String(p.Cluster),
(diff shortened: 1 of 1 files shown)


provider/aws/processes.go
```diff
@@ -308,6 +312,10 @@ func (p *AWSProvider) taskProcesses(tasks []string) (structs.Processes, error) {
 
 	pss := structs.Processes{}
 
+	// THROTTLING: taskProcesses
+	// This one looks hairy, but if we could find some way
+	// to consolidate into a single API call rather than
+	// one for each task, we might get some gains here.
```
thieman: My comment here's a little off on further reflection. Specifically I think the loop with `p.fetchProcess` below this is what's generating a bunch of calls

thieman:
We haven't had a single instance of throttling since we reduced our ECS polling intervals. Without the constant rate of high background calls it looks like the calls made as part of a `convox run` are tolerable.

PR #2653, opened 2018-05-09
if we log in to a host with only one remote rack, switch to it

ddollar:


exalted:
@ddollar thanks for this. Do you think it would also be nice to pick first rack if there are more than one available? Because I feel like that’s a better UX, compared to when you login to console.convox.com and the previously selected rack is not even available. 

WDYT?

ddollar:
Would rather not auto-select if ambiguous as many Convox operations can be dangerous on the wrong Rack.

exalted:
Fair enough. 👍

PR #2655, opened 2018-05-09
Disable CloudWatchLogs debug logs

thieman:
CloudWatch logs a _ton_ of stuff when you have debug logs enabled and are tailing your rack logs. These logs are also recursive (the logs being streamed are once again logged). This significantly impacts the usability of the debug logs.

Proposal here is to disable CloudWatch logging, even when the DEBUG environment variable is set. This could hinder any debugging efforts specifically targeting CloudWatch, but those seem relatively rare and low-priority compared to debugging calls to other AWS services like EC2, ECS, and CF.

### DIFF
--- a/provider/aws/aws.go
+++ b/provider/aws/aws.go
@@ -144,12 +144,11 @@ func (p *AWSProvider) cloudformation() *cloudformation.CloudFormation {
 }
 
 func (p *AWSProvider) cloudwatch() *cloudwatch.CloudWatch {
-	config := p.config().WithLogLevel(aws.LogOff)
-	return cloudwatch.New(session.New(), config)
+	return cloudwatch.New(session.New(), p.config())
 }
 
 func (p *AWSProvider) cloudwatchlogs() *cloudwatchlogs.CloudWatchLogs {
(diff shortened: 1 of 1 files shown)


thieman:
Lest anyone think I'm being hyperbolic, here's what a single line of this output looks like: https://gist.github.com/thieman/6f8993ff2e1db68363e1226e3726157e

provider/aws/aws.go
```diff
@@ -144,7 +144,8 @@ func (p *AWSProvider) cloudformation() *cloudformation.CloudFormation {
 }
 
 func (p *AWSProvider) cloudwatch() *cloudwatch.CloudWatch {
-	return cloudwatch.New(session.New(), p.config())
+	config := p.config().WithLogLevel(aws.LogOff)
```
ddollar: Wouldn't this need to be on the `cloudwatchlogs` helper?

Can also be done without the extra variable.

thieman: Oh interesting, I didn't even see the other client. Agree that that's probably what we want.

PR #2657, opened 2018-05-10
Added tag with process name to App's Target Group (gen2)

radekl:
This allows us to use tags in our dashboards and charts instead of target group names (which are random, changing and hard to set up in general) to filter out metrics. In our case we are using datadog and it's supporting tags out of the box, this feature is really useful because we are using in some application multiple processes and want to check each of them separately - currently only possible for whole application (including all subtasks).

### DIFF
--- a/provider/aws/formation/app.json.tmpl
+++ b/provider/aws/formation/app.json.tmpl
@@ -240,6 +240,9 @@
             { "Key": "deregistration_delay.timeout_seconds", "Value": "30" },
             { "Key": "stickiness.enabled", "Value": "true" }
           ],
+          "Tags": [
+            { "Key": "ProcessName", "Value": "{{ upper .Name }}" }
+          ],
           "TargetType": { "Fn::If": [ "FargateServices", "ip", "instance" ] },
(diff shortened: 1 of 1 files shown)


provider/aws/formation/app.json.tmpl
```diff
@@ -240,6 +240,9 @@
             { "Key": "deregistration_delay.timeout_seconds", "Value": "30" },
             { "Key": "stickiness.enabled", "Value": "true" }
           ],
+          "Tags": [
+            { "Key": "ProcessName", "Value": "{{ upper .Name }}" }
```
ddollar: Let's call this `Service`. Might as well add `App` as well while you're in there.

radekl: App is populated in tag Name from cloudformation tags, though I see no reason why not to add this to be more explicit

ddollar: Gotcha. If `App` is already in there don't worry about it.

radekl: It is not called App, so for the clearness I have added the App as well. Not sure if used proper variables, though looking on other entries it should be ok.

PR #2658, opened 2018-05-10
Expose service role name to extend Role's permissions

radekl:
In addition to ServiceRole ARN it is very useful to have Service Role name as well - When creating new ManagedPolicy in Cloudformation it is required to provide Role name instead of ARN (ARN is good for usecases with S3, or other resource attached permissions). It allows to extend Task roles and add permissions for application to access various resources without using Access Keys.


ddollar:
It is much better to use the `IamPolicy` parameter for an app to add additional policies to the ServiceRole. Any manual changes you make could be lost on a subsequent update.

PR #2673, opened 2018-05-17
use an existing certificate for services if available

ddollar:
if a certificate already exists that matches all of the domains for a service, use it instead of creating a new one


ndbroadbent:
I'm seeing a request for a duplicate certificate in AWS Certificate Manager. 

My convox.yml is like this:

```
services:
  web:
    domain:
      - subdomain.example.com
```

The first request sent verification emails to `postmaster@subdomain.example.com`, `administrator@subdomain.example.com`, etc. 

On my next deploy, it created another certificate for the same domain, with a status of `Pending validation`. But it sent out the verification emails to different addresses: `postmaster@example.com`, `administrator@example.com`, etc.

PR #2685, opened 2018-05-21
autoscale via scheduled lambda

ddollar:


mwarkentin:
Starting to carve out the monitor process into lambda?

PR #2691, opened 2018-05-23
add shared environment section to manifest

ddollar:
adds a new `environment:` section to the manifest which defines env vars which are shared by all services on the app:

```
environment:
  - REQUIRED
  - DEFAULT=value
services:
  web:
    environment:
      - REQUIRED=override
      - DEFAULT=override
```

### DIFF
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -276,7 +276,7 @@ func build2(dir string) error {
 		return err
 	}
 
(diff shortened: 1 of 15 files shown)


idris:
@ddollar it would be nice if we could specify different sets of "shared" environments.  I have 2 services that share almost everything, and then 1 service that shares almost nothing.  But for security, I don't want that latter service to have access to all of those other variables it doesn't need.

PR #2700, opened 2018-05-29
Move default BuildInstance type into install.go, instead of rack.json

### LINKED ISSUE #2699: Respect --build-instance "" for convox install
I would like to install a rack with no build instance, but the rack param `BuildInstance` is being set to `t2.small` by default. It's not possible to override this with `convox install --build-instance ""`.

I noticed `install.go` has:

```
			cli.StringFlag{
				Name:   "build-instance",
				Value:  "",
				Usage:  "instance type for a dedicated build cluster",
				EnvVar: "RACK_BUILD_INSTANCE",
			},

...

			cli.StringFlag{
				Name:  "instance-type",
				Value: "t2.small",
				Usage: "type of instances in the rack",
			},
```

And in `rack.json`:

```
    "BuildInstance": {
      "Type": "String",
      "Description": "Instance type for a dedicated build cluster",
      "Default": "t2.small"
    },

...

    "InstanceType": {
      "Default": "t2.small",
      "Description": "The type of the instances in the runtime cluster",
      "Type": "String"
    },
```

Maybe the default value could only be set in `install.go` for build-instance, instead of `rack.json`?

ndbroadbent:
This allows the `BuildInstance` param to be overridden with `convox install --build-instance ""`.

Fixes #2699

### DIFF
--- a/cmd/convox/install.go
+++ b/cmd/convox/install.go
@@ -87,7 +87,7 @@ func init() {
 			},
 			cli.StringFlag{
 				Name:   "build-instance",
-				Value:  "",
+				Value:  "t2.small",
 				Usage:  "instance type for a dedicated build cluster",
 				EnvVar: "RACK_BUILD_INSTANCE",
 			},
--- a/provider/aws/formation/rack.json
+++ b/provider/aws/formation/rack.json
@@ -398,7 +398,7 @@
     "BuildInstance": {
       "Type": "String",
       "Description": "Instance type for a dedicated build cluster",
-      "Default": "t2.small"
+      "Default": ""
     },
     "BuildMemory": {
       "Type": "String",


ndbroadbent:
Hmm, I'm not sure why this isn't working, but I just tried to install a rack and `BuildInstance` was still set to `t2.small`.

ndbroadbent:
Oh right, it's because I was only using my locally modified CLI, but I was still installing the official release of `convox/rack` that uses the original `rack.json` (with the default value for `BuildInstance`.)

Not too sure how to deploy a custom rack to my cluster. But anyway, it looks like this PR will fix the issue.

provider/aws/formation/rack.json
```diff
@@ -398,7 +398,7 @@
     "BuildInstance": {
       "Type": "String",
       "Description": "Instance type for a dedicated build cluster",
-      "Default": "t2.small"
+      "Default": ""
```
radekl: This looks ok as there is a condition based on that.

cmd/convox/install.go
```diff
@@ -87,7 +87,7 @@ func init() {
 			},
 			cli.StringFlag{
 				Name:   "build-instance",
-				Value:  "",
+				Value:  "t2.small",
```
radekl: This will add a default value of `t2.small` which will each time set the BuildInstance parameter in Cloud formation. I would remove that and leave only the rack.json option.

ndbroadbent: Yes, by default this will cause `convox install` to set the `BuildInstance` parameter to `t2.small`, unless it is overridden with `convox install --build-instance ""`. 

The current behavior is to set the default value of `BuildInstance` in `rack.json`, but the default behavior cannot be overridden during the `convox install` stage.

ndbroadbent: I would also be in favor of removing the default build instance, if that's an option. (Setting both values to `""`.) I think the majority of users probably don't want to run a dedicated instance for builds (but I might be wrong.)

radekl: This relates to following commit: https://github.com/convox/rack/commit/e39400a314e8938dd1768518c4a944228cc90553

ddollar: The default is now for a BuildInstance as the experience is significantly improved with one. I think it's appropriate to keep that default in the CloudFormation stack.

I would be amenable to a `--no-build-instance` flag to `convox install`

ndbroadbent: `--no-build-instance` works for me!

ddollar:
Going to close this for now. Please feel free to re-open with the suggested changes.

PR #2703, opened 2018-05-29
remove period uninstalling convox message

ndbroadbent:
I was a bit confused by the periodic `Uninstalling Convox...` messages in the log output. I thought the command might be restarting or something was going wrong.


ddollar:
I think it would be better to just remove this message entirely. Thoughts?

ndbroadbent:
Either way works for me, or even just leaving it the way it is. Not too important!

ddollar:
Doesn't seem to be accomplishing much. Feel free to remove it.

ndbroadbent:
Thanks, I updated the PR to remove it

PR #2730, opened 2018-06-15
Add AccessLoggingPolicy for g1 apps

mwarkentin:
I attempted to add the required config to do access logging for gen1 ELBs if `LogBucket` parameter is set for the app.

I'll need some help testing / validating that this was done correctly.

### DIFF
--- a/provider/aws/formation/g1/app.json.tmpl
+++ b/provider/aws/formation/g1/app.json.tmpl
@@ -328,6 +328,15 @@
                 "TargetGroupArn": { "Fn::If": [ "Balancer{{ upper $e.Name }}ALB",
                   { "Ref": "{{ $manifest.BalancerResourceName $e.Name }}TargetGroup" },
                   { "Ref": "AWS::NoValue" }
+                ] },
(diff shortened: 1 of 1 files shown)


provider/aws/formation/g1/app.json.tmpl
```diff
@@ -328,6 +328,15 @@
...
                   { "Ref": "{{ $manifest.BalancerResourceName $e.Name }}TargetGroup" },
                   { "Ref": "AWS::NoValue" }
+                ] },
+                "AccessLoggingPolicy": { "Fn::If": [ "BlankLogBucket",
+                  { "Ref": "AWS::NoValue" },
+                  {
+                    "EmitInterval": 5,
+                    "Enabled": true,
+                    "S3BucketName": { "Ref": "LogBucket" },
+                    "S3BucketPrefix": { "Fn::Sub": "convox/logs/${AWS::StackName}/elb/" }
```
mwarkentin: Not sure if we should be more specific with the prefix as there could be multiple ELBs for a single app?

ddollar:
Moved to #2731 and tweaked a bit

PR #2733, opened 2018-06-18
Support pagination in describeRackStacks

thieman:
We're hitting a bug where `convox uninstall` is returning the following error message to us:

```
ERROR: can not find rack named (our rack name)
Are you authenticating with the correct AWS account?
See AWS_ACCESS_KEY_ID, AWS_SECRET_ACCESS_KEY, and ~/.aws/credentials
```

Pretty sure my credentials are fine as I just used the same command and credentials to successfully uninstall a different rack in our account.

I think what's happening here is we're using `DescribeStacks` directly but making no effort to consume multiple pages of information. I suspect the rack in question is not included in the first page of results and causes our error. This PR moves us from using `DescribeStacks` directly to instead use `DescribeStacksPages` which handles consuming the pages for us.

One note: I think the scoping of `apps`, `rack`, and `resources` is fine but I'm a bit nervous about them as I've been bitten by scoping before. This function also does not seem to be covered directly by any tests (though something calling it may be, I'm not sure).

We've had this problem before: https://github.com/convox/rack/pull/1385. It may be worth a wider audit of our `DescribeStacks` usage.

### DIFF
--- a/cmd/convox/uninstall.go
+++ b/cmd/convox/uninstall.go
@@ -400,58 +400,62 @@ func deleteStacks(stackType, rackName string, CF *cloudformation.CloudFormation)
 
 // describeRackStacks uses credentials to describe CF service, app and rack stacks that belong to the rack name and region
 func describeRackStacks(rackName string, CF *cloudformation.CloudFormation) (Stacks, error) {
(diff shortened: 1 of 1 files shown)


ddollar:
This will be handled by the CLI change coming in #2739

PR #2736, opened 2018-06-20
add singleton attribute for services

ddollar:
Add `singleton: true` to a service to cause the deployment scheduler to terminate existing containers before starting new ones


thieman:
Where's the best place to document this? Should it be in here: https://convox.com/docs/convox-yml/

PR #2741, opened 2018-06-21
Add LogServiceEventsIntervalSeconds rack param to control ECS polling interval

thieman:


ddollar:
Thanks! Carrying forward in #2742

Issue #2740, opened 2018-06-21
convox roll api key no longer working


swisslala:
rack version: Version  20180618165943

Click on roll API key - below is an error in the rack logs and then trying to ssh into the instances fails with no key found...

`2018-06-21T10:18:05Z service/web:20180618165943/b7cc659b0950 + mkdir -p /etc/ssl/convox
2018-06-21T10:18:05Z service/web:20180618165943/b7cc659b0950 + openssl genrsa -des3 -passout pass:x -out /etc/ssl/convox/server.pass.key 2048
2018-06-21T10:18:05Z service/web:20180618165943/b7cc659b0950 Generating RSA private key, 2048 bit long modulus
2018-06-21T10:18:05Z service/web:20180618165943/b7cc659b0950 ......................................+++
2018-06-21T10:18:05Z service/web:20180618165943/b7cc659b0950 .......+++
2018-06-21T10:18:05Z service/web:20180618165943/b7cc659b0950 e is 65537 (0x010001)
2018-06-21T10:18:05Z service/web:20180618165943/b7cc659b0950 139806850295040:error:28069065:UI routines:UI_set_result:result too small:../crypto/ui/ui_lib.c:778:You must type in 4 to 1023 characters
2018-06-21T10:18:05Z service/web:20180618165943/b7cc659b0950 139806850295040:error:28069065:UI routines:UI_set_result:result too small:../crypto/ui/ui_lib.c:778:You must type in 4 to 1023 characters
2018-06-21T10:18:05Z service/web:20180618165943/b7cc659b0950 139806850295040:error:0906906F:PEM routines:PEM_ASN1_write_bio:read key:../crypto/pem/pem_lib.c:336:
2018-06-21T10:18:05Z service/web:20180618165943/c09d9d308dc1 ns=provider.aws at=SystemGet state=success elapsed=0.042
`

and

`2018-06-21T10:18:33Z service/web:20180618165943/c09d9d308dc1 ns=api.controllers at=instance.ssh state=error error="no instance key found" location="/go/src/github.com/convox/rack/api/controllers/api.go:92" elapsed=93.562`

swisslala:
Using new rack 20180625222015 still see the same issue. We basically cannot ssh into these machines...

ddollar:
I am unable to reproduce this issue. Key rolling and SSH are working fine in all of my tests.

swisslala:
[...]

PR #2746, opened 2018-06-29
Use a longer password for certificate generation

nitaigao:
The minimum password length is 4 characters from OpenSSL 1.1.0 onwards. When trying to boot a local rack on my workstation unfortunately this stops the rack from starting.

This PR generates a random password of 16 characters in place of the 1 character `x` in the `bin/gen-cert` script.

```
$ ./bin/gen-cert                                       
+ mkdir -p /etc/ssl/convox
+ openssl genrsa -des3 -passout pass:x -out /etc/ssl/convox/server.pass.key 2048
Generating RSA private key, 2048 bit long modulus
....+++
.......+++
e is 65537 (0x010001)
139636344934848:error:28069065:UI routines:UI_set_result:result too small:crypto/ui/ui_lib.c:765:You must type in 4 to 1023 characters
139636344934848:error:28069065:UI routines:UI_set_result:result too small:crypto/ui/ui_lib.c:765:You must type in 4 to 1023 characters
139636344934848:error:0906906F:PEM routines:PEM_ASN1_write_bio:read key:crypto/pem/pem_lib.c:336:
```

### DIFF
--- a/bin/gen-cert
+++ b/bin/gen-cert
@@ -7,8 +7,10 @@ set -ex
 
 mkdir -p /etc/ssl/convox
 
-openssl genrsa -des3 -passout pass:xxxx -out /etc/ssl/convox/server.pass.key 2048
-openssl rsa -passin pass:xxxx -in /etc/ssl/convox/server.pass.key -out /etc/ssl/convox/server.key
+PASSWORD=$(openssl rand -hex 16)
+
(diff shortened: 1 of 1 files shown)


nitaigao:
Also possible fix for https://github.com/convox/rack/issues/2740

PR #2751, opened 2018-07-04
allow 401 to pass health check

ddollar:


mwarkentin:
Seems a bit strange - if you specify an invalid path to your health endpoint and it returns a 404, this would be considered healthy?

ddollar:
good point, modified so that only 401 passes for those third-party apps where modifying the source is difficult and everything is behind auth

PR #2753, opened 2018-07-06
prune docker images from build instance

ddollar:


mwarkentin:
Any thoughts on how to do this on the instances that the api containers run on? We find we have a bigger issue with those as it's 2 out of 45 instances that has the issue. Maybe just do the same thing on all of the cluster instances?

ddollar:
You should not be seeing this issue on the instances the API containers run on unless you are running `docker build` manually there somehow. ECS cleans up after itself manually for all of the images it starts.

ddollar:
Actually thinking more about it I can see how gen1 apps could cause this on the API container instances. I'm hesitant to add this to all instances for fear of interfering with the images the ECS agent expects to be there. Will need to do a bit more research toward that end.

mwarkentin:
Yeah, the api containers do a docker pull on each deploy which doesn’t get cleaned up by the ECS agent. 

Fwiw our fix is to run docker prune —all —force (I think those are the options) on those instances. We haven’t seen any negative impact after doing that.

advance512:
Will this help when using the overlay fs?

PR #2788, opened 2018-07-23
remove unnecessary VpcConfig from syslog resource

ddollar:


jmcvetta:
Fwiw, this change broke my Convox --> private Logstash setup.  The only way I could find to fix it was to leave Logtash port open to the public internet.  :(

PR #2801, opened 2018-07-25
add vpc endpoints for s3

ddollar:


ddollar:
adding this broke network connectivity for a few things, will revisit later

PR #2813, opened 2018-07-30
Ubuntu 18 compatibility fixes for Local rack installation.

ectrimble20:
Adds in an OS check and breaks Linux installation of the local rack's router into two functions, one specific to Ubuntu 18.04 and one as a generic Linux installation.  Ubuntu 18 no longer uses NetworkManager as it's default networking platform.  This fix installs a configuration file for systemd-resolv/resolved which allows the convox router to act as a pass through DNS server which in turn handles .convox address resolution without effecting existing external DNS resolution.


router/helpers.go
```diff
@@ -31,3 +33,36 @@ func writeFile(path string, data []byte) error {
 
 	return ioutil.WriteFile(path, data, 0644)
 }
+
+func linuxRelease() (string, error) {
```
ddollar: can move this to the _linux specific file since it's linux specific

router/helpers.go
```diff
@@ -31,3 +33,36 @@ func writeFile(path string, data []byte) error {
...
+	os_info := exec.Command("/bin/cat", "/etc/os-release")
```
ddollar: best way to read a file contents in go is

```
data, err := ioutil.ReadFile("/etc/os-release")
```

router/helpers.go
```diff
@@ -31,3 +33,36 @@ func writeFile(path string, data []byte) error {
...
+	var buf bytes.Buffer
+	os_info.Stdout = &buf
+
+	err := os_info.Run()
+
+	if err != nil {
+		return "unknown", err
+	}
+
+	output := buf.String()
```
ddollar: I would probably break this into two functions, sort of like this pseudocode

```
var reAttributeMatcher = regexp.MustCompile("...")  (substitute appropriate regex here, see https://golang.org/pkg/regexp/)


func linuxReleaseAttributes() (map[string]string, error) {
  attrs := map[string]string{}
  data = read(/etc/os-release)
  s := bufio.NewScanner(bytes.NewReader(data))
  for s.Scan() {
    if m := reAttributeMatcher.FindStringSubmatch(s.Text()); len(m) >= 2 {
     attrs[m[0]] = m[1]
    }
  }
}

func linuxRelease() (string, error) {
  attrs, err := linuxReleaseAttributes()
  if err return err
  return fmt.Sprintf("%s-%s", attrs["NAME"], attrs["VERSION_ID"])
}

File: router/dns_linux.go [hunk 4]
```diff
@@ -6,6 +6,49 @@ import (
...
+	default:
+		if err := installGenericLinux(domain, ip); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func installUbuntu1804(domain string, ip net.IP) error {
+
```
ddollar: dont need the extra space at the beginning of functions

router/helpers_linux.go
```diff
@@ -0,0 +1,37 @@
...
+		return nil, err
+	}
+	s := bufio.NewScanner(bytes.NewReader(data))
+	for s.Scan() {
+		line := s.Text()
+		if len(line) == 0 {
+			continue
+		}
+		line = strings.Replace(line, "\"", "", -1)
+		line = strings.ToLower(line)
```
ddollar: This ToLower is likely going to conflict with looking for NAME and VERSION_ID in the function below

ddollar: In general I'd recommend a regular expression for this kind of parsing just for flexibility as well.. You should be able to find one in the wild that supports the parsing of `NAME="value"` (or `NAME=value`) style lines

router/dns_linux.go
```diff
@@ -6,6 +6,47 @@ import (
...
+		}
+	default:
+		if err := installGenericLinux(domain, ip); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func installUbuntu1804(domain string, ip net.IP) error {
```
ddollar: Probably best to call these `setupResolverUbuntu1804` and `setupResolverGeneric` just for easy reference in the future

router/helpers_linux.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"regexp"
+)
+
+var kvpair = regexp.MustCompile(`(.*)=([^"].*|\"(.*)\")`)
```
ddollar: Try a regex like this, should simplify your code below a bit

https://play.golang.org/p/cUO1wi_BCj1

ectrimble20: Implemented and tested your recommended Regex.  Appears to work better, produces 3 results instead of 4 with an empty possibility which simplifies testing the response.

ddollar:
Not sure what has happened here in the rebase but it seems to be missing some commits (the first commit doesn't make any sense)

You may want to re-clone the repo somewhere, copy your final changed files in, and just make 1 commit with the changes and push it up to a branch.

PR #2838, opened 2018-08-13
allow network isolation of app services

ddollar:
Adds a new app parameter `Isolate`. When set to `Yes` will configure all app services to use `awsvpc` networking mode and only allow incoming traffic from the Rack's ALB.

Notes:
* This parameter only changes service behavior when the rack is `Private=Yes`
* In this mode each container requires a new ENI so you will want to use an instance type for your Rack with enough ENI capacity.

https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-eni.html#AvailableIpPerENI


mwarkentin:
Will this work with apps / services running on Fargate?

ddollar:
Yes

Issue #2846, opened 2018-08-14
Unable to deploy using latest rack due to PG database InvalidParameterCombination


qnm:
When deploying one of our apps using the latest rack, I see the following error in CF:

```
Invalid storage size for engine name postgres and storage type gp2: 10 (Service: AmazonRDS; Status Code: 400; Error Code: InvalidParameterCombination; Request ID: xxxf)
```

This is applied to `AWS::RDS::DBInstance` in a Convox `ResourceDatabase`.

My database looks like this:

```
Name     mydb
Type     postgres
Status   running
Options  AllocatedStorage=100
         Database=mydb
         Family=postgres9.6
         InstanceType=db.t2.medium
         MaxConnections=
         MultiAZ=false
         Password=****
         Username=blah
URL      postgres://blah
```

Is this related to changing the base postgres database family?

PR #2847, opened 2018-08-14
allow app-based resources to specify options

ddollar:
Generation 2 apps can now specify options to their resources:

```
resources:
  cache:
    type: redis
    options:
      class: cache.t2.micro
      durable: true
      nodes: 2
      version: 2.8.24
  database:
    type: postgres
    options:
      class: db.t2.large
      durable: true
      iops: 1000
      storage: 200
      version: 9.6.6
```

### DIFF
--- a/manifest/manifest_test.go
+++ b/manifest/manifest_test.go
@@ -19,6 +19,9 @@ func TestManifestLoad(t *testing.T) {
 			manifest.Resource{
 				Name: "database",
 				Type: "postgres",
+				Options: map[string]string{
+					"size": "db.t2.large",
+				},
 			},
 		},
 		Services: manifest.Services{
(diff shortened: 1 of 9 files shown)


qnm:
This looks good, thanks. 

This will solve the issue I have in https://github.com/convox/rack/issues/2846 pretty well. 

Longer term I’d expect the behaviour to be more elastic, like ‘scale’, where I can set defaults in the yml and they are applied once, afterwards I adjust parameters as my app grows and changes.

ddollar:
You can adjust all of these options in your `convox.yml` and redeploy without downtime.

qnm:
Yes, very true, but the downside to the yml-only approach is you enforce parity across environments where you may not need to e.g. a `micro` may be suitable for integration but `xlarge` is needed on production.

Regardless, this is much better than the situation I am in right now!

ddollar:
Specify your type like `class: ${DATABASE_CLASS}` and use an env var to make it dynamic per deployment.

PR #2851, opened 2018-08-15
update ecs amis

ddollar:


ndbroadbent:
Just checked out the new AMIs and it's still using the `devicemapper` driver. Hopefully they'll switch to `overlay2` soon

Issue #2853, opened 2018-08-16
Local provider is missing link env variables


slaskis:
Here's a sample project which shows the issue:

```
$ cat convox.yml
services:
  main:
    image: nginx:alpine
    port: 80
    links:
      - other
  other:
    image: nginx:alpine
    port: 80
```

Run it

```
convox start
```

Now we should be able to check the env vars of the main process. And this is where I expect to see a `OTHER_URL=https://other.test-links.convox/`

```
$ docker exec convox.test-links.service.main.1 env
PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
HOSTNAME=880b552ab150
C5099BC08F9C_PORT=tcp://172.17.0.2:5443
C5099BC08F9C_PORT_5443_TCP=tcp://172.17.0.2:5443
C5099BC08F9C_PORT_5443_TCP_ADDR=172.17.0.2
C5099BC08F9C_PORT_5443_TCP_PORT=5443
C5099BC08F9C_PORT_5443_TCP_PROTO=tcp
C5099BC08F9C_NAME=/convox.test-links.service.main.1/c5099bc08f9c
C5099BC08F9C_ENV_COMBINED=true
C5099BC08F9C_ENV_IMAGE=convox/rack:20180815173013
C5099BC08F9C_ENV_PROVIDER=local
C5099BC08F9C_ENV_RACK=convox
C5099BC08F9C_ENV_ROUTER=10.42.0.0
C5099BC08F9C_ENV_VERSION=20180815173013
C5099BC08F9C_ENV_VOLUME=/Users/Shared/convox
C5099BC08F9C_ENV_GOLANG_VERSION=1.10.3
C5099BC08F9C_ENV_GOPATH=/go
APP=test-links
RACK_URL=https://c5099bc08f9c:3000
RELEASE=RSOHGLUMXD
NGINX_VERSION=1.15.2
HOME=/root
```

Possibly related to #2835

slaskis:
As a side note to this. I can work around it locally by manually setting:

```
  main:
    environment:
      - OTHER_URL=http://other.test-links.convox/
```

but then when I deploy to aws I have to remove them again or the deploy will fail with an error about duplicate environment variables on a task.

ddollar:
Here is where resources are added to the env, you would need to run a similar loop for linked services.

https://github.com/convox/rack/blob/master/provider/local/converge.go#L391

slaskis:
[...]

PR #2862, opened 2018-08-22
add links to services in local provider

slaskis:
It adds a link like `OTHER_URL=http://other.test-links.convox` and I'm not sure if this is the correct urls but http seems to work better even with the convox ca installed.

I got my test in #2853 to now show `OTHER_URL` when running this local rack.

### DIFF
--- a/provider/local/converge.go
+++ b/provider/local/converge.go
@@ -361,6 +361,10 @@ func (p *Provider) resourceContainers(resources manifest.Resources, app, release
 	return cs, nil
 }
 
+func (p *Provider) serviceURL(app, name string) (string, error) {
+	return fmt.Sprintf("http://%s.%s.%s", name, app, p.Rack), nil
+}
+
 func (p *Provider) serviceContainers(services manifest.Services, app, release string) ([]container, error) {
 	cs := []container{}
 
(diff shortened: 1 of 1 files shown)


provider/local/converge.go
```diff
@@ -361,6 +361,10 @@ func (p *Provider) resourceContainers(resources manifest.Resources, app, release
 	return cs, nil
 }
 
+func (p *Provider) serviceURL(app, name string) (string, error) {
+	return fmt.Sprintf("http://%s.%s.%s", name, app, p.Rack), nil
```
ddollar: This should probably be `https://`

provider/local/converge.go
```diff
@@ -388,6 +392,20 @@ func (p *Provider) serviceContainers(services manifest.Services, app, release st
...
+		// add links
+		for _, sl := range s.Links {
+			for _, s := range m.Services {
+				if s.Name == sl {
+					u, err := p.serviceURL(app, s.Name)
+					if err != nil {
+						return nil, err
+					}
+
+					e[fmt.Sprintf("%s_URL", strings.ToUpper(sl))] = u
```
ddollar: Seems like this will break on services with dashes in the name, e.g. `my-service`

slaskis: oh yeah. good call!

slaskis: and it seems to be on par with the aws provider 

https://github.com/convox/rack/blob/master/provider/aws/formation/app.json.tmpl#L421

https://github.com/convox/rack/blob/master/provider/aws/template.go#L103

perhaps they should change to use upper / upperName instead?

slaskis: or no nvm...that seems to be "the other way around"

ddollar: Looks like you're right, go ahead and ignore this for now, will have to fix globally

PR #2863, opened 2018-08-22
add auth token and encryption to redis resource

slaskis:
without this i just get an error (`ERR Client sent AUTH, but no password is set`) while trying to connect to the `REDIS_URL` linked from the resource as the redis instance does not actually have a password set.

`TransitEncryptionEnabled` was required to use authtoken and then i added the `Encrypted` parameter as it seemed like a common thing to set

### DIFF
--- a/provider/aws/formation/resource/redis.json.tmpl
+++ b/provider/aws/formation/resource/redis.json.tmpl
@@ -62,6 +62,7 @@
         "Engine": "redis",
         "EngineVersion": { "Ref": "Version" },
         "NumCacheClusters": { "Ref": "Nodes" },
+        "AuthToken": { "Ref": "Password" },
         "Port": "6379",
         "ReplicationGroupDescription": { "Ref": "AWS::StackName" },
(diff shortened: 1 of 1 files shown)


provider/aws/formation/resource/redis.json.tmpl
```diff
@@ -55,16 +60,19 @@
...
       "Properties": {
+        "AtRestEncryptionEnabled": { "Ref": "Encrypted" },
         "AutomaticFailoverEnabled": { "Ref": "Durable" },
         "AutoMinorVersionUpgrade": "true",
         "CacheNodeType": { "Ref": "Class" },
         "CacheSubnetGroupName": { "Ref": "SubnetGroup" },
         "Engine": "redis",
         "EngineVersion": { "Ref": "Version" },
         "NumCacheClusters": { "Ref": "Nodes" },
+        "AuthToken": { "Ref": "Password" },
```
ddollar: Please put this parameter in alphabetical order

PR #2875, opened 2018-08-24
set redis password and protocol when encrypted

### LINKED ISSUE #2874: Cannot connect to redis resource from convox.yml
"Encrypted In Transit" is required for AuthToken on a redis resource. But this encrypts the traffic which is not supported by some client, such as `redis-cli` without tunneling. 

Some clients ([go](https://github.com/go-redis/redis), [node](https://github.com/NodeRedis/node_redis#rediscreateclient)) support encrypted transport by using the `rediss://` protocol instead of `redis://` but I'm not sure how widely used this is besides those. 

So I think the exported redis url should use the `rediss://` protocol if the `password` parameter is set, and the `redis://` protocol otherwise. Especially since overriding the resource url is not possible as far as I can see.

slaskis:
I've never used cloudformation json before but I copied some from the other templates and came up with this.

I've also validated it this time with aws-cli:

```
$ aws cloudformation validate-template --template-body file://${PWD}/provider/aws/formation/resource/redis.json.tmpl
{
    "Parameters": [
        {
            "ParameterKey": "Rack",
            "NoEcho": false
        },
        {
            "ParameterKey": "Version",
            "DefaultValue": "2.8.24",
            "NoEcho": false
        },
        {
            "ParameterKey": "Encrypted",
            "DefaultValue": "false",
            "NoEcho": false
        },
        {
            "ParameterKey": "Durable",
            "DefaultValue": "false",
            "NoEcho": false
        },
        {
            "ParameterKey": "Class",
            "DefaultValue": "cache.t2.micro",
            "NoEcho": false
        },
        {
            "ParameterKey": "Nodes",
            "DefaultValue": "1",
            "NoEcho": false
        },
        {
            "ParameterKey": "Password",
            "NoEcho": true
        }
    ]
}
```

Fixes #2874

### DIFF
--- a/provider/aws/formation/resource/redis.json.tmpl
+++ b/provider/aws/formation/resource/redis.json.tmpl
@@ -1,7 +1,7 @@
 {
   "AWSTemplateFormatVersion": "2010-09-09",
   "Conditions": {
-    "IsEncrypted": { "Fn::Equals": [{ "Ref": "Encrypted" }, "true"] }
+    "Encrypted": { "Fn::Equals": [{ "Ref": "Encrypted" }, "true"] }
   },
   "Parameters": {
     "Class": {
@@ -82,7 +82,7 @@
         "AtRestEncryptionEnabled": { "Ref": "Encrypted" },
         "AuthToken": {
           "Fn::If": [
-            "IsEncrypted",
+            "Encrypted",
             { "Ref": "Password" },
             { "Ref": "AWS::NoValue" }
           ]
--- a/provider/aws/formation/resource/redis.json.tmpl
+++ b/provider/aws/formation/resource/redis.json.tmpl
@@ -1,5 +1,8 @@
 {
   "AWSTemplateFormatVersion": "2010-09-09",
+  "Conditions": {
+    "IsEncrypted": { "Fn::Equals": [{ "Ref": "Encrypted" }, "true"] }
+  },
   "Parameters": {
     "Class": {
       "Type": "String",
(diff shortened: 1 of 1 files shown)


provider/aws/formation/resource/redis.json.tmpl
```diff
@@ -34,15 +37,31 @@
     }
   },
   "Outputs": {
-    "Url": { "Value": { "Fn::Sub": "redis://:${Password}@${ReplicationGroup.PrimaryEndPoint.Address}:${ReplicationGroup.PrimaryEndPoint.Port}/0" } }
+    "Url": {
+      "Value": {
+        "Fn::Sub": [
+          "${Protocol}://:${Password}@${ReplicationGroup.PrimaryEndPoint.Address}:${ReplicationGroup.PrimaryEndPoint.Port}/0",
```
ddollar: Should only include the Password in the URL if it's encrypted

provider/aws/formation/resource/redis.json.tmpl
```diff
@@ -1,5 +1,8 @@
 {
   "AWSTemplateFormatVersion": "2010-09-09",
+  "Conditions": {
+    "IsEncrypted": { "Fn::Equals": [{ "Ref": "Encrypted" }, "true"] }
```
ddollar: Also please call this condition just `Encrypted` for consistency with the other CloudFormation stacks

slaskis: Sure! I wasn't sure if it'd conflict but no problem

provider/aws/formation/resource/redis.json.tmpl
slaskis: Ah thanks, missed that

slaskis:
allright, ready for another review!

ddollar:
Looks good, thanks!

PR #2929, opened 2018-09-07
Bring back release:promote start events

thieman:
These were lost in The Great API Refactor of July 2018 https://github.com/convox/rack/pull/2739/files#diff-3e88de698e2d214b3e9a3822fbb7059fL88

We were using these in our build pipeline to help notify of us of the status of our Convox builds, so it'd be great to have them back

### DIFF
--- a/provider/aws/releases.go
+++ b/provider/aws/releases.go
@@ -304,7 +304,8 @@ func (p *Provider) ReleasePromote(app, id string) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


provider/aws/releases.go
```diff
@@ -304,6 +304,7 @@ func (p *Provider) ReleasePromote(app, id string) error {
 		return err
 	}
 
+	p.EventSend("release:promote", structs.EventSendOptions{Status: "start", Data: map[string]string{"app": a.Name, "id": r.Id}})
```
thieman: Not totally sure whether `a.Name` and `r.Id` are correct here

ddollar:
covered in #2927

PR #2939, opened 2018-09-18
retry logs on error during wait sessions

ddollar:


mwarkentin:
@ddollar Nice - I think this will make our Jenkins jobs a lot more stable! How many retries will it attempt? I didn't see the value in the PR.

PR #2957, opened 2018-09-26
fix for convox exec in awsvpc mode

ddollar:


jontonsoup:
@ddollar does this fix the issue where detached exec commands arent working? We've run into this on latest.

PR #2964, opened 2018-10-03
Making build description visible in releases.

iCreateJB:
Resolves rack/issues#2312


pkg/cli/releases_test.go
```diff
@@ -22,9 +22,9 @@ func TestReleases(t *testing.T) {
 		require.Equal(t, 0, res.Code)
 		res.RequireStderr(t, []string{""})
 		res.RequireStdout(t, []string{
-			"ID        STATUS  BUILD   CREATED   ",
-			"release1  active  build1  2 days ago",
-			"release2          build1  2 days ago",
+			"ID        STATUS  BUILD   CREATED     DESCRIPTION",
+			"release1  active  build1  2 days ago             ",
```
ddollar: Can you please modify the fixture so that one of these has a description that gets tested in the output

pkg/cli/fixtures_test.go
```diff
@@ -162,23 +162,25 @@ func fxRegistry() *structs.Registry {
...
-		Env:      "FOO=bar\nBAZ=quux",
-		Manifest: "manifest",
-		Created:  time.Now().UTC().Add(-49 * time.Hour),
+		Id:          "release2",
+		App:         "app1",
+		Build:       "build1",
+		Env:         "FOO=bar\nBAZ=quux",
+		Manifest:    "manifest",
+		Created:     time.Now().UTC().Add(-49 * time.Hour),
+		Description: "description2",
```
ddollar: Sorry I should have been more clear. Can you please set this one back to blank so we can test both cases (blank description vs one with data)

ddollar:
This failed in CI, looks like if the description is empty it has a problem saving the release.

https://circleci.com/gh/convox/rack/3327

Issue #2817, opened 2018-07-31
Rack password cannot contain certain special characters


stephenchu:
Version:

```
client: 20180728174119
server: 20180521013812
```

After I run `convox rack install ... Password=$PASSWORD`, if my `$PASSWORD` contains some special characters (such as `j6K%RgpygX`), then upon `convox login <elb> -p <password>` I get this error:

```
› convox login <name>.us-west-2.elb.amazonaws.com -p 'j6K%RgpygX'
Authenticating with <name>.us-west-2.elb.amazonaws.com... ERROR: parse https://convox:j6K%RgpygX@<name>.us-west-2.elb.amazonaws.com: invalid URL escape "%Rg"
```

crohr:
Got the same issue today when installing a brand new rack with a password containing a slash `/` character.

crohr:
And the authentication just failed with "invalid login", without the explicit error message @stephenchu got. Manually overriding the password in the Api service on ECS fixed the issue (after removing the slash character).

PR #2979, opened 2018-10-10
Escape special characters.

iCreateJB:
Resolves: #2817


crohr:
@iCreateJB I don't think this fixes #2817, since that issue is related to the convox login phase, not docker login.

iCreateJB:
Thanks @crohr, I believe I have the correct login function now.

iCreateJB:
Thoughts @ddollar ?

PR #2989, opened 2018-10-19
use new amazon linux 2 amis

ddollar:


provider/aws/formation/rack.json
```diff
@@ -603,11 +598,6 @@
       "Description": "Default swap volume size in GB",
       "Default": "5"
     },
-    "RootSize": {
```
ffjia: @ddollar the `RootSize` can be kept, and use it to enlarge the root device? And update the `Default` to 30.

PR #2993, opened 2018-10-24
use t2 instances for rack defaults

ddollar:


mwarkentin:
Seeing issues w/ the t3?

ddollar:
Not all regions have t3 yet

PR #2996, opened 2018-10-25
optionally add RACK_URL to application environment

ddollar:


mwarkentin:
What’s the idea here? Allow apps to log in and interact w/ the convox rack?

ddollar:
Yep

PR #3024, opened 2018-11-20
Update Postgres template parameters

denmat:
Previously we have been able to specify the database name and backup retention like here:
https://github.com/convox/rack/blob/6f95ee8c1e3a03642340ff0e8b42088253bbcd5f/provider/aws/templates/resource/postgres.tmpl#L16-L20

We wish to add the same ability to describe a postgres database via the `convox.yml`.

```
resources:
  event-db:
    type: postgres
    options:
      database: events
      class: db.t2.xlarge
      storage: 150
      durabilty: True
      version: 9.6.10
      backup_retention_period: "30"
      master_username: auser
```

## Change
* Add parameter `Database` that maps to `DBName`
* Add parameter `BackupRetentionPeriod`
* Add parameter `MasterUsername`

## Consideration
We map the `Database` to `DBName` because we did it that way in the previous template (linked above). So this naming was to be consistent with that.

Haven't been able to test this in a development environment so I hope this is all we need.

### DIFF
--- a/provider/aws/formation/resource/postgres.json.tmpl
+++ b/provider/aws/formation/resource/postgres.json.tmpl
@@ -5,10 +5,6 @@
     "Version9": { "Fn::Equals": [ { "Fn::Select": [ 0, { "Fn::Split": [ ".", { "Ref": "Version" } ] } ] }, "9" ] }
   },
   "Parameters": {
-    "BackupRetentionPeriod": {
-      "Type": "String",
-      "Default": "1"
-    },
     "Class": {
       "Type": "String",
       "Default": "db.t2.micro"
@@ -82,7 +78,7 @@
       "DeletionPolicy": "Snapshot",
       "Properties": {
         "AllocatedStorage": { "Ref": "Storage" },
-        "BackupRetentionPeriod": { "Ref": "BackupRetentionPeriod" }
+        "BackupRetentionPeriod": "30"
         "DBInstanceClass": { "Ref": "Class" },
         "DBInstanceIdentifier": { "Ref": "AWS::StackName" },
         "DBName": { "Ref": "Database" },
--- a/provider/aws/formation/resource/postgres.json.tmpl
+++ b/provider/aws/formation/resource/postgres.json.tmpl
@@ -40,7 +40,7 @@
       "Type": "Number",
       "Default": "20"
     },
-    "MasterUsername": {
+    "Username": {
       "Type": "String",
       "Default": "app"
     }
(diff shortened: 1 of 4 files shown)


provider/aws/formation/resource/postgres.json.tmpl
```diff
@@ -65,15 +77,16 @@
       "DeletionPolicy": "Snapshot",
       "Properties": {
         "AllocatedStorage": { "Ref": "Storage" },
+        "BackupRetentionPeriod": { "Ref": "BackupRetentionPeriod" }
```
ddollar: Let's just default this to 30 instead of parameterizing it.

provider/aws/formation/resource/postgres.json.tmpl
```diff
@@ -31,13 +39,17 @@
       "Type": "Number",
       "Default": "20"
     },
+    "MasterUsername": {
```
ddollar: Is there any real advantage to parameterizing this since the result comes out in the URL?

provider/aws/formation/resource/postgres.json.tmpl
```diff
@@ -5,10 +5,18 @@
...
   "Parameters": {
+    "BackupRetentionPeriod": {
+      "Type": "String",
+      "Default": "1"
+    },
     "Class": {
       "Type": "String",
       "Default": "db.t2.micro"
     },
+    "Database": {
```
ddollar: Is there any real advantage to parameterizing this since the result comes out in the URL?

denmat: We have some apps that we will dump data into, which have DB names already declared. This is for apps that we migrating and may have external clients that also connect to DBs.

provider/aws/formation/resource/postgres.json.tmpl
denmat: We have some apps that we will dump data into, which have roles already declared. This is for apps that we migrating and may have external clients that also connect to DBs.

ddollar: This one should be called `Username`

ddollar:
The `k8s-local` branch has landed. If you'd like to land this please rebase onto master and make the following updates:

* These parameters should be implemented in the local [postgres resource](https://github.com/convox/rack/blob/master/provider/local/template/resource/postgres.yml.tmpl)
* These parameters should be implemented in the local [mysql resource](https://github.com/convox/rack/blob/master/provider/local/template/resource/mysql.yml.tmpl)
* These parameters should be implemented in the aws [mysql resource](https://github.com/convox/rack/blob/master/provider/aws/formation/resource/mysql.json.tmpl)

denmat:
@ddollar I've updated the files as suggested (can you please check that I have updated the local files correctly!).

Also I noticed with the mysql local resource you recently made this change: https://github.com/convox/rack/pull/3090.

These changes are going to go against that. Is that your intention?

ddollar:
Closing this for now as it doesn't seem maintanable in the long term, especially compared to the demand for it.

PR #3025, opened 2018-11-26
Allow users to not create user and keys for SQS

denmat:
We would like to not create users in IAM.

  * We prefer to create roles instead of users
  * Keys and secrets are exposed
  * Defaults to `True` since it would probably break existing installations.

This should allow the following command to execute to not create IAM entities:

```
convox resources create sqs --name myqueue CreateUsersAndKeys=False
```


ddollar:
These types of resources are meant to be totally self-contained with their final URL. If you need a simpler SQS resource it would be better to create it through alternative means.

Issue #3013, opened 2018-11-06
Noisy Health Checks on local rack


slaskis:
I find that the health checks when running `convox start` in g2 is very noisy, especially when an app has a bunch of services it takes up a large part of the screen in the logs.

I could just `grep -v` to strip it all but since its useful when booting and reloading I would like to suggest that it only logs once if its ok and always when not ok, which would make it look something like this:

```
convox   | health check api: 502
api | 172.17.0.1 - - [05/Nov/2018 19:08:22] "GET /x HTTP/1.1" 502 -
api | 172.17.0.1 - - [05/Nov/2018 19:08:22] "GET /x HTTP/1.1" 502 -
convox   | health check api: 502
api | 172.17.0.1 - - [05/Nov/2018 19:08:22] "GET /x HTTP/1.1" 502 -
convox   | health check api: 502
api | 172.17.0.1 - - [05/Nov/2018 19:08:22] "GET /x HTTP/1.1" 502 -
convox   | health check api: 502
api | 172.17.0.1 - - [05/Nov/2018 19:08:22] "GET /x HTTP/1.1" 200 -
convox   | health check api: 200 
api | 172.17.0.1 - - [05/Nov/2018 19:08:22] "GET /x HTTP/1.1" 200 -
api | 172.17.0.1 - - [05/Nov/2018 19:08:22] "GET /x HTTP/1.1" 200 -
api | 172.17.0.1 - - [05/Nov/2018 19:08:22] "GET /x HTTP/1.1" 200 -
api | 172.17.0.1 - - [05/Nov/2018 19:08:22] "GET /x HTTP/1.1" 200 -
api | 172.17.0.1 - - [05/Nov/2018 19:08:22] "GET /x HTTP/1.1" 200 -
api | 172.17.0.1 - - [05/Nov/2018 19:08:22] "GET /x HTTP/1.1" 502 -
convox   | health check api: 502
convox   | health check api: 502
api | 172.17.0.1 - - [05/Nov/2018 19:08:22] "GET /x HTTP/1.1" 200 -
convox   | health check api: 200 
```

ddollar:
Going to close this issue but I would accept this as a PR.

PR #3031, opened 2018-12-03
reduce noise from healthchecks for convox start

slaskis:
a PR for #3013 

the test is pretty slow as the shortest healhcheck i could get running was with 1s interval and since i wanted to test the transition from and to ok status it takes a few attempts.


ddollar:
Carrying forward in #3032 (reduced the delay in the testing)

slaskis:
beautiful, thank you!

PR #3041, opened 2018-12-12
WIP: Add Stockholm region

zeeshanabid:
Stockholm region was just released and I would like to add it to the convox as well.


zeeshanabid:
@slaskis Do you know when it will be available in Stockholm region?

ddollar:
Carrying this in #3058

PR #3051, opened 2018-12-16
include query string in websocket proxy

jayjun:
Phoenix passes `?vsn=2.0.0` to establish protocols in WebSocket connections.

According to the [standard](https://tools.ietf.org/id/draft-abarth-thewebsocketprotocol-00.html#rfc.section.3.1), WebSocket URLs include query parameters.


ddollar:
Thanks!

PR #3054, opened 2018-12-21
Don't use minor version in Postgres EngineVersion

denmat:
Having a minor version that is not regularly updated could make systems open to vulnerabilities. People should immediately update a `9.6.3` instance. 

I would err on the side of auto-update unless there is a specific reason to pin to a version. Maybe this issue of automatic minor version updates can be highlighted in the docs so that people understand the risks? 

 * this will default to latest minor version
  * people may not be aware that this minor version requires immediate updating
  * people need to be aware that this can trigger automatic version updates
    * with single instances (not multi-az) this can cause unexpected outage

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-engineversion

> To prevent automatic upgrades, be sure to specify the full version number (for example, 5.6.13). If the default version for the database engine changes and you specify only the major version (for example, 5.6), your DB instance will be upgraded to use the latest default version.

### DIFF
--- a/provider/aws/formation/resource/postgres.json.tmpl
+++ b/provider/aws/formation/resource/postgres.json.tmpl
@@ -33,7 +33,7 @@
     },
     "Version": {
       "Type": "String",
-      "Default": "9.6.6"
+      "Default": "9.6"
     }
   },
   "Outputs": {
--- a/provider/aws/templates/resource/postgres.tmpl
+++ b/provider/aws/templates/resource/postgres.tmpl
@@ -41,7 +41,7 @@
       },
       "EngineVersion": {
         "Description": "Version of Postgres",
-        "Default": "9.6.3",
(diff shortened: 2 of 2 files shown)


daveli:
@ddollar This change seems to break CloudFormation. As you are updating the version number CloudFormation will try to replace the DB instance as it considers it a downgrade, even though this isn't the case in practice, resulting in the following error: 

Embedded stack {arn} not successfully updated. Currently in UPDATE_ROLLBACK_IN_PROGRESS with reason: The following resource(s) failed to update: [Instance].

daveli:
That is a good question as it doesn't seem to happen in all cases. I have upgraded about 10+ apps without issues. Consider this to be "documentation" for when other potentially run into this issue as well.

PR #3058, opened 2019-01-08
support for eu-north-1

ddollar:


chids:
@ddollar shouldn't there also be a change wrt SQS FIFO for this to work?

PR #3077, opened 2019-01-22
Added convox env set-file option

radekl:
convox env set-file option replaces entirely environment variables and loads them from given file.

### DIFF
--- a/pkg/cli/env.go
+++ b/pkg/cli/env.go
@@ -51,6 +51,18 @@ func init() {
 		Usage: "<key=value> [key=value]...",
 	})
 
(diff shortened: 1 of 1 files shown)


ddollar:
I think this might work better as an option to `convox env set`. 

`set-file` doesn't really describe how it differs from `convox env set`. You can already do `cat .env | convox env set` which seems to duplicate this command as currently named. (The replacement behavior is not obvious from the name)

Instead you could add a `--replace` option to `convox env set`.

This would allow both of the following to work:

`convox env set FOO=bar -a myapp --replace`

`cat .env | convox env set -a myapp --replace`

ddollar:
Closing in favor of #3078

PR #3078, opened 2019-01-23
Added --replace flag for convox env set

radekl:


pkg/cli/env.go
```diff
@@ -188,10 +196,7 @@ func EnvSet(rack sdk.Interface, c *stdcli.Context) error {
 		c.Writer().Stdout = c.Writer().Stderr
 	}
 
-	env, err := helpers.AppEnvironment(rack, app(c))
-	if err != nil {
-		return err
-	}
+	env, err := loadEnvs(rack, c, c.Bool("replace"))
```
ddollar: This is no longer checking the error return value.

I think you don't need the extra function here, maybe something like:

```golang
env := structs.Environment{}
var err error

if !c.Bool("replace") {
  env, err = helpers.AppEnvironment(rack, app(c))
  if err != nil {
    return err
  }
}

File: pkg/cli/env.go [hunk 2]
```diff
@@ -46,6 +46,7 @@ func init() {
 			flagId,
 			flagRack,
 			flagWait,
+			flagReplace,
```
ddollar: Sorry one more thing. The `flag*` globals are really only meant for variables that are used in more than one command. Could you please change this to look more like the promote flag below (and alphabetize below it)

radekl:
Fixed both tests. Strange thing that before tests were fine.

ddollar:
Rebased as #3085

PR #3097, opened 2019-01-27
allow adding additional instance policy

ddollar:


provider/aws/formation/rack.json
```diff
@@ -502,6 +503,11 @@
       "MinValue": "1",
       "Type": "Number"
     },
+    "InstancePolicy": {
+      "Default": "",
+      "Description": "ARN of an additional IAM policy to add to the cluster instances",
+      "Type": "String"
```
mwarkentin: Does it make any sense to make this a list so you could attach multiple (more granular) policies?

ddollar: I can't think of an easy way to merge a user-specified list with a static list inside the template.

PR #3119, opened 2019-02-12
Moving lambda to VPC when Rack is private

radekl:
This is, I hope, first step to hide Enterprise Convox console.

Those changes allows webhooks resources to be sent from NAT IPs.


ddoc:
superceded by #3504

PR #3134, opened 2019-02-22
No blocking http requests in router

slaskis:
While inspecting why the local k8s rack on gen2 is still unusable for us I might have a new theory.

It seems like when our next.js dev server is compiling, which sometimes can take up to a minute, it blocks the router. This in turn makes all the healthchecks fail in kubernetes which causes everything to restart and behave generally bad. Next request is the same as the cache is now cleared because of the restart etc...

Is it possible that this PR could fix it?

### DIFF
--- a/pkg/router/http.go
+++ b/pkg/router/http.go
@@ -58,11 +58,11 @@ func (h *HTTP) Serve() error {
 
 func (h *HTTP) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	if h.Handler != nil {
-		h.Handler(w, r)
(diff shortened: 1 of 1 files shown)


ddollar:
I don't believe this is going to solve it as the Go HTTP stack is already channel-based.

If you can put together a simple test case that demonstrates your issue as a repo that I can `convox start` I'd be happy to take a look.

slaskis:
Damn. I was hoping I was on to something...

Right, I'll see if I can get a smaller test setup up and running...

PR #3218, opened 2019-04-12
allow switching rack router to main alb

ddollar:


ddollar:
[rack router on ALB for podium](https://trello.com/c/DYnszNtb/6-rack-router-on-alb-for-podium)

PR #3221, opened 2019-04-16
read parameters overrides from convox.yml

ddollar:
Allows setting parameter overrides inside `convox.yml`

```
params:
  RedirectHttps: no
services:
  web:
    build: .
    port: 3000
```

**Note: Parameters specified in this way will be applied during every release promotion, potentially changing any values you have set manually with `convox apps params set`**


robatwave:
@ddollar should `no` be `No`?

ddollar:
[HttpsRedirect migration for wave](https://trello.com/c/EfD0kibM/5-httpsredirect-migration-for-wave)

PR #3225, opened 2019-04-22
use lowercase origin for hostname matching

ddollar:


ddollar:
[origin hostname matching](https://trello.com/c/SiuJxXCN/71-origin-hostname-matching)

PR #3226, opened 2019-04-30
make release promotion events more resilient

ddollar:


ddollar:
[release promotion events for schibsted](https://trello.com/c/XLLvYPIb/67-release-promotion-events-for-schibsted)

PR #3229, opened 2019-04-30
eks support

ddollar:
This provider should currently be considered **ALPHA** quality and should not be used for anything important.


ddollar:
[kaws](https://trello.com/c/Sm2qrXPq/70-kaws)

PR #3230, opened 2019-05-01
terminate newest instance during scale down to minimize effect on exi…

ddollar:
…sting containers


ddollar:
[terminate newest instance first during scaledown](https://trello.com/c/Nxy8RUp3/72-terminate-newest-instance-first-during-scaledown)

idris:
Won't this cause a lot of processes to churn? I imagine new process will be started on the new instances, then old processes will be shut down, then the new instances will drain and the new processes will have to move over to the old instances.

PR #3232, opened 2019-05-02
ensure version is stored as a string

ddollar:


ddollar:
[numeric version causing local rack issues](https://trello.com/c/Ug4oQOvD/74-numeric-version-causing-local-rack-issues)

PR #3235, opened 2019-05-03
ensure dns and socket settings persist on local rack

ddollar:


ddollar:
[local provider socket persistence](https://trello.com/c/AC3WLCte/83-local-provider-socket-persistence)

PR #3242, opened 2019-05-06
dynamic terminal size for k8s exec

ddollar:


ddollar:
[kaws exec terminal wrapping](https://trello.com/c/Qtzlwlwn/78-kaws-exec-terminal-wrapping)

PR #3244, opened 2019-05-06
update internal docker version and include tests

ddollar:


ddollar:
[use correct docker version for tests](https://trello.com/c/407vblHl/90-use-correct-docker-version-for-tests)

PR #3245, opened 2019-05-06
prevent race during certificate deletion

ddollar:


ddollar:
[prevent race during certificate deletion](https://trello.com/c/bqwlt8fs/93-prevent-race-during-certificate-deletion)

PR #3262, opened 2019-06-05
AWS: Parameter Store - Adds the ability to use secrets

benvinson-teespring:
In order to be able to leverage Parameter Store in Systems Manager, the following must be added to the container definition in the taskdef (as in provider/aws/formation/service.json.tmpl):
```
"secrets": [
        {
          "valueFrom": "PS-Name",
          "name": "VARIABLE_NAME"
        }
]
```

Considering I have _no_ experience in Go, I'm surprised I could get the tests to pass.


agonbina:
Why is this not merge yet? It would be great to have this feature

PR #3278, opened 2019-07-04
Move internal LB onto private subnets if available

mwarkentin:
This matches how the internal LBs for gen1 apps function: https://github.com/convox/rack/blob/master/provider/aws/formation/g1/app.json.tmpl#L469-L475


ddollar:
Carried forward in #3279, thanks!

PR #3284, opened 2019-07-29
restore ContainerDisk rack parameter for controlling container size

ddollar:


swisslala:
@ddollar can we get this merged in - it broke our setup .. again - Haven.

virtualdom:
@ddollar what happened to this? We're currently not able to set our `ContainerDisk` for an application and we're hitting storage limits on startup.

```
$ convox version
client: 20191110173914
server: 20191216210003 (console.convox.com)
```

ddollar:
This change needs to be significantly reworked to handle the new Amazon Linux 2 instances and how they configure ECS. Please submit a ticket with some details on the error you are seeing and we'll slot it into an upcoming iteration.

PR #3297, opened 2019-08-29
pkg/cli: Fix dropped error

alrs:
This PR fixes a dropped error in `cli`.

### DIFF
--- a/pkg/cli/logs.go
+++ b/pkg/cli/logs.go
@@ -36,5 +36,5 @@ func Logs(rack sdk.Interface, c *stdcli.Context) error {
 
 	_, err = io.Copy(c, r)
 
-	return nil
+	return err
 }


ddollar:
Can you please provide a little bit of context about the problem you're trying to solve?

I believe this PR in its current form will exit with an io.EOF error every time.

ddollar:
Actually looks like `io.Copy` will not return `io.EOF` when encounters. Would still be helpful to know the problem you're encountering.

alrs:
You're assigning an error variable and never logging or returning it, which reads to me like a bug.

ddollar:
Was this discovered as a result of an issue that you encountered or some sort of automated linting?

alrs:
No issue, just linting. Dropped `err` variables are the most frustrating of all bugs in Go software, so I hunt them actively.

ddollar:
Appreciate the PR but seems like there could be side-effects and going to stick with the "if it ain't broke..." here.

PR #3302, opened 2019-09-17
update ecs amis

ddollar:


provider/aws/formation/rack.json
```diff
@@ -68,23 +68,22 @@
   },
   "Mappings": {
     "RegionConfig": {
-      "ap-east-1":      { "Ami": "ami-01cb1066e1ad93cba", "EFS": "No",  "ThirdAvailabilityZone": "Yes", "ELBAccountId": "754344448648", "Fargate": "No"  },
```
beastawakens: Hmmm, `ap-east-1` is not in the new json... Is that because of the new opt-in requirement for that and `me-south-1`?

ddollar: Fought with this for a bit, but STS/auth is different/difficult with those two regions. Will revisit if necessary.

PR #3310, opened 2019-10-07
scope FilterLogEvents to relevant log streams

ddollar:


mwarkentin:
Any idea what sort of volume of log calls we might see reduced by this? Iirc log stream requests were extremely frequent in our cloudtrail.

ddollar:
I doubt you'll see a reduction in the number of API calls, this really just scopes the amount of work being done behind those API calls to make the responses more performant.

PR #3327, opened 2019-11-07
metrics enhancement

ddollar:
Reconfigures AWS metrics to use `GetMetricData`. 

Uses metric expressions to fix `service:requests` to be total requests and not requests-per-target.

### New App Metrics

* `process:healthy` (avg, min, max)
* `process:unhealthy` (avg, min, max)
* `service:requests:2xx` (sum)
* `service:requests:3xx` (sum)
* `service:requests:4xx` (sum)
* `service:requests:5xx` (sum)
* `service:response:time` (min, max)


camerondgray:
LGTM

PR #3328, opened 2019-11-07
Extra logging around the logging

beastawakens:


pkg/helpers/aws.go
```diff
@@ -347,6 +347,8 @@ func CloudWatchLogsSubscribe(ctx context.Context, cw cloudwatchlogsiface.CloudWa
 func CloudWatchLogsStream(ctx context.Context, cw cloudwatchlogsiface.CloudWatchLogsAPI, w io.WriteCloser, group, stream string, opts structs.LogsOptions) error {
 	defer w.Close()
 
+	fmt.Printf("CloudWatchLogsStream group=%q stream=%q", group, stream)
```
ddollar: Should generate some sort of random ID at the top here and then thread it through all of the following logs

pkg/helpers/aws.go
```diff
@@ -384,25 +386,31 @@ func CloudWatchLogsStream(ctx context.Context, cw cloudwatchlogsiface.CloudWatch
 
 	sleep := time.Duration(100 * time.Millisecond)
 
+	fmt.Printf("StartTime=%d\n", req.StartTime)
```
ddollar: Should keep the same prefixes all the way through, otherwise these lines can be tough to tell what they're associated with

ddollar:
Pulling this from the release for now. Was very verbose and caused a compounding cycle of watching logs causing more logging.

PR #3352, opened 2019-12-14
Rack param: ImagePullBehavior

mattbriancon:
Allow racks to configure ECS_IMAGE_PULL_BEHAVIOR for the ECS agent. https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-agent-config.html

I can't think of a reason not to do this. The default behavior is to pull the image every time but since image tags are NOT reused in Convox, it's safe to use the cache. Leaving the cache behavior as "opt-in" so there are no surprises but I'd encourage folks with large images to make the switch.

### DIFF
--- a/provider/aws/formation/rack.json
+++ b/provider/aws/formation/rack.json
@@ -502,6 +502,12 @@
       "Type": "String",
       "Default": ""
     },
+    "ImagePullBehavior": {
+      "Type": "String",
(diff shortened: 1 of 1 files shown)


mattbriancon:
Not sure how to test this but happy to do whatever is needed to get this over the line.

ryanwilsonperkin:
Sounds like a great approach to me. Several of our images are >1gb, it would be nice to avoid that pull especially for services with a high scale count

ddollar:
Will go out in the next Rack release, thanks!

PR #3361, opened 2020-02-21
Disable auth for external registries when PrivateBuild is enabled

beastawakens:
ECR registries can't be locked down to IP ranges so aren't affected, only normal registries.
Also corrected the label for the 'Private' parameter.

### DIFF
--- a/provider/aws/registries.go
+++ b/provider/aws/registries.go
@@ -51,16 +51,15 @@ func (p *Provider) RegistryAdd(server, username, password string) (*structs.Regi
 		if _, _, err := p.authECR(server, username, password); err != nil {
 			return nil, fmt.Errorf("unable to authenticate")
 		}
+	case p.PrivateBuild:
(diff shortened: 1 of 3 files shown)


provider/aws/registries.go
```diff
@@ -52,13 +52,15 @@ func (p *Provider) RegistryAdd(server, username, password string) (*structs.Regi
...
 		}
 	default:
-		_, err := dc.AuthCheck(&docker.AuthConfiguration{
-			ServerAddress: server,
-			Username:      username,
-			Password:      password,
-		})
-		if err != nil {
-			return nil, fmt.Errorf("unable to authenticate")
+		if !(p.PrivateBuild) {
```
ddollar: This should probably just be the first case of the "validate login" switch with no body

ddollar:
might want to add a comment in the blank body section about the special case

PR #3367, opened 2020-03-31
Filter invalid headers in ALB

rabidaudio:
This updates the rack CF template to support a rack parameter that toggles the [ALB `routing.http.drop_invalid_header_fields.enabled` attribute](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#load-balancer-attributes).

### DIFF
--- a/provider/aws/formation/rack.json
+++ b/provider/aws/formation/rack.json
@@ -2148,7 +2148,7 @@
           { "Key": "access_logs.s3.bucket", "Value": { "Fn::If": [ "BlankLogBucket", { "Ref": "Logs" }, { "Ref": "LogBucket" } ] } },
           { "Key": "access_logs.s3.prefix", "Value": { "Fn::Sub": "convox/logs/${AWS::StackName}/alb" } },
           { "Key": "idle_timeout.timeout_seconds", "Value": "3600" },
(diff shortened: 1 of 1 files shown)


ddollar:
I think w can drop the Rack parameter and just turn this on. We can always add a parameter is someone needs it for some reason.

rabidaudio:
Got it. updated

provider/aws/formation/rack.json
```diff
@@ -2147,7 +2147,8 @@
           { "Key": "access_logs.s3.enabled", "Value": "true" },
           { "Key": "access_logs.s3.bucket", "Value": { "Fn::If": [ "BlankLogBucket", { "Ref": "Logs" }, { "Ref": "LogBucket" } ] } },
           { "Key": "access_logs.s3.prefix", "Value": { "Fn::Sub": "convox/logs/${AWS::StackName}/alb" } },
-          { "Key": "idle_timeout.timeout_seconds", "Value": "3600" }
+          { "Key": "idle_timeout.timeout_seconds", "Value": "3600" },
+          { "Key": "routing.http.drop_invalid_header_fields.enabled", "true" }
```
beastawakens, line 2151: This should be `"Value": "true"` rather than just `"true"`.

PR #3374, opened 2020-04-24
Enable configuration of logging within the InstancesAutoscaler

beastawakens:


provider/aws/lambda/autoscale/handler.go
```diff
@@ -314,6 +314,13 @@ func ci(ii ...*int64) int64 {
 	return 0
 }
 
+func log(format string, a ...interface{}) {
+	if logging, present := os.LookupEnv("SERVICE_SCALING_LOGGING"); !present || present && logging != "No" {
```
ddollar, line 318: let's call this var `DEBUG` and just look for the string `true`

provider/aws/formation/rack.json
```diff
@@ -633,6 +633,11 @@
       "Description": "The security groups (comma delimited) to assign to the rack router.",
       "Type": "CommaDelimitedList"
     },
+    "ServiceScalingLogging": {
```
ddollar, line 636: Let's not even make this a rack param. If someone really needs to debug this function (unlikely) they can manually set the env var on the Lambda function.

provider/aws/lambda/autoscale/handler.go
```diff
@@ -314,6 +314,13 @@ func ci(ii ...*int64) int64 {
 	return 0
 }
 
+func log(format string, a ...interface{}) {
```
ddollar, line 317: Make two functions, `log` and `debug`.

`debug` should only write out if `DEBUG=true`

Change all other Printfs to use log which should always print

provider/aws/lambda/autoscale/handler.go
```diff
@@ -282,9 +282,9 @@ func desiredCapacity(largest, total *Metrics) (int64, error) {
 	// the number of instances over the amount required to fit the widest service (gen1 loadbalancer-facing)
 	extraWidth := totalCount - largest.Width
 
-	fmt.Printf("extraCapacity = %+v\n", extraCapacity)
-	fmt.Printf("extraFit = %+v\n", extraFit)
-	fmt.Printf("extraWidth = %+v\n", extraWidth)
+	log("extraCapacity = %+v\n", extraCapacity)
```
ddollar, line 285: seems like a lot of these should be `debug()` instead of `log()`

beastawakens: Ah, I think I misunderstood from your previous comments.  I thought you intended to chain the functions together... I'll update so that most are calling debug directly.

PR #3388, opened 2020-08-14
rack: ability to specify LB algorithm

beastawakens:


beastawakens:
> looks good but perhaps we should update: https://github.com/convox/docs/blob/master/docs/reference/app-parameters.md as well

https://github.com/convox/docs/pull/105 😉

PR #3393, opened 2020-08-26
cli: Make noop update checks faster

beastawakens:


pkg/cli/update.go
```diff
@@ -31,6 +32,11 @@ func Update(rack sdk.Interface, c *stdcli.Context) error {
 		target = v
 	}
 
+	if (target == current) {
+		c.Writef("No update to be performed... ")
```
ddollar, line 36: Let's be even more explicit, print `Already on the latest version <release>NUM</release>` and return nil instead of OK() 

(the ... OK pattern is for an action that needs time to complete and can fail with ERROR instead of OK.)

PR #3394, opened 2020-08-26
Resource mariadb support

beastawakens:


provider/local/template/resource/mariadb.yml.tmpl
```diff
@@ -0,0 +1,99 @@
...
+    type: resource
+    resource: {{.Name}}
+data:
+  URL: "mysql://root:{{.Password}}@resource-{{.Name}}.{{.Namespace}}.svc.cluster.local:3306/app"
+---
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  namespace: {{.Namespace}}
+  name: resource-{{.Name}}-mysql
```
ddollar, line 19: This should end in `mariadb`

provider/local/template/resource/mariadb.yml.tmpl
```diff
@@ -0,0 +1,99 @@
...
+kind: Deployment
+metadata:
+  namespace: {{.Namespace}}
+  name: resource-{{.Name}}
+  labels:
+    system: convox
+    rack: {{.Rack}}
+    app: {{.App}}
+    type: resource
+    kind: mysql
```
ddollar, line 42: Should be `mariadb`

PR #3402, opened 2020-09-30
Add process type labels for services and timers

beastawakens:


ddollar (changes requested):
Looks good, just both entries missing a space to match the surrounding formatting.

ddollar (approved):

PR #3427, opened 2021-01-25
Volume gp3 update

beastawakens:


beastawakens:
Uurgh, so gp3 is not currently available for Auto-Scaling groups with Launch Configurations 🙄

liamdawson:
> not currently available for Auto-Scaling groups with Launch Configurations

It won't be. See https://docs.aws.amazon.com/autoscaling/ec2/userguide/LaunchConfiguration.html

> **Important**
>
> We strongly recommend that you do not use launch configurations. They do not provide full functionality for Amazon EC2 Auto Scaling or Amazon EC2. We provide information about launch configurations for customers who have not yet migrated from launch configurations to launch templates.

mwarkentin:
@beastawakens looks like they decided to fix it for you. 😁

image

PR #3453, opened 2021-03-21
Change to use SSM to pull up-to-date AMI's and other information auto…

beastawakens:
…matically.  Adds support for CN regions

### DIFF
--- a/pkg/cli/rack.go
+++ b/pkg/cli/rack.go
@@ -387,6 +387,15 @@ func RackUpdate(rack sdk.Interface, c *stdcli.Context) error {
 		if s.Version == "dev" {
 			target = "dev"
 		} else {
+			latest, err := cv.Latest()
+			if err != nil {
+				return err
+			}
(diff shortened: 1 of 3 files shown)


fk-currenxie:
@beastawakens I see some mentions of ARM and t4g instances. Will that be supported? Will the rack Docker image targeting ARM be part of this as well?

beastawakens:
> @beastawakens I see some mentions of ARM and t4g instances. Will that be supported? Will the rack Docker image targeting ARM be part of this as well?

@fk-currenxie This is beginning to introduce support for those instance types yes.  The Rack image running on ARM won't be part of this PR but is still on the roadmap 😊

Twsouza:
In the current state, once the rack is installed, we can't update the AMI version. We need to be able to update the AMI version after instalation.

Twsouza:
Done by https://github.com/convox/rack/pull/3553

PR #3467, opened 2021-09-22
Internal DNS records based on Rack status, not app status

beastawakens:


beastawakens:
This covers a use case where on a Rack with mixed internal and public services, the 'local' DNS record for an app/service ([service].[app].[rack].convox) is directed at the public load balancer, so the internal services can't access it.

This change checks for the presence of an internal load balancer for the Rack and if it exists, the DNS record will now be pointed at that.

PR #3481, opened 2021-11-17
Add support to enable/disable High Availability when creating racks

heronrs:
This adds a new formation parameter named `HighAvailability`. it's `true` by default and setting it to `false` will reduce the number of subnets/auto-scaling replicas and a few other resources.

### DIFF
--- a/pkg/cli/rack.go
+++ b/pkg/cli/rack.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 	"io"
 	"net/url"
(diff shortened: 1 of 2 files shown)


moisesrodriguez:
Hello, for this new functionality. It would be great to be able to choose the number of subnets/replicas. Example, a dev environment, one replica in just one AZ would be more than enough.

heronrs:
Thanks for the suggestion @moisesrodriguez! You'll still be able to configure the number of instances using the `InstanceCount` parameter. Regarding subnets, we need to create at least two because ALB(used to route requests to services) requires at least two subnets.

beastawakens (approved):
🚢

PR #3482, opened 2021-11-24
Fix inconsistent vendoring error when building the cli

heronrs:
Tidy up go modules and install `xgo` without modifying `go.mod` to avoid `inconsistent vendoring` errors when building the cli.

Edit: build was  successfully created https://github.com/convox/rack/releases/tag/20211124183359-fix-build

### DIFF
--- a/cmd/convox/Makefile
+++ b/cmd/convox/Makefile
@@ -19,7 +19,7 @@ clean:
 package: package-deps package-gopath $(pkg_darwin) $(pkg_linux) $(pkg_windows) package-export
 
 package-deps:
-	go get -u github.com/crazy-max/xgo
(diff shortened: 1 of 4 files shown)


cmd/convox/Makefile
```diff
@@ -19,7 +19,7 @@ clean:
 package: package-deps package-gopath $(pkg_darwin) $(pkg_linux) $(pkg_windows) package-export
 
 package-deps:
-	go get -u github.com/crazy-max/xgo
+	go install github.com/crazy-max/xgo@v0.7.5
```
heronrs, line 22: install the `xgo` binary without modifying `go.mod`

PR #3483, opened 2021-12-06
User can set a schedule to shut down and up the cluster

Twsouza:
User can pass a CRON with ScheduledActionDownCron and ScheduledActionUpCron to set the Schedule Action

Example:

To shutdown the cluster on Friday 6pm (UTC) until Monday 9am (UTC)

```
convox rack install aws -n v2  ScheduleRackScaleDown="0 18 * * 5" ScheduleRackScaleUp="0 9 * * 1"
```

Or update the schedule:

```
convox rack params set ScheduleRackScaleDown="0 18 * * 5" ScheduleRackScaleUp="0 9 * * 1"
```

### DIFF
--- a/provider/aws/formation/rack.json
+++ b/provider/aws/formation/rack.json
@@ -113,6 +113,12 @@
     "RegionHasEFSAndThirdAvailabilityZoneAndHighAvailability": {
       "Fn::And": [ { "Condition": "RegionHasEFS" }, { "Condition": "ThirdAvailabilityZone" }, { "Condition": "HighAvailability" } ]
     },
+    "SetScheduleRackScale": { "Fn::Not" : [{
+      "Fn::And": [
+        { "Fn::Equals": [ { "Ref": "ScheduleRackScaleDown" }, "" ] },
+        { "Fn::Equals": [ { "Ref": "ScheduleRackScaleUp" }, "" ] }
+      ]}
+    ] },
     "SpotInstances": { "Fn::Not": [ { "Fn::Equals": [ { "Ref": "SpotInstanceBid"}, "" ] } ] },
     "SwapEnabled": { "Fn::Not": [ { "Fn::Equals": [ { "Ref": "SwapSize" }, "0" ] } ] },
     "ThirdAvailabilityZone": { "Fn::And": [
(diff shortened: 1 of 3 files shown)


heronrs:
Awsome! Can you please make sure we add this new param to v2 docs?
It be good to also add some example, like setting up a scheduled from mon-fri up and sat-sun down.

Repo https://github.com/convox/docs
Docs: https://docsv2.convox.com/introduction/getting-started

provider/aws/formation/rack.json
```diff
@@ -726,6 +728,16 @@
...
       "Type": "CommaDelimitedList"
     },
+    "ScheduledActionDownCron": {
+      "Type": "String",
+      "Description": "The recurring schedule for when the rack will be shutdown. This format consists of five fields separated by white spaces: [Minute] [Hour] [Day_of_Month] [Month_of_Year] [Day_of_Week]",
+      "Default": ""
+    },
+    "ScheduledActionUpCron": {
+      "Type": "String",
+      "Description": "The recurring schedule for when the rack will be shutdown. This format consists of five fields separated by white spaces: [Minute] [Hour] [Day_of_Month] [Month_of_Year] [Day_of_Week]",
```
heronrs, line 738: `will start`  instead of  `will be shutdown` for this one.

provider/aws/formation/rack.json
```diff
@@ -1722,6 +1734,49 @@
         }
       }
     },
+    "ScheduledActionDown":{
+      "Type":"AWS::AutoScaling::ScheduledAction",
+      "Condition": "SetScheduledActionDown",
+      "Properties":{
+        "AutoScalingGroupName": { "Ref": "Instances" },
```
heronrs, line 1741: should we add `"DesiredCapacity": "0",` to this one aswell?

heronrs (approved):
LGTM! let's just address the param name changes we discussed earlier.

pkg/helpers/aws.go
```diff
@@ -99,6 +99,11 @@ func CloudformationInstall(cf cloudformationiface.CloudFormationAPI, name, templ
 		TemplateURL: aws.String(template),
 	}
 
+	sadown, saup := params["ScheduledActionDownCron"], params["ScheduledActionUpCron"]
+	if (sadown == "" || saup == "") && (sadown != "" || saup != "") {
```
heronrs, line 103: What do you think about adding this validation to https://github.com/convox/rack/blob/master/pkg/cli/rack.go#L114
And the update one to https://github.com/convox/rack/blob/master/pkg/cli/rack.go#L233.

I don't think we need to go much further on the code if the params are incorrectly set.

provider/aws/formation/rack.json
Twsouza: we have `"MaxSize": "0"` and `MinSize": "0"` set to zero, `DesiredCapacity` is not needed.

heronrs (approved):
LGTM. great job!

PR #3490, opened 2022-01-19
Use IMDSv2 calls and a parameter to use only v2

Twsouza:
The default will be the "optional" parameter, so any user can use the v1 or v2. If set to "required" it will use only v2.

image

For example, to create a rack with IMDSv2 only, you'd run:

```
convox rack install aws --name dev-v2-1531 IMDSHttpTokens="required"
```


heronrs (approved):
:fire:

PR #3493, opened 2022-02-03
Adds an app param to configure health check success codes

beastawakens:
Solves an issue for a customer trying to run a Squid proxy, which only returns a 400 code when the service is healthy.


Twsouza (approved):
LGTM. Should we add this parameter to the docs?

beastawakens:
> LGTM. Should we add this parameter to the docs?

Yeah, I will do, was waiting to see if anything needed to be tweaked first 😉

heronrs (approved):

PR #3498, opened 2022-02-11
Build the binary as statically linked

Twsouza:
The binary was not being built as statically linked but as dynamically linked.
This was causing errors on Alpine.


heronrs (approved):
:fire:

PR #3504, opened 2022-02-28
Add s3 SSL only policy and assign lambdas to the rack VPC

heronrs:
Adds two new security rules to rack resources:

- s3 managed buckets will only allow SSL traffic
- lambdas will be assigned to the rack VPC


provider/aws/formation/rack.json
```diff
@@ -445,12 +445,10 @@
       }
     },
     "SubnetPrivate0": {
-      "Condition": "Private",
```
heronrs, line 448: AWS recommends assigning at least 2 private subnets to lambdas assigned to a VPC

Twsouza (commented):
:1st_place_medal:

Twsouza (approved):
:+1:

PR #3506, opened 2022-03-02
Get all processes and stop task from build instance

Twsouza:
You cannot currently kill a stuck build process. `ProccessStop` only tries to kill processes in the `Instances` cluster, it should also try to kill the process in the `BuildInstances`.

### DIFF
--- a/provider/aws/processes_test.go
+++ b/provider/aws/processes_test.go
@@ -264,8 +264,12 @@ func TestProcessRunDetached(t *testing.T) {
 
 func TestProcessStop(t *testing.T) {
 	provider := StubAwsProvider(
-		cycleProcessListTasksRunning,
-		cycleProcessListTasksStopped,
+		cycleProcessListStackResources,
+		cycleProcessDescribeStacks,
+		cycleProcessListTasksByStack,
+		cycleProcessListTasksByService1,
+		cycleProcessListTasksByService2,
+		cycleProcessListTasksByStarted,
 		cycleProcessStopTask,
 	)
 	defer provider.Close()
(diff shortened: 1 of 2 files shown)


provider/aws/processes_test.go
```diff
@@ -264,8 +264,12 @@ func TestProcessRunDetached(t *testing.T) {
 
 func TestProcessStop(t *testing.T) {
 	provider := StubAwsProvider(
-		cycleProcessListTasksRunning,
```
heronrs, line 267: why did you remove these?

provider/aws/processes_test.go
```diff
@@ -1419,26 +1423,6 @@ var cycleProcessStopTask = awsutil.Cycle{
 	},
 }
 
-var cycleProcessStopTaskReason = awsutil.Cycle{
```
heronrs, line 1422: i guess this is not used?

heronrs (approved):
lgtm, just have some questions

provider/aws/processes_test.go
Twsouza: Yes, this is no longer used.

provider/aws/processes_test.go
Twsouza: Because I changed the method to get the process, before it only gets running and stopped, it didn't get the process from the BuildInstance.
The `cycleProcessListTasksRunning` and `cycleProcessListTasksStopped` don't satisfy the appTaskARNs request, I had to change the stub processes.

heronrs (approved):
LGTM

PR #3508, opened 2022-03-03
Add BuildKit support for rack arm64 image

heronrs:
Forgot to add BuildKit support to the ARM image.
Also fixing a failing test step(`docker build` logs are displayed a bit different when using BuildKit)


Twsouza (approved):
:+1:

heronrs:
This is blocked as looks like there's a bug with buildkit on arm64 archs https://github.com/moby/buildkit/issues/2704

PR #3510, opened 2022-03-08
Fix HttpProxy param

Twsouza:
The HttpProxy was failing due to not being able to `yum update`. The rack was giving the 502 status because it didn't have the `NO_PROXY` set.

The user still has to set the environment variables on convox.yml:

```
environment:
  - PORT=3000
services:
  web:
    build: .
    port: 3000
    environment:
      - http_proxy=10.0.1.124:8888
      - https_proxy=10.0.1.124:8888
      - NO_PROXY=169.254.170.2

```


heronrs (approved):
LGTM. Can we also add a section on the HttpProxy documentation highlighting that this won't change how the rack apps access the internet (i.e traffic won't go through the proxy)

PR #3513, opened 2022-03-14
Spot Instances and HA false should have one Instances

Twsouza:
A rack with HA is disabled and the user sets the SpotInstanceBid, instead of using 1 instance for the rack, it changes to use 3 instances. Since the HighAvailabilty is disabled, should use 1 as the desired capacity.


heronrs (approved):
lgtm

PR #3515, opened 2022-03-22
Attach VPC to lambda refactor

heronrs:
When not running in private mode, creating Nat gateways to allow rack lambdas access to AWS services can get expensive. To move away from that, when not running in private mode we won't create NATs but instead, four VPC endpoints to allow access to CloudFormation,S3,KMS and ECS services which is much cheaper than maintaining NATs


Twsouza (approved):
:+1:

PR #3517, opened 2022-03-31
Explicitly declare dependencies for the KMS Custom resource

heronrs:
I noticed that sometimes the rack installation can fail due to a timeout error when creating the `EncryptionKey`. Although it should respect the implicit dependency between `CustomTopic` and the interface endpoints, explicitly adding them to `EncriptionKey` seem to have fixed it.


Twsouza (approved):
:+1:

PR #3532, opened 2022-04-26
Block public access to all buckets

heronrs:
Blocks all public access to S3 rack-created buckets.

### DIFF
--- a/pkg/structs/object.go
+++ b/pkg/structs/object.go
@@ -5,6 +5,5 @@ type Object struct {
 }
 
 type ObjectStoreOptions struct {
-	Public  *bool `header:"Public"`
 	Presign *bool
 }
--- a/pkg/api/object_test.go
+++ b/pkg/api/object_test.go
@@ -8,7 +8,6 @@ import (
 	"strings"
 	"testing"
 
-	"github.com/convox/rack/pkg/options"
 	"github.com/convox/rack/pkg/structs"
 	"github.com/convox/stdsdk"
 	"github.com/stretchr/testify/mock"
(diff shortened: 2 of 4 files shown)


provider/aws/releases.go
```diff
@@ -246,7 +246,7 @@ func (p *Provider) ReleasePromote(app, id string, opts structs.ReleasePromoteOpt
 			return err
 		}
 
-		ou, err := p.ObjectStore(app, "", bytes.NewReader(data), structs.ObjectStoreOptions{Public: options.Bool(true)})
+		ou, err := p.ObjectStore(app, "", bytes.NewReader(data), structs.ObjectStoreOptions{Presign: options.Bool(true)})
```
heronrs, line 249: We are storing the new templates that CF will use for the deployment so we create a pre-signed URL - and CloudFormation will be able to download them.

provider/aws/formation/rack.json
```diff
@@ -2845,7 +2845,7 @@
       "DependsOn": [ "Instances" ],
       "Properties": {
         "Cluster": { "Ref": "Cluster" },
-        "DeploymentConfiguration": { "MinimumHealthyPercent": "50", "MaximumPercent": "200" },
+        "DeploymentConfiguration": { "MinimumHealthyPercent": "50", "MaximumPercent": "200" },
```
ddoc, line 2848: added whitespace?

pkg/structs/object.go
```diff
@@ -5,5 +5,6 @@ type Object struct {
 }
 
 type ObjectStoreOptions struct {
-	Public *bool `header:"Public"`
+	Public  *bool `header:"Public"`
+	Presign *bool
```
Twsouza, line 9: Can `Public` field be deleted?

Twsouza (commented):
Does switching from public to private interfere with current racks/apps?

heronrs:
> Does switching from public to private interfere with current racks/apps?

Nope - I'll still run the e2e tests though.
The migration from public to non-public will occur at the rack level when the rack is updated and at the apps level when there's a new rollout.

Twsouza (approved):
:ship:

PR #3538, opened 2022-05-20
Rack with HighAvailability false should AutoScale

Twsouza:
Implement the autoscale for no-HA racks. It will scale to a max of 3 instances (the max allowed by the AutoScalingGroup).

### DIFF
--- a/provider/aws/lambda/autoscale/handler.go
+++ b/provider/aws/lambda/autoscale/handler.go
@@ -14,9 +14,16 @@ import (
 	"github.com/aws/aws-sdk-go/service/ecs"
 )
 
+const (
+	HAInstanceCountParam   = "InstanceCount"
+	NoHaInstanceCountParam = "NoHaInstanceCount"
+)
+
 var (
-	CloudFormation = cloudformation.New(session.New(), nil)
-	ECS            = ecs.New(session.New(), nil)
+	CloudFormation     = cloudformation.New(session.New(), nil)
+	ECS                = ecs.New(session.New(), nil)
+	InstanceCountParam = HAInstanceCountParam
+	IsHA               = os.Getenv("HIGH_AVAILABILITY")
 )
 
 type Metrics struct {
@@ -72,9 +79,13 @@ func autoscale(desired int64) error {
 		UsePreviousTemplate: aws.Bool(true),
 	}
 
+	if IsHA == "false" {
+		InstanceCountParam = NoHaInstanceCountParam
+	}
+
 	for _, p := range res.Stacks[0].Parameters {
 		switch *p.ParameterKey {
-		case "InstanceCount":
+		case InstanceCountParam:
 			debug("ds = %+v\n", ds)
 			debug("*p.ParameterValue = %+v\n", *p.ParameterValue)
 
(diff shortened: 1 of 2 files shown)


provider/aws/lambda/autoscale/handler.go
```diff
@@ -298,7 +313,12 @@ func desiredCapacity(largest, total *Metrics) (int64, error) {
...
 	// minimum instance count is 3
-	if desired < 3 {
+	if IsHA == "true" && desired < 3 {
+		desired = 3
+	}
+
+	// for no-HA rack the max is 3
+	if IsHA == "false" && desired > 3 {
 		desired = 3
 	}
```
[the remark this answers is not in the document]
heronrs: 1 - I think it's fair to keep it at 3 for non HA - the whole point of non HA clusters is to prevent extra costs so if anyone is looking for more computing power, they should go with HA racks.

2 - Still reviewing it

heronrs (commented):
Still reviewing it but left some comments.

Twsouza:
@heronrs @ddoc as we spoke, non-HA can now scale up to 1000 but starts with only one instance. HA can scale in a minimum of 3 instances while non-HA can scale in 1 instance.
Even in non-HA, if the user sets the `SpotInstanceBid` it will have to set the `OnDemandMinCount` (or use the default 3). The `provider/aws/workers_spotreplace.go` uses the same logic for HA and non-HA, it will try to create the Spot Instances and the difference between the current Spot instances running and the Desired size will be created using OnDemand (respecting the `OnDemandMin`).

Twsouza:
If it's approved, I'm going to update it on documentation.

provider/aws/lambda/autoscale/handler.go
```diff
@@ -74,9 +84,9 @@ func autoscale(desired int64) error {
 
 	for _, p := range res.Stacks[0].Parameters {
 		switch *p.ParameterKey {
-		case "InstanceCount":
-			debug("ds = %+v\n", ds)
-			debug("*p.ParameterValue = %+v\n", *p.ParameterValue)
+		case icParam:
+			debug("param %s = %s\n", icParam, *p.ParameterKey)
+			debug("actual %s param = %+v\n", icParam, *p.ParameterValue)
```
heronrs, line 89: i think this should be `param value`

provider/aws/lambda/autoscale/handler.go
```diff
@@ -14,9 +14,16 @@ import (
...
 
+const (
+	HAInstanceCountParam   = "InstanceCount"
+	NoHaInstanceCountParam = "NoHaInstanceCount"
+)
+
 var (
 	CloudFormation = cloudformation.New(session.New(), nil)
 	ECS            = ecs.New(session.New(), nil)
+	icParam        = HAInstanceCountParam
```
heronrs, line 25: can you move this parameter inside the `autoscale` function?
seems like it's the only place it's being used

provider/aws/workers_spotreplace.go
```diff
@@ -12,6 +12,7 @@ import (
 )
 
 var (
+	icsParam = "InstanceCount"
```
heronrs, line 15: `icParam`?

heronrs (commented):
Looks good! Just added some minor suggestions

heronrs (approved):
Nice!

PR #3541, opened 2022-06-01
Don't create VPC endpoint if exist

Twsouza:
New rack param `ExistingVpcEndpoint`, the default is false, which means that the VPCEndpoints will be created. However, if the user already has the endpoints created, he can set it as `false` end the creation will be skipped (if not, will fail).

### DIFF
--- a/provider/aws/formation/rack.json
+++ b/provider/aws/formation/rack.json
@@ -35,6 +35,7 @@
     "ExistingVpcAndInternetGateway": {
       "Fn::And": [ { "Condition": "ExistingVpc" }, { "Condition": "InternetGateway" } ]
     },
+    "CreateVpcEndpoint": { "Fn::Equals": [ { "Ref": "ExistingVpcEndpoint" }, "false" ] },
     "HighAvailability": { "Fn::Equals": [ { "Ref": "HighAvailability" }, "true" ] },
     "HttpProxy": { "Fn::Not": [ { "Fn::Equals": [ { "Ref": "HttpProxy" }, "" ] } ] },
     "InstanceARM": {
@@ -593,6 +594,12 @@
       "Type": "String",
       "Default": ""
     },
+    "ExistingVpcEndpoint": {
+      "AllowedValues": [ "true", "false" ],
+      "Default": "false",
+      "Description": "Mark as true if you already have a VPC Endpoint for cloudformation, ecs, kms and s3",
+      "Type": "String"
+    },
     "NoHaInstanceCount": {
       "Default": "1",
       "Description": "The number of instances in the runtime cluster for no HighAvailable racks",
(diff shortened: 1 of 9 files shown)


provider/aws/formation/rack.json
```diff
@@ -2014,8 +2025,11 @@
         "Description": { "Fn::Join": [ "", [ "{\"Cluster\": \"", { "Ref": "Cluster" }, "\", \"Rack\": \"", { "Ref": "AWS::StackName" }, "\"}" ] ] },
         "Environment": {
           "Variables": {
+            "AS_ENDPOINT": { "Fn::Select": [ 1, { "Fn::Split": [ ":", { "Fn::Select": [ 0, { "Fn::GetAtt": [ "ASEndpoint", "DnsEntries" ] } ] } ] } ] },
```
Twsouza, line 2028: 1. The `{ "Fn::GetAtt": [ "ASEndpoint", "DnsEntries" ] }` returns a list of values
2. But we only need the first one: `{ "Fn::Select": [ 0, { "Fn::GetAtt": [ "ASEndpoint", "DnsEntries" ] } ] }`
3. the value returned by DnsEntries contain the hosted zone id then the dns entry, so we need to split: `{ "Fn::Split": [ ":", { "Fn::Select": [ 0, { "Fn::GetAtt": [ "ASEndpoint", "DnsEntries" ] } ] } ] }`
4. Then we only need the DNS entry, so we get only the index 1: `{ "Fn::Select": [ 1, { "Fn::Split": ...`

provider/aws/lambda/autoscale/handler.go
```diff
@@ -354,5 +354,21 @@ func min(ii ...int64) int64 {
 }
 
 func main() {
+	sessionConfig := &aws.Config{Region: aws.String(os.Getenv("REGION"))}
+	session, err := session.NewSession(sessionConfig)
+	if err != nil {
+		fmt.Printf("Error creating new session: %+v", err)
+		return
+	}
```
Twsouza, line 362: Change to use the `NewSession` because the `New` method is deprecated.

> session.New is deprecated: Use NewSession functions to create sessions instead. NewSession has the same functionality as New except an error can be returned when the func is called instead of waiting to receive an error until a request is made.  (SA1019)

heronrs:
Looks pretty good! Just some nit picks

provider/aws/templates/resource/postgres.tmpl
```diff
@@ -46,12 +46,12 @@
...
-        "Default": "9.6",
+        "Default": "12",
         "Type": "String"
       },
       "Family": {
         "Type": "String",
-        "Default": "postgres9.6",
+        "Default": "postgres12",
         "Description": "Postgres version family"
       },
```
Twsouza, line 56: AWS deprecated the Postgress 9.6, see [Deprecation of PostgreSQL 9.6](https://docs.aws.amazon.com/AmazonRDS/latest/PostgreSQLReleaseNotes/postgresql-versions.html#postgresql-versions-version96-deprecation-notice).

AWS is recommending upgrading Postgres to 12, so I changed the template to use this version.

> You should upgrade all your PostgreSQL 9.6 DB instances to PostgreSQL 12 or higher as soon as possible. We recommend that you first upgrade to minor version 9.6.20 or higher and then upgrade directly to PostgreSQL 12 rather than upgrading to an intermediate major version

provider/aws/lambda/autoscale/handler.go
```diff
@@ -354,5 +354,21 @@ func min(ii ...int64) int64 {
...
+
+	CloudFormation = cloudformation.New(
```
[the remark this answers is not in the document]
Twsouza: We need to initialize the session var and, if returns an error, stop the execution. So we need to check if returned error. We can't verify the error on the var block.

heronrs (approved):
:fire:

PR #3543, opened 2022-06-15
Classic to Application Load Balancer migration

heronrs:
Implementation for migrating the rack API Classic Load Balancer to an Application Load Balancer.

- cloudformation changes
- new cli command `convox rack sync` to retrieve the new rack API URL

It's important to note that with the ELB migration, the rack API hostname will change.
The switch between hostname will be transparent for console managed racks. For users running racks outside of the console, we provide the `convox rack sync --name rack-name` command that should be executed after the update. This will print the new hostname to stdout.

PR #3544, opened 2022-06-21
Add rack sync cli command

heronrs:
Same CLI changes from https://github.com/convox/rack/pull/3543/files#diff-51b5da2a582f49d4721b891b549cb71824d49e1820e60d6451e8580389d5d1c9R12

I'm adding in a separate PR so we can release it separately from the Cloud Formation changes


Twsouza (approved):
🔥

PR #3548, opened 2022-06-22
Use Customer Managed Key to default encrypt S3 buckets

Twsouza:
The rack and apps bucket will be encrypted using a Customer Managed Key (by default).

We will create a new bucket to store ELB (Classic or Application LB) logs, AWS doesn't allow CMK encryption for logs bucket.

### DIFF
--- a/provider/aws/formation/rack.json
+++ b/provider/aws/formation/rack.json
@@ -3269,7 +3269,9 @@
                 "kms:GenerateDataKey*",
                 "kms:DescribeKey"
               ],
-              "Resource": "*"
+              "Resource": [
+                { "Fn::GetAtt": ["CustomerManagedKey", "Arn"]}
+              ]
             }
           ]
         }


provider/aws/formation/rack.json
```diff
@@ -3228,13 +3228,80 @@
         }]
       }
     },
+    "CustomerManagedKMS": {
+      "DependsOn": ["ServiceRole", "SecureEnvironmentRole", "InstancesLifecycleRole", "InstancesLifecycleHandlerRole", "CustomTopicRole", "ApiRole"],
```
heronrs, line 3232: You don't need to specify dependencies for resources you explicitly reference in other properties.
like `{ "Fn::GetAtt": ["ServiceRole", "Arn" ]},`

provider/aws/formation/app.json.tmpl
```diff
@@ -514,14 +514,70 @@
 {{ end }}
 
 {{ define "state" }}
+  "AppCustomerManagedKMS": {
```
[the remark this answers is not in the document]
Twsouza: l need to give access to `ExecutionRole`, `ServiceRole`, and `TimerRole`, and they are created in the app.json.tmpl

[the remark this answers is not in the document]
heronrs: You can reference resources across stacks
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/walkthrough-crossstackref.html
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/intrinsic-function-reference-importvalue.html

we already do this in some places https://github.com/convox/rack/blob/master/provider/aws/formation/app.json.tmpl#L326

Twsouza:
Apps bucket will be updated on Deploy/Release promote. It won't cause any disruption.

provider/aws/formation/rack.json
```diff
@@ -3228,23 +3252,92 @@
...
+            {
+              "Effect": "Allow",
+              "Action": [
+                "kms:Encrypt",
+                "kms:Decrypt",
+                "kms:ReEncrypt*",
+                "kms:GenerateDataKey*",
+                "kms:DescribeKey"
+              ],
+              "Resource": "*"
```
heronrs, line 3272: We should probably reference `CustomerManagedKey` here so we only allow those actions for the key we're creating. There might be other KMS keys in the AWS account that we don't care about.

provider/aws/formation/app.json.tmpl
```diff
@@ -520,7 +527,12 @@
     "Properties": {
       "AccessControl": "Private",
       "BucketEncryption": {
-        "ServerSideEncryptionConfiguration": [ { "ServerSideEncryptionByDefault": { "SSEAlgorithm": "aws:kms" } } ]
+        "ServerSideEncryptionConfiguration": [ {
+          "ServerSideEncryptionByDefault": {
+            "SSEAlgorithm": "aws:kms",
```
heronrs, line 532: should we enable s3 bucket keys here as well?

provider/aws/formation/g1/app.json.tmpl
```diff
@@ -874,7 +891,12 @@
     "Properties": {
       "AccessControl": "Private",
       "BucketEncryption": {
-        "ServerSideEncryptionConfiguration": [ { "ServerSideEncryptionByDefault": { "SSEAlgorithm": "aws:kms" } } ]
+        "ServerSideEncryptionConfiguration": [ {
+          "ServerSideEncryptionByDefault": {
+            "SSEAlgorithm": "aws:kms",
```
heronrs, line 896: should we enable s3 bucket keys here as well?

heronrs (approved):

PR #3550, opened 2022-07-13
feat-487: Add cpu and mem utilization percentage for rack processes

nightfury1204:
This provides backend support for : https://github.com/convox/issues-private/issues/487

aws cloudwatch metrics only provides cpu and mem utilization metrics for ecs services and in this PR we used that for processes cpu and mem current utilization


heronrs (approved):

exalted:
Running on a v2 rack with:

```
client: 20220817173516
server: 20220817173516
```

… I wonder where do I see CPU and mem usage?

Neither `convox rack ps` nor `convox ps -a <app>` seems to show additional columns for those.

PR #3556, opened 2022-07-29
Fix environment variable bug when using convox apps cancel

heronrs:
When using `convox apps cancel` we were not injecting the environment variables properly

### DIFF
--- a/pkg/cli/apps.go
+++ b/pkg/cli/apps.go
@@ -118,7 +118,7 @@ func Apps(rack sdk.Interface, c *stdcli.Context) error {
 func AppsCancel(rack sdk.Interface, c *stdcli.Context) error {
 	aname := coalesce(c.Arg(0), app(c))
 
-	c.Startf("Cancelling deployment of <app>%s</app>\n", aname)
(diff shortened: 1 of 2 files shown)


pkg/cli/apps.go
```diff
@@ -118,37 +118,29 @@ func Apps(rack sdk.Interface, c *stdcli.Context) error {
 func AppsCancel(rack sdk.Interface, c *stdcli.Context) error {
 	aname := coalesce(c.Arg(0), app(c))
 
-	c.Startf("Cancelling <app>%s</app>", aname)
+	c.Startf("Cancelling deployment of <app>%s</app>\n", aname)
 
-	if err := rack.AppCancel(aname); err != nil {
-		return err
-	}
+	rl, err := helpers.ReleaseLatest(rack, aname)
```
heronrs, line 123: `ReleaseLatest` is the proper method for retrieving all release information. `ReleaseList` doesn't populate the environment variables param.

Twsouza (approved):
🚀

PR #3563, opened 2022-08-05
Using LogDriver to send logs to CloudWatch, Syslog or disable it

Twsouza:
By default, CloudWatch is enabled, it can be changed to `Syslog` or turned off setting `LogDriver` empty.

`SyslogDestination` must be provided if `LogDriver` is `Syslog`.

```
 $ convox rack params set LogDriver=Syslog
ERROR: to enable Syslog as the LogDriver you must pass SyslogDestination parameter
```

Syslog format can be changed on`SyslogFormat`, the default is `rfc5424` (same as the current Syslog resource).

### DIFF
--- a/provider/aws/formation/app.json.tmpl
+++ b/provider/aws/formation/app.json.tmpl
@@ -5,6 +5,7 @@
 
     "BlankIamPolicy": { "Fn::Equals": [ { "Ref": "IamPolicy" }, "" ] },
     "BlankLogBucket": { "Fn::Equals": [ { "Ref": "LogBucket" }, "" ] },
+    "EnableCloudWatch": { "Fn::Equals": [ { "Ref": "EnableCloudWatch" }, "Yes" ] },
     "BlankLogRetention": { "Fn::Equals": [ { "Ref": "LogRetention" }, "" ] },
     "BlankResourcePassword": { "Fn::Equals": [ { "Ref": "ResourcePassword" }, "" ] },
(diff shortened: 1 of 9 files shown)


Twsouza:
Hi @moisesrodriguez! We have a parameter to customize the log retention period, it's set at the rack level. You can see the documentation here https://docsv2.convox.com/management/logs#retention

provider/aws/formation/app.json.tmpl
```diff
@@ -139,6 +147,16 @@
...
     },
+    "SyslogDestination": {
+      "Type": "String",
+      "Description": "Syslog address destination, you need to pass the protocol to be used, e.g. tcp+tls://logsX.syslog.com:1234",
+      "Default": ""
+    },
+    "SyslogFormat": {
+      "Type": "String",
+      "Description": "Syslog format to sent to SyslogDestination.",
+      "Default": "rfc5424"
```
nightfury1204, line 158: if user provide invalid `SyslogFormat` will it cause to fail?
Do we need  like allowed values validation?

Twsouza: Invalid `SyslogFormat` will not update the stack and will fail. We would have to use a regex or something more complex to validate the format as it accepts ABNF reference and other RFCs.

See: 

- https://docs.docker.com/config/containers/logging/syslog/
- https://www.rfc-editor.org/rfc/rfc5424#section-6

nightfury1204: Sounds like it will be very hard to put a validation for something like that. I think we can proceed without that.

nightfury1204 (approved):

PR #3564, opened 2022-08-16
Do not fetch metrics if services don't have any process

Twsouza:
`ProcessList` will return an "empty" slice.

Changed the `cloudwatchMetrics` function to return a message for empty metrics queries.

### DIFF
--- a/provider/aws/metrics.go
+++ b/provider/aws/metrics.go
@@ -23,6 +23,10 @@ var (
 )
 
 func (p *Provider) cloudwatchMetrics(mdqs []metricDataQuerier, opts structs.MetricsOptions) (structs.Metrics, error) {
+	if len(mdqs) == 0 {
+		return nil, fmt.Errorf("can't fetch metrics with empty query")
+	}
+
 	period := ci(opts.Period, 3600)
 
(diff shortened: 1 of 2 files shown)


yuriadams (approved):

provider/aws/metrics.go
```diff
@@ -23,6 +23,10 @@ var (
 )
 
 func (p *Provider) cloudwatchMetrics(mdqs []metricDataQuerier, opts structs.MetricsOptions) (structs.Metrics, error) {
+	if len(mdqs) == 0 {
```
heronrs, line 26: is the current behavior failing somewhere?
can't we just return an empty result instead of the error?

Twsouza: > is the current behavior failing somewhere?
It fails on [ProcessList](https://github.com/convox/rack/blob/58cea6e71221f0883f7708224d67cbd9083ef3f0/provider/aws/processes.go) which is used by the `convox ps`, `convox exec` and `convox run`.

> can't we just return an empty result instead of the error?

My opinion is that returning an error can be more helpful in cases where it should return metrics and it's not returning.

heronrs (approved):

PR #3569, opened 2022-08-24
Set the Encrypted to false as the default value

Twsouza:
The default value for `StorageEncrypted` is false.

https://docs.aws.amazon.com/pt_br/AWSCloudFormation/latest/UserGuide/aws-resource-rds-dbinstance.html#cfn-rds-dbinstance-storageencrypted

### DIFF
--- a/provider/aws/formation/resource/mariadb.json.tmpl
+++ b/provider/aws/formation/resource/mariadb.json.tmpl
@@ -78,7 +78,7 @@
         "AllowMajorVersionUpgrade": "true",
         "AutoMinorVersionUpgrade": { "Ref": "AutoMinorVersionUpgrade" },
         "DBInstanceClass": { "Ref": "Class" },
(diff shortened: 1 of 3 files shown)


provider/aws/formation/resource/mariadb.json.tmpl
```diff
@@ -1,15 +1,14 @@
 {
   "AWSTemplateFormatVersion": "2010-09-09",
   "Conditions": {
-    "BlankEncrypted": { "Fn::Equals": [ { "Ref": "Encrypted" }, "" ] },
```
heronrs, line 4: why remove this one? not used?

Twsouza: It was not being used but I included it again because setting the value as `false` will cause the replacement.

Twsouza:
Implemented on https://github.com/convox/rack/pull/3577

PR #3572, opened 2022-08-30
Add e2e test for timers

yuriadams:


examples/httpd/convox.yml
```diff
@@ -2,3 +2,11 @@ services:
   web:
     image: httpd
     port: 80
+  worker:
+    build: ./worker
+    scale: 0
+timers:
+  cleanup:
```
Twsouza, line 9: can you rename the timer to `example` or something more descriptive?
`cleanup` can be misleading.

examples/httpd/convox.yml
```diff
@@ -2,3 +2,11 @@ services:
...
+    command: ./cmd.sh
+    schedule: "*/1 * * * ?"
+    service: worker
```
Twsouza, line 12: Can you link the timer to the `web` service? So you won't need to create another service, keeping it simpler.

Twsouza (changes requested):
I suggest you rename the timer and link it to the `web` service (and remove the worker).

heronrs (approved):
Looks good to me. Agree with @Twsouza on the comments.

examples/httpd/convox.yml
yuriadams: hey, @Twsouza, thank for your comments. Could you have a look again, please?

Twsouza (approved):
🔥

PR #3570, opened 2022-08-25
Add CI to test updating/downgrading rack and app

Twsouza:
To test when update a rack CI will:

1. Install a rack with the latest release
2. Create and deploy an app
3. Update the rack to the release-candidate
4. Run the tests
5. Uninstall

To test when downgrading a rack CI will:

1. Install a rack with the release-candidate
2. Create and deploy an app
3. Update the rack to the latest version
6. Run the test
7. Uninstall

PR #3574, opened 2022-09-06
Use MicroSecond instead of Millisecond

Twsouza:
Sometimes the request has a slight difference from the expected startTime. Possibly because when dividing, it would round up to the closest number.

The difference in the request caused it to fail intermittently. The error was happening a lot.

image

https://app.circleci.com/pipelines/github/convox/rack/4125/workflows/faa681cf-9cbe-4887-9ffa-15c68f14218a/jobs/9824

_I didn't want to include this commit into #3570 because the e2e was running and committing this would restart it._

### DIFF
--- a/pkg/helpers/aws.go
+++ b/pkg/helpers/aws.go
@@ -359,7 +359,7 @@ func CloudWatchLogsStream(ctx context.Context, cw cloudwatchlogsiface.CloudWatch
 	var start int64
 
 	if opts.Since != nil {
-		start = time.Now().UTC().Add((*opts.Since)*-1).UnixNano() / int64(time.Millisecond)
+		start = time.Now().UTC().Add((*opts.Since)*-1).Unix() * int64(time.Microsecond)
 		req.StartTime = aws.Int64(start)
 	}
 


heronrs (approved):

pkg/helpers/aws.go
```diff
@@ -359,7 +359,7 @@ func CloudWatchLogsStream(ctx context.Context, cw cloudwatchlogsiface.CloudWatch
 	var start int64
 
 	if opts.Since != nil {
-		start = time.Now().UTC().Add((*opts.Since)*-1).UnixNano() / int64(time.Millisecond)
+		start = time.Now().UTC().Add((*opts.Since)*-1).Unix() * int64(time.Microsecond)
```
nightfury1204, line 362: there is UnixMicro() bultin func in time pkg, use that?
```
// UnixMicro returns t as a Unix time, the number of microseconds elapsed since
// January 1, 1970 UTC. The result is undefined if the Unix time in
// microseconds cannot be represented by an int64 (a date before year -290307 or
// after year 294246). The result does not depend on the location associated
// with t.
func (t Time) UnixMicro() int64 {
	return t.unixSec()*1e6 + int64(t.nsec())/1e3
}
```

nightfury1204 (changes requested):

pkg/helpers/aws.go
nightfury1204: sorry go 1.16 doesn't have UnixMicro() 
which version we are using?

nightfury1204: but i still think there is issue:
https://cs.opensource.google/go/go/+/refs/tags/go1.16.15:src/time/time.go;l=601
```
const (
	Nanosecond  Duration = 1
	Microsecond          = 1000 * Nanosecond
	Millisecond          = 1000 * Microsecond

```
calculation should be `Unix() * 1000000`

let's just use
```
.UnixNano() / int64(time.Microsecond)
```

Twsouza: @nightfury1204 the problem was that, sometimes, with `time.Now().UTC()` we would get a high number, and dividing by `Microsecond` or `Millisecond` would cause it to round up to `1546300800001` instead of `1546300800000`.
I've tried to use both and it didn't work.

nightfury1204 (approved):

PR #3576, opened 2022-09-13
Add unit tests for generate, sync and toke pkg

nightfury1204:
https://github.com/convox/issues-private/issues/597
https://github.com/convox/issues-private/issues/599

For atom didn't add any test because it's only used in k8s


Twsouza (approved):
👍

heronrs (approved):

PR #3578, opened 2022-09-14
Configure Retryer and Max Retry to avoid throttling error

Twsouza:
Use MaxRetry on aws.Config and set custom timers on DefaultRetryer.

Also, updates Go to 1.17 as `golang.org/x/sys` only allows 2 versions behind the latest one. Go 1.19 was released last month.

### DIFF
--- a/pkg/cli/rack.go
+++ b/pkg/cli/rack.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
-	ss "github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/convox/rack/pkg/helpers"
 	"github.com/convox/rack/pkg/options"
@@ -379,7 +378,7 @@ func RackSync(rack sdk.Interface, c *stdcli.Context) error {
 
 	if c.String("name") != "" {
 		rname = c.String("name")
-		s, err := ss.NewSession(&aws.Config{})
+		s, err := helpers.NewSession()
 		if err != nil {
 			return err
 		}
--- a/pkg/helpers/aws.go
+++ b/pkg/helpers/aws.go
@@ -15,12 +15,18 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/client"
+	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/cloudformation/cloudformationiface"
 	"github.com/aws/aws-sdk-go/service/cloudwatchlogs"
 	"github.com/aws/aws-sdk-go/service/cloudwatchlogs/cloudwatchlogsiface"
 	"github.com/convox/rack/pkg/structs"
 	yaml "gopkg.in/yaml.v2"
+)
+
+const (
+	MAX_RETRY = 10
 )
 
 func AwsCredentialsLoad() error {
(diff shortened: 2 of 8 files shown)


pkg/cli/rack.go
```diff
@@ -379,14 +379,8 @@ func RackSync(rack sdk.Interface, c *stdcli.Context) error {
 
 	if c.String("name") != "" {
 		rname = c.String("name")
-		s, err := ss.NewSession(&aws.Config{})
```
heronrs, line 382: Can't we just configure the Retry/Throttling logic in this session instead?

pkg/helpers/aws.go
```diff
@@ -71,12 +72,12 @@ func AwsErrorCode(err error) string {
 	return ""
 }
 
-func CloudformationDescribe(cf cloudformationiface.CloudFormationAPI, stack string) (*cloudformation.Stack, error) {
+func CloudformationDescribe(stack string) (*cloudformation.Stack, error) {
```
heronrs, line 75: Same here, looks like `cf cloudformationiface.CloudFormationAPI` is the provider.Cloudformation struct. We can change the session there.
https://github.com/convox/rack/blob/aws-retryer-throttling/provider/kaws/system.go#L271
https://github.com/convox/rack/blob/master/provider/kaws/kaws.go#L159

It's a good practice to have a retry/throtling logic so it's ok that we configure for the other services too. (in the `initializeAwsServices` function)

provider/aws/apps.go
```diff
@@ -400,7 +401,7 @@ func (p *Provider) cleanup(app *structs.App) error {
 	shouldRetry := true
 
 	for i := 0; i < 60; i++ {
-		res, err := p.cloudformation().DescribeStacks(&cloudformation.DescribeStacksInput{
```
heronrs, line 403: Same here, configure session in p;cloudformation().

provider/aws/lambda/autoscale/handler.go
```diff
@@ -354,8 +357,20 @@ func min(ii ...int64) int64 {
 }
 
 func main() {
-	sessionConfig := &aws.Config{Region: aws.String(os.Getenv("REGION"))}
-	session, err := session.NewSession(sessionConfig)
+	session, err := session.NewSession(
```
heronrs, line 360: :+1:

provider/aws/system.go
```diff
@@ -286,7 +287,7 @@ func (p *Provider) SystemInstall(w io.Writer, opts structs.SystemInstallOptions)
 		bar.Finish()
 	}
 
-	dres, err := cf.DescribeStacks(&cloudformation.DescribeStacksInput{
+	dres, err := pkgcf.DescribeStacks(&cloudformation.DescribeStacksInput{
```
heronrs, line 290: same here, use session config from L255

provider/aws/system.go
```diff
@@ -455,7 +456,9 @@ func (p *Provider) SystemUninstall(name string, w io.Writer, opts structs.System
 
 	cf := cloudformation.New(session.New(&aws.Config{}))
```
heronrs, line 457: configure session here.

provider/aws/system.go
```diff
@@ -663,7 +666,7 @@ func cloudformationProgress(stack, token string, template []byte, w io.Writer) e
 			}
 		}
 
-		dres, err := cf.DescribeStacks(&cloudformation.DescribeStacksInput{
+		dres, err := pkgcf.DescribeStacks(&cloudformation.DescribeStacksInput{
```
heronrs, line 669: configure session on L623

provider/aws/system.go
```diff
@@ -690,14 +693,12 @@ func cloudformationProgress(stack, token string, template []byte, w io.Writer) e
 }
 
 func rackDependencies(name string) ([]string, error) {
-	cf := cloudformation.New(session.New(&aws.Config{}))
```
heronrs, line 693: configure session here.

heronrs (commented):
It looks good but i don't think we need the wrapper around the DescribeStack operation.
We can configure the retry logic on all existing session objects.
To make it a bit more DRY you could have a helper function, like 
`helpers.NewSession() ` so you don't need to pass the same config over and over

heronrs (approved):
:100:

PR #3581, opened 2022-09-15
Allow network whitelisting on the service level

beastawakens:
I think this syntax is correct but I'd recommend some manual testing of this, just to be sure!


provider/aws/formation/service.json.tmpl
```diff
@@ -238,7 +238,7 @@
           {{ end }}
           "Properties": {
             "Actions": [ { "Type": "forward", "TargetGroupArn": { "Ref": "BalancerTargetGroup{{ if .Internal }}Internal{{ end }}" } } ],
-            "Conditions": [ { "Field": "host-header", "Values": [ { "Fn::Join": [ ".", [ "{{$.App}}-{{.Name}}", { "Fn::ImportValue": { "Fn::Sub": "${Rack}:{{ router .Name $.Manifest }}Host" } } ] ] } ] } ],
+            "Conditions": [ { "Field": "host-header", "Values": [ { "Fn::Join": [ ".", [ "{{$.App}}-{{.Name}}", { "Fn::ImportValue": { "Fn::Sub": "${Rack}:{{ router .Name $.Manifest }}Host" } } ] ] } ] }, { "Field": "source-ip", "SourceIpConfig": { "Values": [ {{ range safeWhitelist .Whitelist }} "{{ . }}", {{ end }} ] } } ],
```
Twsouza, line 241: @beastawakens the e2e tests got an error here:

```
Promoting RHQCJJDYHME... ERROR: json syntax error: line 248 pos 241: invalid character ']' looking for beginning of value:             "Conditions": [ { "Field": "host-header", "Values": [ { "Fn::Join": [ ".", [ "ci2-web", { "Fn::ImportValue": { "Fn::Sub": "${Rack}:RouterHost" } } ] ] } ] }, { "Field": "source-ip", "SourceIpConfig": { "Values": [  "0.0.0.0/0",  ] } } ],
```

beastawakens: Uuurgh, I know the syntax is a nested mess but that's CloudFormation for you.  I'm not seeing an extra closing bracket though 🤔 there's 5 opening, and 5 closing by my reckoning? Can you spot it?

Twsouza: I don't see any extra closing. I think I've found the problem when you create the app with an empty whitelist it will create   `"SourceIpConfig": { "Values": [ "0.0.0.0/0", ] } } ]`, you can see that will create an invalid comma in the values list.

heronrs:
@beastawakens if you could also create some tests to assert this new functionality we would greatly appretiate

PR #3584, opened 2022-09-22
added /var/lib/docker/container path to volume allow list

radekl:
This is to allow log file tailing by log collectors that runs inside a container. See: https://docs.datadoghq.com/containers/docker/log/?tab=containerinstallation for an example usage.

I hope that this is a place for changes to gen2.


Twsouza:
@radekl can you update the branch with the latest changes? We introduced new unit and e2e tests...

Twsouza (approved):

heronrs (approved):

PR #3587, opened 2022-09-27
Ignore release test for forked PRs

Twsouza:
We don't expose env vars for forked PRs, so we won't run the release process on them. 
The `ci` job will not run on forked PRs. The `test` job will always run.


Twsouza:
https://github.com/convox/rack/pull/3586 was not configured correctly.

yuriadams (approved):

PR #3589, opened 2022-09-27
Create E2E test for canceling apps deployment

Twsouza:
It will run convox deploy and when the app is set as updating it will cancel the deployment. After completion, it will test if the deployment was canceled (the new feature should not be released) and if the env vars are still set.


yuriadams (dismissed):

nightfury1204 (approved):
:fire:

heronrs (dismissed):

nightfury1204 (approved):

heronrs (approved):

PR #3592, opened 2022-10-03
add region us-east-2 (ci v3 racks) on clean up script

yuriadams:
[task 645](https://github.com/convox/issues-private/issues/645)


Twsouza:
By the way, can you create a script to delete EC2 key pairs for v2 racks? they are not deleted and we have almost 2k keys.

yuriadams:
> I suggest moving this to [convox/convox](https://github.com/convox/convox), V3 uses some resources that are not deleted in this script.

Okay, I'm closing this PR to open in another project

PR #3594, opened 2022-10-07
Add additional env var for resources

Twsouza:
Split the `${RESOURCE_NAME}_URL` into : 

```
${RESOURCE_NAME}_USER
${RESOURCE_NAME}_PASS
${RESOURCE_NAME}_HOST
${RESOURCE_NAME}_PORT
${RESOURCE_NAME}_NAME
```

### DIFF
--- a/provider/aws/formation/resource/memcached.json.tmpl
+++ b/provider/aws/formation/resource/memcached.json.tmpl
@@ -10,6 +10,10 @@
       "Type": "String",
       "Default": "cache.t2.micro"
     },
+    "Encrypted": {
+      "Type": "String",
+      "Default": ""
+    },
     "Nodes": {
       "Type": "Number",
       "Default": "1"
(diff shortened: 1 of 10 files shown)


provider/aws/formation/resource/memcached.json.tmpl
```diff
@@ -5,11 +5,15 @@
...
-    },     
+    },
     "Class": {
       "Type": "String",
       "Default": "cache.t2.micro"
     },
+    "Encrypted": {
+      "Type": "String",
+      "Default": ""
+    },
```
Twsouza, line 16: The actual `memcached` resource is broken because it passes the Encrypted even if the parameter doesn't exist in the template.

provider/aws/formation/resource/memcached.json.tmpl
```diff
@@ -29,7 +33,12 @@
     }
   },
   "Outputs": {
-    "Url": { "Value": { "Fn::Sub": "${CacheCluster.ConfigurationEndpoint.Address}:${CacheCluster.ConfigurationEndpoint.Port}" } }
+    "Url": { "Value": { "Fn::Sub": "${CacheCluster.ConfigurationEndpoint.Address}:${CacheCluster.ConfigurationEndpoint.Port}" } },
+    "User": { "Value": "" },
```
heronrs, line 37: If user/pass and name are always empty I thin we can just remove them

provider/aws/formation/resource/redis.json.tmpl
```diff
@@ -56,7 +56,12 @@
           }
         ]
       }
-    }
+    },
+    "User": { "Value": "" },
```
heronrs, line 60: Same here

heronrs (commented):
Looks good! Also, since v2 and v3 docs are separated I think we should update the v2 docs too.

provider/aws/formation/resource/memcached.json.tmpl
heronrs: If these were used then I'd say yes(like maybe Memcached would expose a different variable name for "User" and we convert the variable name to "User" to keep it consistent) but since it's never used I don't think we need it. Up to you though

heronrs (approved):

PR #3607, opened 2022-11-21
Delete VPC endpoints and remove lambda from VPC

Twsouza:
### What is the feature/fix?

The VPC endpoint feature increased the costs, not worth the advantages. It removes the VPC Endpoints resources and usages. Also removes the lambda from the VPC.

No

### How to use/test it?

Update/create a rack using the RC version (to be created)

### Checklist
- [ ] New coverage tests
- [x] Unit tests passing
- [x] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,7 +56,7 @@ workflows:
           <<: *filter-releases-any
           name: ci/aws/public-existing-vpc
           provider: aws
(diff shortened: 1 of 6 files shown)


heronrs (approved):
:100:

nightfury1204 (approved):
:fire:

PR #3620, opened 2022-12-09
Append scheme to the URL to parse correctly

Twsouza:
### What is the feature/fix?

The std url package does not parse correctly if missing http(s)://

```
$ ./convox resources proxy postgres
proxying localhost:5432 to dev-v2-efsresource-httpd-resourcepostgres-n0w2v7qwskwy.c5mts8687wtb.us-east-2.rds.amazonaws.com:5432
$ ./convox resources proxy memcached
proxying localhost:11211 to dev-ca-j1jlewp39pdr.rctvfw.cfg.use2.cache.amazonaws.com:11211
```

No

### How to use/test it?

Install the version (to be created), deploy an app with Memcached resource, and run `convox resources proxy memcached`.

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated


pkg/cli/resources.go
```diff
@@ -196,7 +196,12 @@ func ResourcesProxy(rack sdk.Interface, c *stdcli.Context) error {
 		return fmt.Errorf("no url for resource: %s", r.Name)
 	}
 
-	u, err := url.Parse(r.Url)
+	resourceUrl := r.Url
+	if !strings.Contains(r.Url, "http") {
```
nightfury1204, line 200: nit
`strings.HasPrefix`?

pkg/cli/resources.go
```diff
@@ -196,7 +196,12 @@ func ResourcesProxy(rack sdk.Interface, c *stdcli.Context) error {
...
+		resourceUrl = fmt.Sprintf("http://%s", r.Url)
```
heronrs, line 201: should it be https instead?

Twsouza: it won't matter if it's not HTTP or HTTPS, the problem was that URL parsing expects a scheme and if doesn't have one then fails to parse the port. The important thing is the port.

nightfury1204 (approved):

heronrs (approved):

crohr:
@Twsouza is it me or did this change break proxying rack resources?

We've recently upgraded to the latest rack v2 version, and we can no longer proxy resources because the remote port is always 80. For instance:

```
convox rack resources proxy common-addons-db -p 13306
proxying localhost:13306 to mysql:80
```

Same for redis/postgres/etc resources.

When I try to replicate what I think is going on, I suspect the change introduced by this PR always ends up defaulting to port 80 since `http` scheme is always prepended to the URL?

image

PR #3623, opened 2022-12-14
feat-619: Add support to add custom tags

nightfury1204:
### What is the feature/fix?

https://github.com/convox/issues-private/issues/619

when updating racks tags, it will update asg tags not the instances since they are deployed by the asg. after asg tags updates any instances created after that will have the tags.

https://user-images.githubusercontent.com/12232198/207680693-7051c311-c3b6-4abd-9d4f-00f017d98705.mp4

no

### How to use/test it?

- use the cli of this version to install with tags or set the tags parameter
```
./convox2 rack install aws Tags="k1=v1,k2=v2" HighAvailability=false --name v2-local3
```

```
./convox2 rack params set Tags="k1=v1,k2=v2"
```

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/provider/aws/system.go
+++ b/provider/aws/system.go
@@ -34,6 +34,8 @@ import (
 const (
 	keyLength   = 32
 	nonceLength = 24
+
+	ParameterNameTags = "Tags"
 )
 
 type envelope struct {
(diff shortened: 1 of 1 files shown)


Twsouza (approved):
🔥

PR #3624, opened 2022-12-14
Worker monitor should not get LogGroup from Rack API service

Twsouza:
### What is the feature/fix?

Since the Monitor and Web API had their name changed to include in the beginning, the worker was mistakenly assuming that the stack was from an app.

The monitor worker should only get the LogGroup from service apps.

No.

### How to use/test it?

Install the RC and run `convox rack logs`, it shouldn't show any error.

### Checklist
- [ ] New coverage tests
- [x] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/provider/aws/aws.go
+++ b/provider/aws/aws.go
@@ -73,6 +73,7 @@ type Provider struct {
 	Private             bool
 	PrivateBuild        bool
 	Rack                string
+	RackApiServiceName  string
 	SecurityGroup       string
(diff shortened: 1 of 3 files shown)


yuriadams (approved):

heronrs (approved):
thanks for fixing this

nightfury1204 (approved):
:fire:

PR #3629, opened 2023-01-17
feat-829: Add support to restrict rack api access from certain ips

nightfury1204:
### What is the feature/fix?

Add support to restrict rack api access from certain ips

https://github.com/convox/issues-private/issues/829

https://user-images.githubusercontent.com/12232198/212930443-544bb3a5-4d05-4218-a573-7c44dd456338.mp4

no, but if user set wrong ips, then convox console and cli will not able to access it

### How to use/test it?

- Install/update rack to this version: convox rack update 20230117133524-feat-829_whitelist_ips
- then `convox rack params set WhiteList="54.144.53.172/24"`
- and check access

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/provider/aws/formation/rack.json
+++ b/provider/aws/formation/rack.json
@@ -156,7 +156,8 @@
         { "Condition": "NotExistingVpcAndBlankInternetGateway" },
         { "Condition": "ThirdAvailabilityZone" }
       ]
-    }
+    },
+    "WhiteListCIDRs": { "Fn::Not": [ { "Fn::Equals": [ { "Ref": "WhiteList" }, "" ] } ] }
   },
   "Mappings": {
     "RegionConfig": {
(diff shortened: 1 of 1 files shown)


Twsouza (approved):
lgtm

provider/aws/formation/rack.json
```diff
@@ -895,6 +896,11 @@
       "Description": "Dedicated Hardware",
       "Default": "default",
       "AllowedValues": [ "default", "dedicated" ]
+    },
+    "WhiteList": {
+      "Type": "String",
+      "Description": "Comma delimited list of CIDRs, e.g. `10.0.0.0/24,172.10.0.1/32`, to allow access to the rack api. Maximum 4 CIDR can be specified.",
```
Twsouza, line 902: Why is only 4 CIDR supported?

nightfury1204: As you can see, I am adding the CIRD block in the ALB rule, they have restriction on that(Target Groups per Action per Application Load Balancer: 5): https://docs.aws.amazon.com/elasticloadbalancing/latest/application/load-balancer-limits.html
So one is host header, other 4 is for the source ip

heronrs (approved):
Nice!

provider/aws/formation/rack.json
radekl: Why not updating Security Group?

PR #3631, opened 2023-01-27
Save racks installed locally through the CLI

Twsouza:
### What is the feature/fix?

When a rack is installed using the CLI locally, not through the console or local rack, the installation finishes but don't keep the rack host anywhere. The CLI now will store the racks installed through it in the `self-managed` file.

No, however past racks installed using the CLI won't be recoverable, only if the user creates the rack entry on `self-managed`.

### How to use/test it?

1. Login into a console (e.g. console.convox.com)
2. Install the rack through the RC CLI (to be created).
3. Check if the rack is listed on `convox racks`
4. Switch to the new rack
5. Uninstall it
6. Run `convox racks` and the rack should not be on the list

### Checklist
- [x] New coverage tests
- [x] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/ci/dependencies.sh
+++ b/ci/dependencies.sh
@@ -23,4 +23,6 @@ chmod +x ${GOPATH}/bin/convox
 
 # set ci@convox.com as id
 mkdir -p ~/.convox/
-echo ci@convox.com > ~/.convox/id
+
+# convox config dir path
+echo ci@convox.com > ~/.config/convox2/id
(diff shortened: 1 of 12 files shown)


Twsouza:
Depends on https://github.com/convox/stdcli/pull/4

Twsouza:
I'm waiting for the stdcli PR to be merge so I can update this PR to include the changes.

Twsouza:
stdcli package is updated

yuriadams (approved):

nightfury1204 (approved):

PR #3645, opened 2023-02-22
Restrict ApiPolicy acess 

Twsouza:
### What is the feature/fix?

ApiPolicy and ApiRole are used by the rack api to create the services (and resources such as services roles and their assignments). The policy scope should be restricted to convox resources.

No.

### How to use/test it?

Update/Create the rack on version `20230222192752-chore-873-apipolicy-permission`, and deploy a new app.

### Checklist
- [ ] New coverage tests
- [x] Unit tests passing
- [x] E2E tests passing
- [x] E2E downgrade/update test passing
- [ ] Documentation updated

### DIFF
--- a/provider/aws/formation/rack.json
+++ b/provider/aws/formation/rack.json
@@ -2933,13 +2933,9 @@
             {
               "Effect": "Allow",
               "Action": [
-                "iam:DeleteServerCertificate",
                 "iam:*Role",
-                "iam:GetServerCertificate",
-                "iam:ListServerCertificates",
                 "iam:ListPolicyVersions",
-                "iam:*Policy",
-                "iam:UploadServerCertificate"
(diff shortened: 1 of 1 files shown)


provider/aws/formation/rack.json
```diff
@@ -2932,7 +2932,11 @@
           "Statement": [
             {
               "Effect": "Allow",
-              "Action": [ "iam:*" ],
+              "Action": [
+                "iam:*Role",
```
nightfury1204, line 2936: does `*` allowed in the begining?

Twsouza: yes, we need access to get, create, delete, attach, detach roles
that's a shortcut.

nightfury1204 (approved):

PR #3674, opened 2023-05-04
Prepend the url scheme only if the resource doesn't have one

Twsouza:
### What is the feature/fix?

The resource `memcached` does not have a default scheme, and when the CLI tries to parse the URL it fails. Only when the resource doesn't have the scheme, it will prepend `https://` to it so the parse can be done correctly. The `https://` is just a mock, it will not be used by the proxy.

However in other cases, the resources already have a default scheme, so we don't need to prepend anything.

```
$ ./convox resources proxy memcached -r company/test-v2 -a nodejs
resourceUrl:  https://tes-ca-a9w1wj3z6inb.ggqx34.cfg.use1.cache.amazonaws.com:11211
proxying localhost:11211 to tes-ca-a9w1wj3z6inb.ggqx34.cfg.use1.cache.amazonaws.com:11211
^C

$ ./convox resources proxy db -r company/test-v2 -a nodejs       
resourceUrl:  mysql://app:9da1d8b0-e918-11ed-85a0-0e0742ca01a9@test-v2-nodejs-resourcedb-dxaplp4g31i7.crsmvbsswb2b.us-east-1.rds.amazonaws.com:3306/app
proxying localhost:3306 to test-v2-nodejs-resourcedb-dxaplp4g31i7.crsmvbsswb2b.us-east-1.rds.amazonaws.com:3306
^C

$ ./convox resources proxy postgres -r company/test-v2 -a nodejs 
resourceUrl:  postgres://app:9da1d8b0-e918-11ed-85a0-0e0742ca01a9@test-v2-nodejs-resourcepostgres-no8wpe8bxfby.crsmvbsswb2b.us-east-1.rds.amazonaws.com:5432/app
proxying localhost:5432 to test-v2-nodejs-resourcepostgres-no8wpe8bxfby.crsmvbsswb2b.us-east-1.rds.amazonaws.com:5432
^C

```

No

### How to use/test it?

- Create a rack and deplot an app with mysql, memcached and postgres resource
- Run `convox resources proxy` as showed above.

### Checklist
- [x] New coverage tests
- [x] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/pkg/cli/resources.go
+++ b/pkg/cli/resources.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"fmt"
 	"net/url"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -11,6 +12,10 @@ import (
 	"github.com/convox/rack/pkg/structs"
 	"github.com/convox/rack/sdk"
 	"github.com/convox/stdcli"
+)
+
+const (
+	RESOURCE_URL_PATTERN = `.+:\/{2}.+`
 )
 
 func init() {
@@ -197,8 +202,13 @@ func ResourcesProxy(rack sdk.Interface, c *stdcli.Context) error {
 	}
 
 	resourceUrl := r.Url
-	if !strings.HasPrefix(r.Url, "http") {
-		resourceUrl = fmt.Sprintf("http://%s", r.Url)
+	rex, err := regexp.Compile(RESOURCE_URL_PATTERN)
+	if err == nil {
+		// Test if the string matches the resource url pattern
+		if !rex.MatchString(resourceUrl) {
+			// add the scheme to the url
+			resourceUrl = fmt.Sprintf("https://%s", resourceUrl)
+		}
 	}
 
 	u, err := url.Parse(resourceUrl)
--- a/pkg/cli/resources.go
+++ b/pkg/cli/resources.go
@@ -15,6 +15,7 @@ import (
 )
 
 const (
+	// RESOURCE_URL_PATTERN is used to check if the resrouce URL starts with http:// or https://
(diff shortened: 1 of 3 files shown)


pkg/cli/resources.go
```diff
@@ -197,8 +202,13 @@ func ResourcesProxy(rack sdk.Interface, c *stdcli.Context) error {
...
 
 	resourceUrl := r.Url
-	if !strings.HasPrefix(r.Url, "http") {
-		resourceUrl = fmt.Sprintf("http://%s", r.Url)
+	rex, err := regexp.Compile(RESOURCE_URL_PATTERN)
+	if err == nil {
+		// Test if the string matches the resource url pattern
+		if !rex.MatchString(resourceUrl) {
+			// add the scheme to the url
+			resourceUrl = fmt.Sprintf("https://%s", resourceUrl)
```
nightfury1204, line 210: previous one had `http`, but here we are adding `https://`?

pkg/cli/resources.go
```diff
@@ -13,6 +14,10 @@ import (
 	"github.com/convox/stdcli"
 )
 
+const (
+	RESOURCE_URL_PATTERN = `.+:\/{2}.+`
```
nightfury1204, line 18: what pattern it solves
add example or unit test on that

pkg/cli/resources.go
Twsouza: no reason, whether it's http or https the proxy won't use it, it's just used by the url parser.

pkg/cli/resources.go
Twsouza: I added the unit test, I'll add a comment above explaining it.

PR #3684, opened 2023-06-13
Remove ExistingGatewayAttachment on installing into existing VPC

Twsouza:
### What is the feature/fix?

AWS changed the behavior for existing for VPC Gateway attachment, even previous versions are not working anymore (I couldn't find any notes on that, I came at this conclusion by testing old releases)

No.

### How to use/test it?

Create a first rack (it can be the default settings), then try to install using the same VPC and IG, see https://docsv2.convox.com/reference/vpc-configurations#installing-into-an-existing-vpc

### Checklist
- [ ] New coverage tests
- [ ] Unit tests passing
- [ ] E2E tests passing
- [ ] Documentation updated

### DIFF
--- a/provider/aws/formation/rack.json
+++ b/provider/aws/formation/rack.json
@@ -1084,15 +1084,6 @@
         "VpcId": { "Ref": "Vpc" }
       }
     },
-    "ExistingGatewayAttachment": {
-      "Type": "AWS::EC2::VPCGatewayAttachment",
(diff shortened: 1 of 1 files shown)


ntner (approved):


- Rack install to existing VPC from previous rack versions
- Rack install to existing VPC at rack version
- App and stress testing in existing VPC

PR #3735, opened 2024-07-04
Fix scaling when service autoscaling is enabled and also fix logs rate limit error

nightfury1204:
### What is the feature/fix?

Fix scaling when service autoscaling is enabled and also fix logs rate limit error

### DIFF
--- a/pkg/helpers/aws.go
+++ b/pkg/helpers/aws.go
@@ -364,8 +364,10 @@ func CloudWatchLogsStream(ctx context.Context, cw cloudwatchlogsiface.CloudWatch
 	var start int64
 
 	if opts.Since != nil {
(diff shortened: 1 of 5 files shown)


ntner (approved):
<br>

- Scale configuration and command testing
- Application logs testing

bladealslayer:
@nightfury1204 @ntner this PR seems to introduce a regression - `$.Autoscale` in the service template now checks the rack-level autoscaling setting and doesn't take properly into account per-service auto-scaling config. As result, if the rack itself has Autoscale=No, one can't deploy apps with services that have autoscaling enabled. The `AutoscalingTarget` resource doesn't render in the template resulting in a broken template.

PR #3736, opened 2024-07-30
Removed app specific unique params when importing/exporting to avoid error

numcys:
### What is the feature/fix?

Encountered error where some app parameters which are unique to every app, were being copied while doing app import, this caused failure in copying app parameters while using `convox apps import` . To resolve this before importing app parameters, we are removing the unique parameters.

### DIFF
--- a/pkg/cli/apps.go
+++ b/pkg/cli/apps.go
@@ -412,7 +412,7 @@ func appExport(rack sdk.Interface, c *stdcli.Context, app string, w io.Writer) e
 	}
 
 	// Remove app unique parameters before exporting app
-	uniqueParams := []string{"Rack","LogBucket","ResourcePassword"}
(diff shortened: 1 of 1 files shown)


pkg/cli/apps.go
```diff
@@ -576,10 +577,16 @@ func appImport(rack sdk.Interface, c *stdcli.Context, app string, r io.Reader) e
...
+		uniqueParams := []string{"Rack","LogBucket","ResourcePassword"}
+
+		for _,param := range uniqueParams{
+			delete(a.Parameters, param);
+		}
+
 		for k, v := range a.Parameters {
 			if v != ae.Parameters[k] {
 				change = true
-				break
```
nightfury1204, line 582: why remove this?

nightfury1204 (approved):

PR #3741, opened 2024-09-12
Adding policy to cloudformation stack role of GEN-1 app

numcys:
### What is the feature/fix?
Gen-1 apps have a param - "TaskRole" to attach a IAM role to the ECS Task running the container. Currently due to insufficient permissions to assign IAM roles, the cloudformation stack isn't able to attach the IAM role with permissions to the task definition. This fix attaches the necessary policy of 'iam:PassRole' to the IAM role attached to CloudFormation stack.

### DIFF
--- a/provider/aws/formation/g1/app.json.tmpl
+++ b/provider/aws/formation/g1/app.json.tmpl
@@ -171,6 +171,21 @@
                 }
               ]
             }
+          },
+          {
+            "PolicyName": "PassRolePolicy",
+            "PolicyDocument": {
+              "Version": "2012-10-17",
+              "Statement": [
+                {
+                  "Effect": "Allow",
+                  "Action": [
+                    "iam:PassRole"
+                  ],
+                  "Resource": "*"
+                }
+              ]
+            }
           }
         ]
       }
--- a/provider/aws/formation/rack.json
+++ b/provider/aws/formation/rack.json
@@ -1299,6 +1299,21 @@
                 }
               ]
             }
+          },
+          {
+            "PolicyName": "PassRolePolicy",
+            "PolicyDocument": {
+              "Version": "2012-10-17",
+              "Statement": [
+                {
+                  "Effect": "Allow",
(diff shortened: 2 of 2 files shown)


provider/aws/formation/rack.json
```diff
@@ -1299,6 +1299,21 @@
...
+            "PolicyName": "PassRolePolicy",
+            "PolicyDocument": {
+              "Version": "2012-10-17",
+              "Statement": [
+                {
+                  "Effect": "Allow",
+                  "Action": [
+                    "iam:PassRole"
+                  ],
+                  "Resource": "*"
```
nightfury1204, line 1313: Why this role is needed for gen1 app but not for gen2 app?

provider/aws/formation/g1/app.json.tmpl
```diff
@@ -171,6 +171,21 @@
...
+            "PolicyName": "PassRolePolicy",
+            "PolicyDocument": {
+              "Version": "2012-10-17",
+              "Statement": [
+                {
+                  "Effect": "Allow",
+                  "Action": [
+                    "iam:PassRole"
+                  ],
+                  "Resource": "*"
```
nightfury1204, line 185: Does it has to be `*`?
gen2 app doesn't need `*`, but why gen1 need that?

nightfury1204 (changes requested):

provider/aws/formation/g1/app.json.tmpl
numcys: This is open role according to whichever permission user wants to assign, and can range from any service on aws upto any level the user wishes to assign, as such we cannot limit anything here otherwise it will fail for some Role arn. Gen2 assigns Policy and as such does not need us to do it, as the user already creates their policy and then assigns it but in gen1 we are responsible for giving necessary permissions to cloud formation stack to assign the role to ecs task definition.

nightfury1204 (approved):

PR #3743, opened 2024-10-23
convox run supports policies added to service

numcys:
### What is the feature/fix?
Added a fix for convox run to support the policies added to service.

### DIFF
--- a/provider/aws/processes.go
+++ b/provider/aws/processes.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/cloudformation"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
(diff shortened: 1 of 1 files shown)


nightfury1204 (approved):

provider/aws/processes.go
```diff
@@ -1191,10 +1192,38 @@ func (p *Provider) generateTaskDefinition2(app, service string, opts structs.Pro
 		}
 	}
 
+	td, err := p.stackResource(fmt.Sprintf("%s-%s", p.Rack, app), "ServiceWeb")
```
khoinguyenm, line 1195: our cluster doesn't have service `web` so it throw error `ERROR: resource not found: ServiceWeb` when we do `convox run [other service name]`

nightfury1204: @numcys can you fix this

nightfury1204: @khoinguyenm we have released the fix: https://github.com/convox/rack/releases/tag/20241104181809
Thanks for your patience

PR #3751, opened 2025-03-24
Fix spot replace, env value with '=' and remove sensitive information autoscaling logs

nightfury1204:
### What is the feature/fix?

- Fix env to support '=' sign to be included in the env value
- Remove sensitive information from autoscale handler
- Fix spot replace when instance is in rollback state


ntner (approved):
<br>

- Convox ENV testing
- Additional Spot Instance and ASG testing
- Import & Export testing

PR #3759, opened 2025-05-02
Enable build using Fargate tasks with Kaniko and crane

Twsouza:
### What is the feature/fix?

This feature adds support for building container images using ECS Fargate instead of EC2 instances with Docker. The build process now uses Kaniko (for daemonless Docker builds) and Crane (for pushing images to ECR), enabling fully serverless builds without Docker-in-Docker.

The previous system depended on EC2 instances running Docker. This update introduces a `BuildMethod = fargate` option and associated task definition (`ApiBuildFargate`) that runs builds in a secure, isolated Fargate environment using Kaniko. Build logs are streamed from CloudWatch. 

The implementation supports both the new Fargate-based and classic EC2-based builds.

No breaking changes to existing EC2-based build behavior.

### How to use/test it?

1. Set the required parameters in the rack:
   ```bash
   convox rack params set \
     FargateBuildCpu="1 vCPU" \
     FargateBuildMemory="3 GB" \
     BuildMethod=fargate
   ```

2. Deploy an application:
   ```bash
   convox deploy
   ```

3. Wait for the build to complete and open the service(s):
   ```bash
   convox services
   ```


pkg/build/build.go
```diff
@@ -36,39 +42,42 @@ type Options struct {
...
+	writer io.Writer    // tee to Output + logs
 }
 
+// New prepares a Build instance but does NOT start any long‑running work.
 func New(opts Options) (*Build, error) {
 	b := &Build{Options: opts}
 
-	b.Exec = &exec.Exec{}
-
+	// sane defaults ----------------------------------------------------------
```
nightfury1204, line 62: remove?

pkg/build/build.go
```diff
@@ -36,39 +42,42 @@ type Options struct {
...
+		if b.Generation == "2" {
 			b.Manifest = "convox.yml"
-		default:
+		} else {
 			b.Manifest = "docker-compose.yml"
 		}
 	}
 
-	r, err := sdk.NewFromEnv()
+	// dependencies -----------------------------------------------------------
```
nightfury1204, line 71: remove?

pkg/build/build.go
```diff
@@ -78,76 +87,92 @@ func New(opts Options) (*Build, error) {
...
 
-func (bb *Build) Printf(format string, args ...interface{}) {
+// printf helper that always targets the build‐scoped writer
+func (bb *Build) Printf(format string, args ...interface{}) { // nolint: revive  — this is intentional
 	fmt.Fprintf(bb.writer, format, args...)
 }
 
+// ---------------------------------------------------------------------------
+// private helpers below
+// ---------------------------------------------------------------------------
```
nightfury1204, line 105: remove?
why there's unnecessary comment?
Did you use some tools to generate the comments?

nightfury1204 (changes requested):
Can remove unnecessary comments?

pkg/build/build.go
Twsouza: removed

pkg/build/build.go
Twsouza: removed

pkg/build/build.go
Twsouza: removed
yes, I used the Copilot to review the code and some comments.

nightfury1204 (approved):

PR #3796, opened 2026-05-23
Add build cache support to EC2 and Fargate BuildMethods

ntner:
### What is the feature/update/fix?

**Feature: Persistent Build Cache (EC2 and Fargate)**

Apps can now reuse build layers across builds through a persistent cache. When enabled, build layers are written to a dedicated `{rack}-build-cache` ECR repository and pulled back on later builds, so unchanged layers are restored instead of rebuilt. The cache repository is created separately from your app image registries, so cache data is never co-mingled with deployable images. This works on both build methods: EC2 builds (which use `docker buildx`) and Fargate builds (which use kaniko) both write to and read from the same cache repository.

Caching is off by default and is controlled entirely through rack parameters. The cache repository is created only when you enable caching, and a separate parameter controls whether old cache images are automatically expired. The new parameters are:

| Parameter | Type | Default | Effect |
|-----------|------|---------|--------|
| `BuildCache` | String (`Yes`/`No`) | `No` | Enable persistent build caching. Builds write layers to the dedicated `{rack}-build-cache` ECR repository and reuse them on later builds. |
| `BuildCacheCleanup` | String (`Yes`/`No`) | `No` | Automatically expire old cache images. When `No`, cache images are kept indefinitely. |
| `BuildCacheRetentionDays` | Number (1-365) | `30` | Days to retain cache images before expiry. Applies only when `BuildCacheCleanup=Yes`. |

The cache repository is created only when `BuildCache=Yes` and is removed together with the rack (it uses `EmptyOnDelete`). Cleanup is a CloudFormation-managed ECR lifecycle policy that is attached to the cache repository only when `BuildCacheCleanup=Yes`. With cleanup off there is no expiry policy, so cache images are kept indefinitely. This is a storage-cost consideration rather than a data-loss concern, since cache images are regenerable. Disabling `BuildCache` removes the cache repository and all of its images.

---

### How to use it?

Build caching is opt-in. Enable it with a single rack parameter change. Batch related parameters into one command:

```
convox rack params set BuildCache=Yes -r my-rack
```

To enable caching and also expire old cache images automatically (for example, after 14 days):

```
convox rack params set BuildCache=Yes BuildCacheCleanup=Yes BuildCacheRetentionDays=14 -r my-rack
```

After the rack update completes, your next build populates the cache, and subsequent builds reuse the cached layers. There is nothing to change in `convox.yml` or your Dockerfiles.

To turn caching off again (this also removes the `my-rack-build-cache` repository and its images):

```
convox rack params set BuildCache=No -r my-rack
```

Confirm the parameter values at any time:

```
convox rack params -r my-rack
```

---

### Does it have a breaking change?

No. This change is purely additive and backward compatible.

All three parameters default to behavior-preserving values (`BuildCache=No`, `BuildCacheCleanup=No`, `BuildCacheRetentionDays=30`), so a rack that does not set them builds exactly as it did before. The cache repository and its lifecycle policy are created only when you opt in. No CloudFormation Outputs, Parameters, or resource logical IDs are removed or renamed, and no existing parameter defaults change.

A note on coverage and expectations: the build cache does not apply to gen1 apps, which continue to build exactly as before without it. Fargate (kaniko) caching is less aggressive than EC2 (buildkit). On Fargate it does not resume after the first cache miss, does not cache `COPY` layers by default, and has limited multi-stage caching, so hit rates on Fargate are lower than on EC2 for the same Dockerfile. On the EC2 build method, if the `docker buildx` plugin is unavailable the build proceeds normally without caching and prints a warning.

Downgrade is clean. Because the parameters are additive with safe defaults, a downgrade to a version that does not declare them drops them without breaking stack updates.

---

### Requirements

To use this feature, you must update to rack version `20260531001024` or newer.

- Check your rack's version with `convox rack`
- Update your rack with `convox rack update`

### DIFF
--- a/cmd/build/Dockerfile
+++ b/cmd/build/Dockerfile
@@ -22,7 +22,8 @@ RUN curl -sL "https://github.com/google/go-containerregistry/releases/download/$
     rm go-containerregistry.tar.gz
 
 # Kaniko runtime
-FROM gcr.io/kaniko-project/executor:v1.23.2-debug
+FROM ghcr.io/osscontainertools/kaniko:v1.27.5-debug
+ENV FF_KANIKO_OCI_STAGES=0
 
 # Create /tmp so Kaniko can write
 RUN mkdir -p /tmp && chmod 1777 /tmp
--- a/cmd/build/Dockerfile.arm
+++ b/cmd/build/Dockerfile.arm
@@ -24,7 +24,8 @@ RUN curl -sL "https://github.com/google/go-containerregistry/releases/download/$
     rm go-containerregistry.tar.gz
 
 # Kaniko runtime
-FROM gcr.io/kaniko-project/executor:v1.23.2-debug
+FROM ghcr.io/osscontainertools/kaniko:v1.27.5-debug
+ENV FF_KANIKO_OCI_STAGES=0
 
 # Create /tmp so Kaniko can write
 RUN mkdir -p /tmp && chmod 1777 /tmp
--- a/cmd/build/main.go
+++ b/cmd/build/main.go
@@ -26,7 +26,9 @@ var (
 	flagApp         string
 	flagAuth        string
 	flagBuildArgs   StringSlice
+	flagBuildCache  string
 	flagCache       string
+	flagCacheRepo   string
 	flagDevelopment string
 	flagEnvWrapper  string
 	flagGeneration  string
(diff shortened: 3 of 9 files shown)


mzihlmann:
Hi @ntner,

First off, thank you for the trust in osscontainertools, great to see it running on the Fargate build path.

I noticed `cmd/build/Dockerfile` sets `FF_KANIKO_OCI_STAGES=0`, which keeps the legacy tarball/docker-v2 path for inter-stage storage. A heads-up: that flag is on the way out, it's scheduled for removal in v1.28.0, after which OCI-layout stages become the default and only behaviour. If you disable it to preserve prior behaviour through the v1.23 → v1.27 upgrade, you can most likely drop it now and move to OCI layout with no downside. If you set it deliberately, e.g. to keep docker-v2 mediatypes for the ECR-backed build cache or the crane push step, I'd really like the specifics. I want to make sure v1.28.0 doesn't break your pipeline when the flag goes, and if there's a genuine OCI-mediatype gap I'd rather fix it on our side before then.

Based on your writing here, you might be interested in [`--cache-copy-layers`](https://github.com/osscontainertools/kaniko/#flag---cache-copy-layers) and [`FF_KANIKO_CACHE_PROBE_AFTER_MISS`](https://github.com/osscontainertools/kaniko/#flag-ff_kaniko_cache_probe_after_miss)

Happy to help either way, and thanks for your trust again :bow:

